PR #3, opened 2019-09-10
Enable HLint

cocreature:


fmt.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+set -eou pipefail
+curl -sSL https://raw.github.com/ndmitchell/hlint/master/misc/run.sh | sh -s .
```
garyverhaegen-da: Should this point to a specific commit? e.g. lock it to its current definition:
```suggestion
curl -sSL https://raw.githubusercontent.com/ndmitchell/neil/80e8e6b9727929fc1366110874969f532ed47774/misc/run.sh | sh -s -- hlint .
```

cocreature: I would just go with the unfixed version. HLint is pretty stable so at worst we’ll pick up new hints.

Just pinning the script also doesn’t pin the version of `hlint` so it’s not going to help much.

garyverhaegen-da: Ok.

PR #24, opened 2019-09-11
Add multi environment support

ndmitchell:
Following on from https://github.com/digital-asset/daml/pull/2808, here is the rebased version.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -23,6 +23,7 @@ import Development.IDE.Types.Location
 import Development.IDE.Types.Diagnostics
 import Development.IDE.Types.Options
 import Development.IDE.Types.Logger
+import Development.IDE.GHC.Util
 import qualified Data.Text as T
 import qualified Data.Text.IO as T
 import Language.Haskell.LSP.Messages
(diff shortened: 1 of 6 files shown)


exe/Main.hs
```diff
@@ -84,7 +82,7 @@ main = do
 
         putStrLn "\n[3/6] Initialising IDE session"
         vfs <- makeVFSHandle
-        ide <- initialise mainRule (showEvent lock) logger (defaultIdeOptions $ return env) vfs
+        ide <- initialise mainRule (showEvent lock) logger (defaultIdeOptions $ return $ const $ return env) vfs
```
cocreature: This will need to change once multiple environments are actually used. Maybe worth adding a comment so we don’t forget about it later.

src/Development/IDE/Core/Rules.hs
```diff
@@ -317,14 +321,33 @@ generateCoreRule =
...
         setPriority priorityGenerateCore
-        packageState <- useNoFile_ GhcSession
+        packageState <- hscEnv <$> use_ GhcSession file
         liftIO $ compileModule packageState tms tm
 
+
+-- A local rule type to get caching. We want to use newCache, but it has
+-- thread killed exception issues, so we lift it to a full rule.
+-- https://github.com/digital-asset/daml/pull/2808#issuecomment-529639547
+type instance RuleResult GhcSessionIO = GhcSessionFun
```
cocreature: These should go to the `RuleTypes` module to be consistent with the rest.

cocreature: Sounds reasonable, let’s go with that for now.

PR #27, opened 2019-09-11
WIP: Code action for type wildcards

hsenag:
Follow on from https://github.com/digital-asset/daml/pull/2830
Not for merging now, I'm waiting for https://github.com/digital-asset/daml/pull/2809 to land first.

It makes it possible to fill in type signatures interactively by typing _ and then using the code action to fill in what GHC inferred.

It doesn't always produce a valid type signature as the one GHC mentions might use unavailable type variables or be missing constraints, but it at least gets the user started. It also always adds parentheses around the inserted signature as it's hard to know when they are really needed or not.

@ndmitchell and @cocreature: We might want to think about how to organise code actions, as I can imagine more appearing quite quickly.

- Do we want the current fall through behaviour of suggestAction or to replace it with something that's more explicitly composed? For example my change will remove the existing code action that would add PartialTypeSignatures to the module. I think that's reasonable in this case as mostly people wouldn't want that, but there will likely be cases in future where we want both.

- Should we organise all the ad-hoc text recognition better? I understand it'll be replaced by an ADT exposed from GHC at some point in the future.

- Should we abstract the way we write code action test cases to reduce the boilerplate?


ndmitchell:
@hsenag - all the things you suggest are good ideas! The fall-through behaviour isn't right. Maybe we should write real parsers? Although those tend to be more brittle and have greater impacts from small changes or GHC version upgrades.

hsenag:
I was also wondering about writing a real parser and perhaps also having a proper ADT as an intermediate step between the error messages and the code actions. The biggest thing that stopped me considering it further was the code action for finding language extensions which is rather free form :-)

But I'm also not sure how much effort this is worth given the expected change in the GHC API itself to use an ADT, and what our migration strategy will be. Perhaps having our own ADT will help with that migration? I guess it'll require CPP for a while, and two different ways of recognising the messages.

ndmitchell:
Writing the ADT version is super simple, so I don't think we gain much by moving there first - other than by writing tests, which should confirm it still works.

The extension messages produced by GHC vary wildly based on where they are encountered. I wouldn't want to try and parse them all, but it is really the "correct" thing to do. At the moment if you name a data type `RecordWildCards` it will suggest that if you make an error.

hsenag:
The end state using the GHC ADT is very simple, but how will the transition be managed - won't there be a period where we support versions of GHC that include the ADT in the API and versions that don't?

Without a transition strategy for that period, we can't use the GHC ADT until we drop support for the older GHCs. So we'll either waste time writing new code actions with text processing, or we cut off older versions quickly which won't be great for users. I think it'd be reasonable for new code actions to only work with the newer GHCs at that point. But having new versions of `ghcide` cut off old users completely wouldn't be great given that other improvements are likely to be flowing at the same time.

If there is a transition period then I think it'll need CPP and having our own ADT might reduce the extent of that CPP.

I don't think we should try to improve the existing extension messages action until we do have the GHC ADT. That'd be a lot of work for not much gain.

ndmitchell:
I was imagining when the new ADT comes out we create `CodeActionsOld.hs` which is a copy and paste of whatever we have on that day, and then write `CodeActionsNew.hs` being a port of that to the ADT. We then freeze `CodeActionsOld.hs` and stop touching it. I think parsing the freeform strings to the ADT is going to be a lot of work.

Note that the current state of the ADT is that it doesn't even have a proposal yet. It's likely to be many years!

hsenag:
This ought to be ready to merge, once #40 lands.

hsenag:
This now ought to be ready to merge, once #40 lands.
I'll look into refactoring the structure+tests in a follow-up.

cocreature:
CI is failing due to an HLint error. There is a `fmt.sh` script in this repo (which should be significantly easier to use than the one in the old repo) but I’m not sure it works on windows. Just running HLint manually in the repo root should probably do the trick as well.

Issue #26, opened 2019-09-11
Add more GHC version tests


ndmitchell:
We know it works with GHC 8.4 today, but without CI, that's likely to stop being true relatively quickly.

Probably also want a GHC 8.8 test too.

cocreature:
PRs welcome :slightly_smiling_face:

ndmitchell:
PR in https://github.com/digital-asset/ghcide/pull/29.

cocreature:
We now have CI for 8.4. I believe `lsp-test` is blocked on something for `8.8` atm but I don’t recall the details.

ndmitchell:
Close this an open a separate ticket for 8.8 support? There are plenty of things that are blocked, so I don't suggest trying to hard for a few weeks.

cocreature:
Closing in favor of #33

PR #29, opened 2019-09-11
#26, test GHC 8.4

ndmitchell:
Done blind, but from local experiments.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -44,6 +44,43 @@ jobs:
       - bash: stack test || stack test || stack test
         # ghcide stack tests are flaky, see https://github.com/digital-asset/daml/issues/2606.
         displayName: 'stack test'
+      - bash: |
+          mkdir -p .azure-cache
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -44,6 +44,13 @@ jobs:
       - bash: stack test || stack test || stack test
         # ghcide stack tests are flaky, see https://github.com/digital-asset/daml/issues/2606.
         displayName: 'stack test'
+      - bash: stack setup --stack-yaml=stack84.yaml
```
cocreature: This isn’t quite right. This should be a separate CI job so we can run them in parallel.

ndmitchell: I don't have the skills to do that, at the very least without active experimentation, so if you want that before merge I suggest we close this PR.

cocreature: If you don’t mind, I can push to your PR and try to fix it when I find the time.

ndmitchell: Happily!

PR #32, opened 2019-09-11
building extension in correct path

CodiePP:
Signed-off-by: Alexander Diemand <codieplusplus@apax.net>

minor fix of correct path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Once you have got `ghcide` working outside the editor, the next step is to pick 
 
 Install the VS code extension (see https://code.visualstudio.com/docs/setup/mac for details on adding `code` to your `$PATH`):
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +59,7 @@ Once you have got `ghcide` working outside the editor, the next step is to pick
 
 Install the VS code extension (see https://code.visualstudio.com/docs/setup/mac for details on adding `code` to your `$PATH`):
 
-1. `cd compiler/ghcide/extension`
+1. `cd extension/`
```
ndmitchell: Any reason to include the trailing quote? I tend to see `cd` without that most of the time.

cocreature: Doesn’t matter but it also doesn’t do any harm so I’m happy to merge it this way.

PR #35, opened 2019-09-11
Update cabal file metadata to point to the new repo

cocreature:


stack.yaml.lock
```diff
@@ -0,0 +1,72 @@
+# This file was autogenerated by Stack.
```
shayne-fletcher-da: I'm not sure what these files are. I thought they were generated and temporary. Did you mean to commit this?

cocreature: Yes, commiting this makes sure builds are deterministic.

cocreature: Turns out that it doesn’t work well with testing multiple GHC versions so I’ve killed it again :woman_shrugging:

Issue #37, opened 2019-09-11
Module not found when using "this" package imports


ollef:
In a project I define my own `Data.IntMap` module which clashes with the one from containers, so I use the following code to import it:

```haskell
import qualified "this" Data.IntMap as IntMap
```
GHC accepts this as importing the `Data.IntMap` from the current project, but ghcide complains that it can't find the module `Data.IntMap` on this line.

To reproduce, either open `src/Context.hs` from [Sixty](https://github.com/ollef/sixty), or set up a project with e.g. these modules:

Lib.hs:
```haskell
{-# language PackageImports #-}

module Lib where

import "this" M -- ghcide errors on this line
```

M.hs
```haskell
module M where
```

It'd be nice if this worked.

Cheers!

ndmitchell:
I didn't know that syntax existed! Should be easy to support, just change https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Import/FindImports.hs#L70-L79 to have a case for `Just "this"` that _only_ looks the item up as a file. Tempted to try a PR?

ollef:
Sure thing, that doesn't sound too bad!

PR #39, opened 2019-09-11
Handle PackageImporting "this"

ollef:
"this" means that we only look for the module in the current package, so
we only look for a file in that case.

Fixes https://github.com/digital-asset/ghcide/issues/37.

### DIFF
--- a/src/Development/IDE/Import/FindImports.hs
+++ b/src/Development/IDE/Import/FindImports.hs
@@ -68,6 +68,12 @@ locateModule
     -> m (Either [FileDiagnostic] Import)
 locateModule dflags exts doesExist modName mbPkgName isSource = do
   case mbPkgName of
+    -- "this" means that we should only look in the current package
+    Just "this" -> do
+      mbFile <- locateModuleFile dflags exts doesExist isSource $ unLoc modName
+      case mbFile of
+        Nothing -> return $ Left $ notFoundErr dflags modName $ LookupNotFound []
+        Just file -> return $ Right $ FileImport file
     -- if a package name is given we only go look for a package
     Just _pkgName -> lookupInPackageDB dflags
     Nothing -> do


ndmitchell:
Looks great! Two notes (not necessarily for this PR, but just thoughts):

* The package lookup code is now getting a bit messy - there are three choices, 1) lookup in module, 2) lookup in package and 3) error. You can hit 1,2,3 (normal import), 1,3 (this import), 2,3 (package import). The code doesn't really flow that way, and it would be nice if it did, but potentially the overhead of abstracting it to look like that is more work than it saves.
* A test would be great to keep this working - I had never heard of "this" modules so can imagine someone breaking them in future.

src/Development/IDE/Import/FindImports.hs
```diff
@@ -68,6 +68,12 @@ locateModule
     -> m (Either [FileDiagnostic] Import)
 locateModule dflags exts doesExist modName mbPkgName isSource = do
   case mbPkgName of
+    -- "this" means that we should only look in the current package
+    Just "this" -> do
```
cocreature: Maybe just add a `_pkgName /= "this"` case to the first case and turn the `Nothing` case into a wildcard. Then we don’t need to duplicate the code.

ollef: Something like that might work, but the code is not actually the same for the `"this"` case and `Nothing` case because we don't even look in the package DB if there's a `"this"`.

cocreature: Ah very good point, probably reasonable to keep it that way for now then.

ollef:
Thanks for pointing me to where the tests should be added. I've added a simple test now. :)

I see the point about the code being a bit messy, but if it's alright I'll leave refactoring of it to you, as I don't exactly know what fix you had in mind.

Related note (but probably not something to fix in _this_ PR): I think we're missing an error for ambiguous (non-PackageImport) imports in the case where there's both a local and a package module with the same name. In the non-PackageImport case I think we need to look _both_ among the local modules and package modules to determine whether a name is ambiguous. I don't have time to fix that right now though.

ndmitchell:
I thought where there was a local module and a package module it preferred the local? I guess I may well have implemented what I thought was the semantics of imports, rather than the actual semantics... definitely important to nail that down before refactoring.

cocreature:
Totally fine to leave the refactoring out of this PR. My understanding was the same as @ndmitchell’s wtr to ambiguous imports and I seem to recall accidentally running into that with cabal but I could very well be wrong.

PR #41, opened 2019-09-11
Add preprocessor module

ndmitchell:
A non-functional refactoring. There are three preprocessor stages - unlit, cpphs and running custom -fpgm preprocessors (which I'm sure we'll do eventually). They have a nice segregated interface from the rest of the parser, so make sense to live entirely on their own, in their own module, which I've done. I'm also aware that GHC 8.10 exposes more preprocessor stuff, and that we might be able to switch to it at a future date based on runPhase. As such, having the "ideal" interface lets interested parties (e.g. @DanielG) see what we'd like it to be, and what we'd like it to do, clearly.

I did make it work for `.lhs-boot` files and optimised Literate Haskell but neither of those are very interested.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -32,7 +32,6 @@ import           Lexer
 import ErrUtils
 
 import qualified GHC
-import           Panic
 import           GhcMonad
 import           GhcPlugins                     as GHC hiding (fst3, (<>))
 import qualified HeaderInfo                     as Hdr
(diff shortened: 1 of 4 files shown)


src/Development/IDE/Core/Compile.hs
```diff
@@ -342,28 +272,9 @@ parseFileContents
        -> FilePath  -- ^ the filename (for source locations)
        -> Maybe SB.StringBuffer -- ^ Haskell module source text (full Unicode is supported)
        -> ExceptT [FileDiagnostic] m ([FileDiagnostic], ParsedModule)
-parseFileContents preprocessor filename mbContents = do
+parseFileContents sourcePlugin filename mbContents = do
```
cocreature: I don’t think this name is a good idea given that those are not actually source plugins. It would be great if we could make them source plugins at some point but for now something like `customPreprocessor` is probably a better name.

src/Development/IDE/Core/Preprocessor.hs
```diff
@@ -0,0 +1,133 @@
...
+import qualified GHC.LanguageExtensions as LangExt
+import Data.Maybe
+
+
+-- | Given a file and some contents, apply any necessary preprocessors,
+--   e.g. unlit/cpp. Return the resulting buffer and the DynFlags it implies.
+preprocessor :: GhcMonad m => FilePath -> Maybe StringBuffer -> ExceptT [FileDiagnostic] m (StringBuffer, DynFlags)
+preprocessor filename mbContents = do
+    -- Perform unlit
+    (isOnDisk, contents) <- if isLiterate filename then do
```
cocreature: The indentation here confuses me, I would prefer if we put `then` on a new line, indent both `then` and `else` and align the two.

PR #47, opened 2019-09-12
Defer type errors

jacg:
At present, all sorts of useful IDE featrues, such as diagnostics reporting and go-to-definition, stop working at the first sign of a type error. This PR attempts to fix this. The plan is:

+ defer type errors
+ report them as warnings during typechecking
+ re-upgrade them to errors after type checking

TODO:

- [x] (EDIT: fixed by #59) ~~The current implementation breaks old behaviour by changing the text of type error messages to contain fully qualified type names, e.g: `GHC.Types.Int` appears instead of `Int`.~~
- [x] There are other things besides type errors that could be deferred?
    - [x] Type Errors
    - [x] Typed Holes
    - [x] Out of scope variables

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -96,13 +96,6 @@ typecheckModule packageState deps pm =
             tcm2 <- mkTcModuleResult tcm
             return (map unDeferTypeErrors warnings, tcm2)
 
-unDeferTypeErrors :: (WarnReason, FileDiagnostic) -> FileDiagnostic
-unDeferTypeErrors (Reason Opt_WarnDeferredTypeErrors, fd) = upgradeWarningToError fd
-unDeferTypeErrors (                                _, fd) = fd
-
-upgradeWarningToError :: FileDiagnostic -> FileDiagnostic
-upgradeWarningToError (nfp, fd) = (nfp, fd{_severity = Just DsError})
-
 -- | Compile a single type-checked module to a 'CoreModule' value, or
 -- provide errors.
 compileModule
@@ -136,16 +129,26 @@ compileModule packageState deps tmr =
             return (map snd warnings, core)
 
 demoteTypeErrorsToWarnings :: ParsedModule -> ParsedModule
-demoteTypeErrorsToWarnings = (update_pm_mod_summary . update_hspp_opts) demoteTEsToWarns
+demoteTypeErrorsToWarnings =
+  (update_pm_mod_summary . update_hspp_opts) demoteTEsToWarns where
 
-demoteTEsToWarns :: DynFlags -> DynFlags
-demoteTEsToWarns = (`wopt_set` Opt_WarnDeferredTypeErrors) . (`gopt_set` Opt_DeferTypeErrors)
+  demoteTEsToWarns :: DynFlags -> DynFlags
+  demoteTEsToWarns = (`wopt_set` Opt_WarnDeferredTypeErrors)
+                   . (`gopt_set` Opt_DeferTypeErrors)
 
-update_hspp_opts :: (DynFlags -> DynFlags) -> ModSummary -> ModSummary
-update_hspp_opts up ms = ms{ms_hspp_opts = up $ ms_hspp_opts ms}
+  update_hspp_opts :: (DynFlags -> DynFlags) -> ModSummary -> ModSummary
+  update_hspp_opts up ms = ms{ms_hspp_opts = up $ ms_hspp_opts ms}
 
-update_pm_mod_summary :: (ModSummary -> ModSummary) -> ParsedModule -> ParsedModule
-update_pm_mod_summary up pm = pm{pm_mod_summary = up $ pm_mod_summary pm}
+  update_pm_mod_summary :: (ModSummary -> ModSummary) -> ParsedModule -> ParsedModule
+  update_pm_mod_summary up pm =
+    pm{pm_mod_summary = up $ pm_mod_summary pm}
+
+unDeferTypeErrors :: (WarnReason, FileDiagnostic) -> FileDiagnostic
+unDeferTypeErrors (Reason Opt_WarnDeferredTypeErrors, fd) = upgradeWarningToError fd
+  where
+    upgradeWarningToError :: FileDiagnostic -> FileDiagnostic
+    upgradeWarningToError (nfp, fd) = (nfp, fd{_severity = Just DsError})
+unDeferTypeErrors (                                _, fd) = fd
 
 addRelativeImport :: ParsedModule -> DynFlags -> DynFlags
 addRelativeImport modu dflags = dflags
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -145,9 +145,10 @@ demoteTypeErrorsToWarnings =
     pm{pm_mod_summary = up $ pm_mod_summary pm}
 
 unDefer :: (WarnReason, FileDiagnostic) -> FileDiagnostic
-unDefer (Reason Opt_WarnDeferredTypeErrors, fd) = upgradeWarningToError fd
-unDefer (Reason Opt_WarnTypedHoles        , fd) = upgradeWarningToError fd
-unDefer (                                _, fd) = fd
+unDefer (Reason Opt_WarnDeferredTypeErrors         , fd) = upgradeWarningToError fd
+unDefer (Reason Opt_WarnTypedHoles                 , fd) = upgradeWarningToError fd
+unDefer (Reason Opt_WarnDeferredOutOfScopeVariables, fd) = upgradeWarningToError fd
+unDefer ( _                                        , fd) = fd
 
 upgradeWarningToError :: FileDiagnostic -> FileDiagnostic
 upgradeWarningToError (nfp, fd) = (nfp, fd{_severity = Just DsError})
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -88,12 +88,11 @@ typecheckModule
     -> IO ([FileDiagnostic], Maybe TcModuleResult)
 typecheckModule packageState deps pm =
     fmap (either (, Nothing) (second Just)) $
-    runGhcEnv packageState $ do
-        demoteTypeErrorsToWarnings
+    runGhcEnv packageState $
         catchSrcErrors "typecheck" $ do
             setupEnv deps
             (warnings, tcm) <- withWarnings "typecheck" $ \tweak ->
-                GHC.typecheckModule pm{pm_mod_summary = tweak $ pm_mod_summary pm}
+                GHC.typecheckModule $ demoteTypeErrorsToWarnings pm{pm_mod_summary = tweak $ pm_mod_summary pm}
             tcm2 <- mkTcModuleResult tcm
             return (warnings, tcm2)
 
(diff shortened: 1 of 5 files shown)


hsenag:
It looks like the qualified names are something general to do with how `ghcide` gets warnings and nothing to do with setting `Opt_DeferTypeErrors`.

In the following code, there's a warning about the unnecessary constraint `Ord a`, and the whole warning has fully qualified names. If you add `-Werror` to `OPTIONS_GHC` to turn it into an error, it has unqualified names.

```
{-# OPTIONS_GHC -Wredundant-constraints #-}
module Foo where
foo :: Ord a => Int -> a -> Int
foo a b = a + 1
```

hsenag:
I tracked down why it's happening (I think) - the warnings collection code uses `mkPlainWarnMsg` which always qualifies symbols. We can extract the right setting with `queryQual` from the `PprStyle` and use `mkWarnMsg` instead, I'll send a patch later.

hsenag:
I think the three things we can/want to defer are
 - type-errors
 - typed-holes
 - out-of-scope-variables

I think at least typed-holes is implied by type-errors.

A few comments on the code:
 - I'm not sure about the merits of the current gyrations compared to just setting the option once-and-for-all in `runGhcEnv` (which is a lot simpler code-wise). @ndmitchell may have a view.
 - Perhaps `ghcide` should detect if the user has explicitly set the relevant `-fdefer` option, and in that case *not* promote the warnings back to errors. I think this a nice-to-have though, you could just open up a follow-up issue rather than doing it.
 - I'm a bit confused about how the various flags get set, but when I was experimenting to understand the qualified names issue, I found I didn't need to set the `Opt_Warn` flag to get the warnings. So perhaps that can be removed.

ndmitchell:
> I'm not sure about the merits of the current gyrations compared to just setting the option once-and-for-all in runGhcEnv (which is a lot simpler code-wise). @ndmitchell may have a view.

The gyrations are only ugly because Haskell records suck, but we should just fix that problem [at source](https://github.com/ndmitchell/record-dot-preprocessor/) 😝.

I prefer having the dynflag stuff localised where possible since it's just unclear what the impact is, and this way we can actually get at the original to see if the user had already set deferred type errors and thus avoid promoting them, whereas setting them in runGhcEnv makes that harder. However, I'm not against the other if people think it has benefits.

jacg:
> I think the three things we can/want to defer are
> 
> + type-errors
> + typed-holes
> + out-of-scope-variables
>
> I think at least typed-holes is implied by type-errors.

That hypothesis seems to be disproved by the commit entitled `TEST: Downgrade severity of typed holes Error -> Warning` which I have just pushed.

BTW, this branch is now rebased on top of #59, as the latter seems to be cruicial to its correctness.

jacg:
This seems to be working now, including the extra deferals proposed by @hsenag, so I think it's ready for review.

It does crucially depend on #59, therefore it is rebased on top of it.

I have gradually modified the tests and implementation in separate commits, in order to document which approaches work / don't work and which flags are needed / not need.

jacg:
> I have gradually modified the tests and implementation in separate commits, in order to document which approaches work / don't work and which flags are needed / not need.

Unfortunately Github presents the commits in timestamp order, rather than DAG order, thoroughly confusing the story I'm trying to tell!

hsenag:
> This seems to be working now

Excellent!

> Unfortunately Github presents the commits in timestamp order, rather than DAG order, thoroughly confusing the story I'm trying to tell!

Wouldn't you have needed to use merge commits to get an actual DAG? I just see a linear history in `gitk` as well.

I'd be inclined to squish it down to a single commit now (maybe that'll happen anyway when it's merged, I'm not too familiar with what github does). Having lots of logical commits is great during development to keep track of uncertain options, but is not necessarily so helpful when the feature is done. If there's something future readers of the code should understand, a comment is likely to be more discoverable than git history.

jacg:
> Wouldn't you have needed to use merge commits to get an actual DAG?

It's Directed, it's Acyclic, and it's a Graph, so it's a DAG! (A single vertex is a DAG ... even an empty graph is a DAG ... not very interesting ones, admittedly :-)

> I'd be inclined to squish it down to a single commit now

I'll leave that to the owners of this repo to decide. I've crafted the commits like this, so that the necessity, or otherwise, of various bits is easily verifiable.

ndmitchell:
At DA we always used squash merges so that you have a guarantee your history all passed all CI tests, which is useful for bisecting.

jacg:
It certainly was nice having it available in my editor while hacking on it!

jacg:
> At DA we always used squash merges so that you have a guarantee your history all passed all CI tests, which is useful for bisecting.

Haven't done it recently, but I recall having a workflow where `--first-parent` let you have the best of both worlds: easily bisectable master and detailed, documenting feature branches.

cocreature:
Thanks a lot for the contruption and sorry for my lack of response. I hope to get around to taking a closer look at this next week.

hsenag:
One thing I noticed after using this for a bit is that the text of the "error" still says warning:

```
C:\msys64\home\ganesh\darcs\wip\src\Darcs\Patch\Bundle.hs:181:5: warning:
```

Regardless of minor glitches like that, now I've been using it for a little while, I can also confirm that it is indeed awesome for improving the development experience :-)

jacg:
> the text of the "error" still says warning

Yes, I had also noticed this glitch while using it, but it seemed very much a second-order problem, and it slipped my mind. Maybe I'll manage to have a look at it and fix it before @cocreature gets around to reviewing it.

test/exe/Main.hs
```diff
@@ -83,6 +83,18 @@ diagnosticTests = testGroup "diagnostics"
...
           )
         ]
+  , testSession "typed hole" $ do
+      let content = T.unlines
+            [ "module Testing where"
+            , "foo :: Int -> String"
+            , "foo a = _ a"
+            ]
+      _ <- openDoc' "Testing.hs" "haskell" content
+      expectDiagnostics
```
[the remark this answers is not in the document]
cocreature: I guess you want to test that the undemoting works? Looks reasonable in that case but a comment would be good and it would still be nice to have a test that checks that actually relies on deferring the errors.

jacg:
> I would like this behavior to be configurable in IdeOptions so add a field there that allows you to turn this on/off.

Presumably you would want a single option that switches all three deferrals together ... or three different options, one for each deferral?

For completeness, the three things being deferred are:

1. type errors
2. typed holes
3. out of scope variables

If there are to be 3 separate ones, the names are fairly obvious:

1. `optDeferTypeErrors`
2. `optDeferTypedHoles`
3. `optDeferOutOfScopeVariables`

which closely follow similar names in `DynFlags`.

It's less obvious what the name would be if there is to be a single option controlling all three. `optDefer`? Otherwise the name starts getting veeeeery long indeed.

hsenag:
A single option could maybe be `optDeferErrors`. But actually from the point of view of the user, errors aren't really being deferred here. They're not getting any binary with actual embedded errors, they're just getting more diagnostics and better working intellisense in the IDE. So maybe `optContinueAfterErrors`?

@cocreature what's the motivation for having it be configurable? And what do you expect the default to be? (I'd argue for the feature to be on by default, I think it's something almost everyone would want.)

jacg:
> And what do you expect the default to be? (I'd argue for the feature to be on by default, I think it's something almost everyone would want.)

Te-he, I read this literally 3 seconds after having had exactly the same thoughts, and having written `optDefer = True` in the definition of `defaultIdeOptions`.

Yes, I want this on, and I think everybody would/should want it on!

cocreature:
Totally fine to have a single option and have it be on by default.

The reason why I want this to be configurable is that for DAML, our CLI compiler is also based on `ghcide` and I don’t want this enabled there. In addition to that, we also have rules in our IDE which depend on the typechecker output and do some kind of code generation to run tests automatically and display results and that will at the very least require changes to make it work with this so I’d like to just be able to turn it off for now.

Note that I’m only asking for a field in `IdeOptions`. I am totally fine with not adding a CLI flag to the `ghcide` executable to configure this.

jacg:
> I would like this behavior to be configurable in `IdeOptions` so add a field there that allows you to turn this on/off.

Can you suggest how best to do the plumbing between `IdeOptions` and the place where the deferal dance happens?

test/exe/Main.hs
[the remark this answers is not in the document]
jacg: Yes, the point is that the definition of this test (and a couple of others like it) evolves through this sequence of commits, and the implementation has to follow the tests through the Error -> Warning -> Error route. When all is said and done, we end up where we started (error) but these test document, in the sequence of commits, that each individual step did what it was supposed to; or, quite importantly in the case of the first attempt to defer type errors, that the obvious approach does NOT work, justifying the more complex implementation that ended up being used.

Anyway, I have added tests which detect if any of the deferrals are removed. These, of course, will have to be duplicated once I've added the deferral on/off switch.

cocreature:
The easiest option to get access to the options is probably to call `getIdeOptions` in `typeCheckRule` and then pass that as an argument to `typecheckModule`.

As for the tests, I don’t mind if we don’t test the non-defer codepath right now.

jacg:
If you don't need tests for it, then I have pushed something that seems to do the job. I have tested it informally by setting the default to `False` and observing the deferral tests fail, while they pass with the default set to `True`.

jacg:
Any preference as to the position in which the new parameter of `typecheckModule` should occupy?

src/Development/IDE/Core/Compile.hs
```diff
@@ -82,19 +82,22 @@ computePackageDeps env pkg = do
 
 -- | Typecheck a single module using the supplied dependencies and packages.
 typecheckModule
-    :: HscEnv
+    :: Bool
```
cocreature: Putting it as the first parameter is very reasonable but it would be great if we could get a newtype for this to get something a bit more informative than `Bool`.

src/Development/IDE/Types/Options.hs
```diff
@@ -44,9 +45,16 @@ data IdeOptions = IdeOptions
     -- ^ the ```language to use
   , optNewColonConvention :: Bool
     -- ^ whether to use new colon convention
+  , optDefer :: Bool
```
cocreature:
```suggestion
  , optDefer :: IdeDefer
```

src/Development/IDE/Types/Options.hs
```diff
@@ -63,6 +71,7 @@ defaultIdeOptions session = IdeOptions
     ,optReportProgress = IdeReportProgress False
     ,optLanguageSyntax = "haskell"
     ,optNewColonConvention = False
+    ,optDefer = True
```
cocreature:
```suggestion
    ,optDefer = IdeDefer True
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -312,7 +312,8 @@ typeCheckRule =
         tms <- uses_ TypeCheck (transitiveModuleDeps deps)
         setPriority priorityTypeCheck
         packageState <- hscEnv <$> use_ GhcSession file
-        liftIO $ typecheckModule packageState tms pm
+        IdeOptions{ optDefer = defer} <- getIdeOptions
+        liftIO $ typecheckModule (IdeDefer defer) packageState tms pm
```
cocreature:
```suggestion
        liftIO $ typecheckModule defer packageState tms pm
```

jacg:
Hmm, that was confusing: I think that your suggestions, which were almost identical to the changes I had just pushed, appeared here just after the push, so I was left wondering why on earth my changes were appearing as your suggestions.

jacg:
You're welcome. It's a pleasure. Thank *you* for `ghcide`; it's going to lead to a huge improvement in the Haskell development experience.

PR #48, opened 2019-09-12
Add support for -ddump-parsed,-ddump-parsed-ast and -dsource-stats

shayne-fletcher-da:
This PR adds code to `parseFileContents` to support syntax tree debug dumps.


shayne-fletcher-da:
> I guess you want this for DAML where we also use this code for the CLI compiler? In that case, I think it probably makes sense to do the dumping on the DAML side unless there is an IDE usecase for this.

Yes, that's right. It enables parse tree dumps using `damlc compile`. If it doesn't go here, where does it go then?

PR #55, opened 2019-09-13
fix type in README

2mol:


cocreature:
You’ll have to sign our CLA before we can merge this. (If you are worried about signing anything, I recommend taking a look at the actual text, it’s quite readable).

PR #59, opened 2019-09-13
Make sure warnings use unqualified names where appropriate

hsenag:
Because we are constructing the message objects ourselves, as opposed to error messages which are constructed by GHC, we need to take care to respect the passed-in `PprStyle`.

This should fix the problems @jacg was seeing in #47 (though I haven't double-checked with his specific code).


jacg:
> This should fix the problems @jacg was seeing in #47 (though I haven't double-checked with his specific code).

The last commit (at the time of writing) in #47 introduces a test that demonstrates the problem with fully qualified names, that is addressed by this PR. Rebasing #47 on top of this results in the test passing. So it seems to work.

[Here](/jacg/ghcide/tree/defer-te-unqualified) is a rebased copy of #47

PR #67, opened 2019-09-14
Add feature list

maralorn:
This is a certainly incomplete list of features already implemented. Please expand!


mitchellwrosen:
I'd also like to see the LSP name for the feature, if possible. For example, I use `coc.vim`, and I'm not sure what hotkey to add to get the quickfix-import-list thing..

PR #69, opened 2019-09-14
Code actions for filling typed holes

jacg:
Add code actions which replace typed holes, using GHC's "Relevant bindings include" hints.

### DIFF
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -132,7 +132,8 @@ suggestAction contents Diagnostic{_range=_range@Range{..},..}
     | "Valid hole fits include" `T.isInfixOf` _message = let
       findSuggestedHoleFits :: T.Text -> [T.Text]
       findSuggestedHoleFits = extractFitNames . selectLinesWithFits . dropPreceding . T.lines
-      proposeHoleFit name = ("replace hole with " <> name, [TextEdit _range name])
+      proposeHoleFit name = ("replace hole `" <> holeName <>  "` with " <> name, [TextEdit _range name])
+      holeName = T.strip $ last $ T.splitOn ":" $ head . T.splitOn "::" $ head $ filter ("Found hole" `T.isInfixOf`) $ T.lines _message
       dropPreceding       = dropWhile (not . ("Valid hole fits include" `T.isInfixOf`))
       selectLinesWithFits = filter ("::" `T.isInfixOf`)
       extractFitNames     = map (T.strip . head . T.splitOn " :: ")
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -22,6 +22,8 @@ import Data.Char
 import Data.Maybe
 import Data.List.Extra
 import qualified Data.Text as T
+import Data.Version
+import System.Info
 
 -- | Generate code actions.
 codeAction
@@ -129,17 +131,21 @@ suggestAction contents Diagnostic{_range=_range@Range{..},..}
 --         (imported from ‘Prelude’ at ...Development/IDE/LSP/CodeAction.hs:7:8-37
 --          (and originally defined in ‘GHC.Base’)) (lsp-ui)
 
-    | "Valid hole fits include" `T.isInfixOf` _message = let
+    | topOfHoleFitsMarker `T.isInfixOf` _message = let
       findSuggestedHoleFits :: T.Text -> [T.Text]
       findSuggestedHoleFits = extractFitNames . selectLinesWithFits . dropPreceding . T.lines
       proposeHoleFit name = ("replace hole `" <> holeName <>  "` with " <> name, [TextEdit _range name])
       holeName = T.strip $ last $ T.splitOn ":" $ head . T.splitOn "::" $ head $ filter ("Found hole" `T.isInfixOf`) $ T.lines _message
-      dropPreceding       = dropWhile (not . ("Valid hole fits include" `T.isInfixOf`))
+      dropPreceding       = dropWhile (not . (topOfHoleFitsMarker `T.isInfixOf`))
       selectLinesWithFits = filter ("::" `T.isInfixOf`)
       extractFitNames     = map (T.strip . head . T.splitOn " :: ")
       in map proposeHoleFit $ nubOrd $ findSuggestedHoleFits _message
 
 suggestAction _ _ = []
+
+topOfHoleFitsMarker = if versionBranch compilerVersion >= [8,6]
+                      then "Valid hole fits include"
+                      else "Valid substitutions include"
 
 mkRenameEdit :: Maybe T.Text -> Range -> T.Text -> TextEdit
 mkRenameEdit contents range name =
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -2,6 +2,8 @@
 -- SPDX-License-Identifier: Apache-2.0
 
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE CPP #-}
+#include "ghc-api-version.h"
 
 -- | Go to the definition of a variable.
 module Development.IDE.LSP.CodeAction
@@ -22,8 +24,6 @@ import Data.Char
 import Data.Maybe
 import Data.List.Extra
 import qualified Data.Text as T
-import Data.Version
-import System.Info
 
 -- | Generate code actions.
 codeAction
@@ -143,9 +143,12 @@ suggestAction contents Diagnostic{_range=_range@Range{..},..}
 
 suggestAction _ _ = []
 
-topOfHoleFitsMarker = if versionBranch compilerVersion >= [8,6]
-                      then "Valid hole fits include"
-                      else "Valid substitutions include"
+topOfHoleFitsMarker =
+#if MIN_GHC_API_VERSION(8,6,0)
+  "Valid hole fits include"
+#else
+  "Valid substitutions include"
+#endif
 
 mkRenameEdit :: Maybe T.Text -> Range -> T.Text -> TextEdit
 mkRenameEdit contents range name =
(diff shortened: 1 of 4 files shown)


mitchellwrosen:
Just curious, would this involve parsing those things out of GHC's stringy output?

src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -132,7 +132,8 @@ suggestAction contents Diagnostic{_range=_range@Range{..},..}
     | "Valid hole fits include" `T.isInfixOf` _message = let
       findSuggestedHoleFits :: T.Text -> [T.Text]
       findSuggestedHoleFits = extractFitNames . selectLinesWithFits . dropPreceding . T.lines
-      proposeHoleFit name = ("replace hole with " <> name, [TextEdit _range name])
+      proposeHoleFit name = ("replace hole `" <> holeName <>  "` with " <> name, [TextEdit _range name])
+      holeName = T.strip $ last $ T.splitOn ":" $ head . T.splitOn "::" $ head $ filter ("Found hole" `T.isInfixOf`) $ T.lines _message
```
jacg: I'm not at all proud of the pattern matching in places like `holeName` and surroundings. Any hints on how to do a better job of pattern matching on `Text` would be most welcome.

mitchellwrosen:
(oh, oops, this is a PR, not a feature request)

ndmitchell:
@mitchellwrosen yes, it involves parsing crap out of the GHC error message, which is sad, but effective. We turn on Unicode all the time to slightly reduce the amount of surface area we have to cover. We are also working with someone who is writing an error messages as ADT proposal, so as of GHC 8.12 or so we hope to avoid this string parsing.

jacg:
> We turn on Unicode all the time to slightly reduce the amount of surface area we have to cover.

Which probably explains the mismatch I had in quotes between the action name that I generate, and the action name that I have to search for in the test.

> We are also working with someone who is writing an error messages as ADT proposal, so as of GHC 8.12 or so we hope to avoid this string parsing.

So for 8.6 and 8.4 we're stuck with string parsing ... and I'm guessing that some subtle differences between the crap that 8.4 and 8.6 spit out are responsible for the CI failures that have shown up. So now I have to get `ghcide` working on the same project with two different GHC versions simultaneously ... I'm so looking forward to that!

jacg:
Given that parsing text is central to writing these actions, should we use some parser combinator library? If so, which one?

ndmitchell:
It's unclear if a parsing library makes it easier or harder. It probably makes it more demanding about the precise layout, which may not be what we want. But someone would need to try it to know.

jacg:
But at least there's no upfront objection to adding a parsing library as a dependency?

ndmitchell:
Not from me. If the code looks better or more robust after it sounds like a good idea. I believe attoparsec would be the "default" parser lib nowadays

jacg:
All the CI checks now pass, but the implementation simply comments out two tests which work on GHC 8.6 but fail on 8.4, because the latter doesn't provide hints about local bindings.

Is there a way of excluding these tests on GHC 8.4 only, without resorting to CPP?

ndmitchell:
Use https://hackage.haskell.org/package/base-4.12.0.0/docs/System-Info.html#v:compilerVersion probably together with https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Version.html#v:makeVersion

hsenag:
@ndmitchell btw, who's working on error messages as an ADT? I had some thoughts about it after digging through the GHC code a bit so would be interested to know what's planned.

ndmitchell:
@hsenag @bgamari and @alpmestan I think (but I might be wrong)

src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -102,13 +104,49 @@ suggestAction contents Diagnostic{_range=_range@Range{..},..}
...
+      dropPreceding       = dropWhile (not . (topOfHoleFitsMarker `T.isInfixOf`))
+      selectLinesWithFits = filter ("::" `T.isInfixOf`)
+      extractFitNames     = map (T.strip . head . T.splitOn " :: ")
+      in map proposeHoleFit $ nubOrd $ findSuggestedHoleFits _message
 
 suggestAction _ _ = []
 
+topOfHoleFitsMarker = if versionBranch compilerVersion >= [8,6]
```
cocreature: `versionBranch` isn’t going to work when compiling against `ghc-lib`. I’ve added a `MIN_GHC_API_VERSION` macro in #73  which should always be used for checking the GHC API version.

test/exe/Main.hs
```diff
@@ -453,6 +456,71 @@ importRenameActionTests = testGroup "import rename actions"
...
+          "globalConvert" "n" "n"
+
+  , check "replace hole `_b` with globalInt"
+          "_a" "_b"        "_c"
+          "_a" "globalInt" "_c"
+
+  , check "replace hole `_c` with globalInt"
+          "_a" "_b"        "_c"
+          "_a" "_b" "globalInt"
+  ] <> if versionBranch compilerVersion < [8,6] then [] else
```
cocreature: Same here, this should also use `MIN_GHC_API_VERSION`.

cocreature:
As for nicer ways of doing the string matching, I’m definitely not opposed to using a parser library (but I would prefer `megaparsec` over `attoparsec`). However, ime for these kind of grep-style tasks regex libraries actually work quite well and with `regex-applicative(-text)` we have a nice library that avoids the usual issue with regex quickly turning into a non-composable and unreadable mess.

Definitely not something for this PR but if someone wants to give it a shot I’d certainly be interested in whether it ends up being cleaner (I hope it does).

src/Development/IDE/LSP/CodeAction.hs
jacg: OK, but this requires CPP, and CI rejected my first attempt because it used CPP, which hlint didn't like. Do I have to permit CPP for this module somewhere?

cocreature: We have a whitelist of modules that are allowed to use `CPP` to make people a bit more cautious when using it. This is definitely a valid usecase, so you can just add your module to the whitelist

jacg: How do I add the test `Main` to this whitelist?

cocreature: I believe that `Main` should do the trick?

jacg: Just plain, unqualified `Main`? Tried that already: doesn't work ... unless I'm doing something stupid elsewhere, which is perfectly possible.

jacg: ... yes, I'm being stupid ...

cocreature: Hm not sure why this doesn’t work, maybe @ndmitchell has an idea?

jacg: I'm struggling to get CPP to work in `Main.hs` of the tests:

It works in `CodeAction.hs`, and I've done exactly the same in `Main.hs` AFAICT, that is:

Add 
````
{-# LANGUAGE CPP #-}
#include "ghc-api-version.h"
````
at the top, then 

````#if MIN_GHC_API_VERSION(8,6,0)````


 I did need to add `include-dirs: include` to `test-suite ghcide-tests` in `ghcide.cabal`, for the `#include` to work at all in `Main.hs'.

Now I get:

````
$ cabal v2-test -w ghc-8.6.4
Build profile: -w ghc-8.6.4 -O1
In order, the following will be built (use -v for more details):
 - ghcide-0.0.2 (test:ghcide-tests) (file test/exe/Main.hs changed)
Preprocessing test suite 'ghcide-tests' for ghcide-0.0.2..
Building test suite 'ghcide-tests' for ghcide-0.0.2..

test/exe/Main.hs:507:0: error:
     error: missing binary operator before token "("
     #if MIN_GHC_API_VERSION(8,6,0)
     
    |
507 | #if MIN_GHC_API_VERSION(8,6,0)
    | ^

test/exe/Main.hs:521:0: error:
     error: missing binary operator before token "("
     #if MIN_GHC_API_VERSION(8,6,0)
     
    |
521 | #if MIN_GHC_API_VERSION(8,6,0)
    | ^
`cc' failed in phase `C pre-processor'. (Exit code: 1)
````

What am I missing?

cocreature: Oh I see what’s going on here. We rely on the `MIN_VERSION` pragmas which only exist if you depend on `ghc` which the test suite does not atm. Not quite sure what a nice solution would be but for now, just adding a dependency on `ghc` is not that bad.

jacg: Thanks, that solves the problem.

Where shall I document the ghc-dependence in the tests? Comment in the cabal file itself?

cocreature: Yeah I think a comment in the cabal file should be sufficient, thank you!

jacg:
It seems to have passed CI, though I am confused by the hlint output, which includes stuff like:

````
Warning: Can't find file "ghc-api-version.h" in directories
	./src/Development/IDE/Core
	.
  Asked for by: ./src/Development/IDE/Core/Compile.hs  at line 6 col 1
Warning: Can't find file "ghc-api-version.h" in directories
	./src/Development/IDE/GHC
	.
  Asked for by: ./src/Development/IDE/GHC/CPP.hs  at line 12 col 1
Warning: Can't find file "ghc-api-version.h" in directories
	./src/Development/IDE/GHC
…	.
  Asked for by: ./test/exe/Main.hs  at line 6 col 1
````

cocreature:
Ah, it looks like we need to pass `--cpp-include` to HLint. We can do that in a separate PR.

alpmestan:
> @hsenag @bgamari and @alpmestan I think (but I might be wrong)

Correct, Ben and I have been discussing various approaches for this and I will in fact soon write a proposal for a design that we settled on with lots of useful inputs from Richard Eisenberg and Neil (Mitchell). One of its goals will be to make the life of IDE-style tooling devs a lot easier, essentially doing away with the need to parse GHC's messages, instead getting good old Haskell values to manipulate, inspect, use and print.

I can perhaps ping the ghcide team once I have something in a reviewable state, if some of you would like to give some feedback.

PR #80, opened 2019-09-17
Fix race condition in shakeRun

### LINKED ISSUE #79: Exception: Error when calling shakeRunDatabase
We sometimes seem to see error messages like the following one
```
Exception: Error when calling shakeRunDatabase, currently running:
  Existing call:  shakeRunDatabase
```
This appears to be triggered by codelens on DAML codebase requests but I highly doubt that it’ specific to this so opening this here rather than on the `daml` repository.

I haven’t managed to reproduce it so far but I have a suspicion on what is going on:

1. We have an existing call A to `shakeRun`.
2. Another call B to `shakeRun` kills the first call in https://github.com/digital-asset/ghcide/blob/eba0185710042976004467d6a8fdb20d8ed6cc57/src/Development/IDE/Core/Shake.hs#L350.
3. The new call spawns a new thread that calls `shakeRunDatabase`.
4. Now a third call C to `shakeRun` interrupts us after spawning the thread but before setting the `MVar` to the new stop action, so somewhere here https://github.com/digital-asset/ghcide/blob/eba0185710042976004467d6a8fdb20d8ed6cc57/src/Development/IDE/Core/Shake.hs#L362
5. If an exception is thrown in `withMVar` the original contents will be restored so our stop action now points to the previous action for stopping A which is already dead.
6. When C calls `duration stop` it will therefore continue immediately without waiting for `C` to die and we run into the issue described by the error message.

@ndmitchell Does that sound plausible to you? If so, it seems like we can fix this by masking between the `forkFinally` and writing the new stop action to the `MVar`.

cocreature:
fixes #79 which also contains a detailed description of the issue.

cc @ndmitchell

### DIFF
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -342,31 +342,41 @@ shakeShut IdeState{..} = withMVar shakeAbort $ \stop -> do
     stop
     shakeClose
 
+-- | This is a variant of withMVar where the first argument is run unmasked and if it throws
+-- an exception, the previous value is restored while the second argument is executed masked.
+withMVar' :: MVar a -> (a -> IO b) -> (b -> IO (a, c)) -> IO c
+withMVar' var unmasked masked = mask $ \restore -> do
+    a <- takeMVar var
+    b <- restore (unmasked a) `onException` putMVar var a
+    (a', c) <- masked b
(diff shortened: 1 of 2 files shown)


cocreature:
Good idea! I’ve factored it out into `withMVar'` for lack of a better name.

PR #81, opened 2019-09-17
Add code actions for missing type signatures

### LINKED ISSUE #77: Code action: Add missing signatures
With #69 merged it should now be possible to offer a code action to deal with the missing signatures warning.

The most simple approach would be to just generate a signature with a typed hole and then let the user use the code action on that one.

A more sophisticated approach would be to fake the typed hole to ghc and then supply the response directly in the code action.

The longer I think about it the better I like the first option, since it gives the user the ability to write an own signature if wished, the other solution actually sounds a bit complex to implement.

jacg:
Will need more thorough testing, and maybe more discussion of the different approaches considered in #77.

### DIFF
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -141,6 +141,16 @@ suggestAction contents Diagnostic{_range=_range@Range{..},..}
       extractFitNames     = map (T.strip . head . T.splitOn " :: ")
       in map proposeHoleFit $ nubOrd $ findSuggestedHoleFits _message
 
+    | "Top-level binding with no type signature" `T.isInfixOf` _message = let
+      filterNewlines = T.concat  . T.lines
+      unifySpaces    = T.unwords . T.words
+      signature      = T.strip $ unifySpaces $ last $ T.splitOn "type signature: " $ filterNewlines _message
+      startOfLine    = Position (_line _start) 0
+      beforeLine     = Range startOfLine startOfLine
+      title          = "add signature: " <> signature
+      action         = TextEdit beforeLine $ signature <> "\n"
+      in [(title, [action])]
+
 suggestAction _ _ = []
 
 topOfHoleFitsMarker :: T.Text
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -272,6 +272,7 @@ codeActionTests = testGroup "code actions"
   , removeImportTests
   , importRenameActionTests
   , fillTypedHoleTests
+  , addSigActionTests
   ]
 
 renameActionTests :: TestTree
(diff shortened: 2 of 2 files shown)


jacg:
One problem that I am observing locally, is that if I apply one of these actions in Emacs in the context of a larger file, then the signatures are sometimes garbled. As I seem to have lost the ability to apply code actions in VSCode, I can't tell whether this is an Emacs problem, or a problem with the implementation.

jacg:
> the signatures ar sometimes garbled

For example, adding the line `foo = True` just after the imports in `ghcide/src/Development/IDE/LSP/CodeAction.hs` and running the suggested action, results in 

````
foo :: Bool
foo = True
````

If the same line is placed at the end of the same file, then the code action results in

````
fooo::oBool
foo = True
````

Similar story with `abcdefgh a b c = True` which gives `abcdefgh :: p1 -> p2 -> p3 -> Bool` or `abcdefgh ::ap1b->cp2d-> p3 -> Bool`.

**Comment by jacg** [comment]
The LSP log for the last one shows

````
  "contentChanges": [
    {
      "range": {
        "start": {
          "line": 238,
          "character": 0
        },
        "end": {
          "line": 238,
          "character": 8
        }
      },
      "rangeLength": 8,
      "text": "abcdefgh ::bp1c->dp2e-> p3 -> Bool\nabcdefgh"
    }
  ]
````

Which suggests that it's the server which is sending this nonsense?

**Comment by jacg** [comment]
Managed to get code actions working in VS Code once more: it all looks good over there, so this seems to be a problem with Emacs rather than this implementation.

**Comment by jacg** [comment]
The big question that remains is what to do about enabling `-Wmissing-signatures`, without which none of this works.

The tests force the issue with `{-# OPTIONS_GHC -Wmissing-signatures #-}`.

Over in #77, @ndmitchell suggested

> Maybe ghcide should always turn on this warning, always suggest it as a code action, but then if the user didn't turn on this warning, suppress the resulting warning notification.

Should I try doing it this way, or are there alternative proposals?

**Comment by jacg** [comment]
... or should we just merge it as is, and think deal with enabling the warnings in a separate PR?

**Comment by ndmitchell** [comment]
I think this is a great first step for people who enable this warning.

My ideal end state for everyone who doesn't enable the warning would be to have these quick fixes available, but not see even an info level diagnostic for missing signatures. I'm not sure if you can attach a quick fix without a diagnostic. The other option would be that top-level signatures that are omitted show up as code lenses, so you see:

```
foo :: Int -> Int
foo 1 =1
```

But the `foo :: Int -> Int` line is a hyperlink that's useful documentation, but if you click it, inserts the file for real. I could definitely see myself using that.

jacg:
> I think this is a great first step for people who enable this warning.

In which case I propose that we merge this as is, and leave further considerations for another PR.

jacg:
But as it stands, it doesn't fully address #77, so I've removed the automatic closing.

ndmitchell:
See #88 for how we should be doing inferred signatures if the warning is off, using the proper things in VS Code.

maralorn:
Fwiw, I think. I would appreciate merging this and caring about the case without the warning enabled later.
That fits together with neils suggestion.

PR #106, opened 2019-09-20
Add --version CLI option

### LINKED ISSUE #103: No --version flag and version information is not precise
```
[nix-shell:~/ghc]$ ghcide --version
Starting ghcide (GHC v8.6)
Invalid option `--version'

Usage: ghcide [--lsp] [--cwd DIR] [FILES/DIRS...]
  Used as a test bed to check your IDE will work
```

1. The version does not mention the minor version number, the version of `ghcide` needs to match precisely the compiler version used by the project.
2. The `--version` flag doesn't exist in the first place to nicely print out this version.

jacg:
- [x] where can more precise GHC version information be found?
- [x] where is the single source of wisdom on `ghcide` version?

Closes #103.

### DIFF
--- a/exe/Arguments.hs
+++ b/exe/Arguments.hs
@@ -10,6 +10,7 @@ data Arguments = Arguments
     {argLSP :: Bool
     ,argsCwd :: Maybe FilePath
     ,argFiles :: [FilePath]
+    ,argsVersion :: Bool
     }
 
 getArguments :: IO Arguments
@@ -25,3 +26,4 @@ arguments = Arguments
       <$> switch (long "lsp" <> help "Start talking to an LSP server")
       <*> optional (strOption $ long "cwd" <> metavar "DIR" <> help "Change to this directory")
       <*> many (argument str (metavar "FILES/DIRS..."))
+      <*> switch (long "version" <> help "Show ghcide and GHC versions")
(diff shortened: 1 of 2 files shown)


ndmitchell:
See https://downloads.haskell.org/~ghc/7.8.3/docs/html/users_guide/options-phases.html 4.12.3 for the things GHC gives you to get the version. Unfortunately it doesn't have enough info. However, there is a base version which forms a 1:1 mapping with GHC version. https://wiki.haskell.org/Base_package#Versions has a partial map, but someone seems to have forgotten to define it recently. 

I suggest we add System.Info.Extra.compilerVersionFull :: Version which gives us what we need. We then propose that for addition to GHC at some point. We would build it using base versions. It's a pain since it needs updating on every point release.

jacg:
So should we just merge this as is, to provide `--version`, and leave the more precise GHC version info for a separate PR?

ndmitchell:
Makes sense, yep.

ndmitchell:
I looked into adding precise GHC version in `extra`, but I only have the `base` version available, and that's insufficient as there are multiple GHC versions with the same `base`.

However, Cabal predefines:

```
#define VERSION_ghc "8.6.5"
```

So just using CPP should be sufficient.

PR #114, opened 2019-09-23
Use a consistent include dir for cwd

cocreature:
This only matters for the DAML codebase (so I’ll add a test on that
side) where we use relative paths:

Previously, we would produce the include dir "." for moduleImportPath
"./A.daml"
and "" for moduleImportPath "./A/B.daml". This resulted in us ending
up with ./A.daml and A.daml in the Shake graph which resulted in
issues like https://github.com/digital-asset/daml/issues/2929.

We should move this logic completely over to the DAML repo at some
point but I’ll leave that for a separate PR.


cocreature:
Not in `ghcide`, the test in in https://github.com/digital-asset/daml/pull/2978

PR #119, opened 2019-09-23
No warnings

ndmitchell:
Warnings are bad. Delete the 3 that got introduced, then fix the Azure pipelines, which was previously checking that the dependencies had no warnings (don't care), but not that the build had no errors (care a lot).


ndmitchell:
Depends on #117, since there were too many warnings floating around.

ndmitchell:
Rebased on master, so hopefully this now passes.

cocreature:
Thanks again!

PR #120, opened 2019-09-23
Fix deferred messages

jacg:
The text of the error messages of deferred type errors / typed holes / out of scope variables, refer to them as warnings, even after they have been upgraded back to errors. This PR fixes these messages.


jacg:
There's bound to be a better way to replace 'warning' with 'error' in `Text`: suggestions welcome.

Furthermore, in theory, the message could contain a second ocurrence of the substring ": warning:" elsewhere in the message, which should *not* be changed to ": error:", but would be by this implementation: only the occurrence on the end of the first line of the message should be replaced. I'm not sure what would be a clean way to do this.

PR #121, opened 2019-09-24
Use a stack cradle in hie.yaml

cocreature:
I realize that @ndmitchell likes the direct cradle but it makes for a terrible default imho and I really don’t want to recommend that people start messing around with their global package db (I’m not even willing to do that myself).


ndmitchell:
Note that with this new cradle startup time increases (I'll live with that) and 2 modules in test fail to load (which sucks, but again, liveable). I'm not too unhappy, given the reduction in complexity.

hsenag:
FYI the direct cradle worked well for me even though I use cabal v2-install. cabal writes out .ghc-environment files that makes it work even without the global DB. Stack is a bad default for me so I'll need to override it locally.

ndmitchell:
I wonder if there should be a convention that hie.yaml.user is always used in preference to hie.yaml if it exists, and that we put hie.yaml.user in the gitignore - that would make situations like the above more palatable.

hsenag:
It might also be useful to continue maintaining the direct yaml file in the repo under a different name.

PR #127, opened 2019-09-24
Support multiple hie.yaml files

ndmitchell:
Each file looks up its `hie.yaml` separately, and if they match, they share the same `hie.yaml`. I encountered a few bugs around the place on the way here, links to the upstream issues included.

I also rewrote the user setup/debug logic for `ghcide` on its own, to be based around the idea that a directory may have multiple hie.yaml files, and also to print out less stuff that might confuse users.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -39,6 +39,7 @@ import System.Exit
 import Paths_ghcide
 import Development.Shake hiding (Env)
 import qualified Data.Set as Set
+import qualified Data.Map as Map
 
 import GHC hiding (def)
 import qualified GHC.Paths
@@ -64,13 +65,12 @@ main = do
 
     -- lock to avoid overlapping output on stdout
     lock <- newLock
-    let logger = Logger $ \pri msg -> withLock lock $
+    let logger p = Logger $ \pri msg -> when (pri >= p) $ withLock lock $
             T.putStrLn $ T.pack ("[" ++ upper (show pri) ++ "] ") <> msg
 
     whenJust argsCwd setCurrentDirectory
 
     dir <- getCurrentDirectory
-    hPutStrLn stderr dir
 
     if argLSP then do
         t <- offsetTime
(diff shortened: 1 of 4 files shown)


exe/Main.hs
```diff
@@ -147,15 +160,35 @@ newSession' cradle = do
...
+
+loadSession :: FilePath -> IO (FilePath -> Action HscEnvEq)
+loadSession dir = do
+    cradleLoc <- memoIO $ \v -> do
+        res <- findCradle v
+        -- Sometimes we get C: and sometimes we get c:, try and normalise that
+        -- e.g. see https://github.com/digital-asset/ghcide/issues/126
+        case res of
+            Nothing -> return Nothing
+            Just res -> Just <$> canonicalizePath res
```
cocreature: Do we really want to resolve symlinks? It isn’t clear to me that this is the right thing to do here. If you just want to normalise drive letters `normalise` should do the job.

exe/Main.hs
```diff
@@ -147,15 +160,35 @@ newSession' cradle = do
...
+--
+--   * If multiple people ask for a result simultaneously, make sure you only compute it once.
+--
+--   * If there are exceptions, repeatedly reraise them.
+--
+--   * If the caller is aborted (async exception) finish computing it anyway.
+memoIO :: Ord a => (a -> IO b) -> IO (a -> IO b)
+memoIO op = do
+    ref <- newVar Map.empty
+    return $ \k -> join $ uninterruptibleMask_ $ modifyVar ref $ \mp ->
```
cocreature: Why do we need an `uninterruptibleMask_` here? Afaict, the only interruptible operation is the `takeMVar ref` that is part of `modifyVar ref` and it seems fine to be interrupted here?

exe/Main.hs
```diff
@@ -39,6 +39,7 @@ import System.Exit
 import Paths_ghcide
 import Development.Shake hiding (Env)
 import qualified Data.Set as Set
+import qualified Data.Map as Map
```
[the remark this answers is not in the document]
ndmitchell: We already use containers.Data.Set in Main, but have no dependency on unordered-containers, so I kept with consistency. I note that `setFilesOfInterest` takes a `Set` in its API. I generally view using both `containers` and `unordered-containers` within on project as a code smell - just like use of both `transformers` and `mtl` (which we also do).

exe/Main.hs
ndmitchell: We almost certainly don't - at one point I was getting weird results and just started upping the masking levels. Then it turned out to be `C` vs `c` and totally not masking

exe/Main.hs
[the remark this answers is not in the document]
cocreature: I’m confused `Data.Map.Strict` is not part of `unordered-containers`?

exe/Main.hs
ndmitchell: Yep, that seems sufficient. Although once we normalise drive letters, we can probably avoid it entirely.

ndmitchell:
Changes addressed.

exe/Main.hs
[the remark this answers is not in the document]
ndmitchell: Sorry, I misread as Data.HashMap.Strict. Fixed

PR #133, opened 2019-09-25
Get rid of mtl dependency

ndmitchell:
In some cases we were using it to get access to functions like `void`. Everything was in `transformers` already.

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -42,7 +42,6 @@ library
         hashable,
         haskell-lsp-types,
         haskell-lsp >= 0.15,
-        mtl,
         network-uri,
         prettyprinter-ansi-terminal,
         prettyprinter-ansi-terminal,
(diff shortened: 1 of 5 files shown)


ndmitchell:
I have no opposition to mtl per say, but if you are also using transformers then it's mostly just confusing having two places for the same stuff.

cocreature:
The CI failures on 8.4 look genuine.

ndmitchell:
The type of `except` got generalised in transformers-0.5.6.0. We rely on that generalisation. You can't depend on a newer `transformers` because `ghc` also depends on it, so you get missing instances. We have two choices:

* Keep `mtl`, at least until we drop 8.4 support. Most of the cleanups are still useful, so can be broken out.
* Inline the definition of `except` in the 1 place it is used. The definition is `except = ExceptT . return` so the code bloat isn't huge.

cocreature:
I would just keep the dependency on `mtl`.

PR #134, opened 2019-09-25
Point to the VSCode marketplace in the readme

cocreature:


domenkozar:
I think it's better to split up user documentation from developer documentation.

Installing the extension from source should be in "hacking" section or similar.

cocreature:
Good idea, moved it to a separate seciton.

Issue #137, opened 2019-09-26
VSCode and nix-shell


chris-martin:
When using the VSCode plugin, VSCode is what launches the ghcide server, right? How do I get it to launch it in my project's nix shell so that the dependencies and things will be available?

ndmitchell:
In the extension you can set the command it runs and the arguments it passes. By default that's `ghcide` and `--lsp`, but you can change it to things like `stack` and `exec -- ghcide --lsp` to get it working with stack. The other option is to run `stack exec code` to launch VS Code inside the stack environment. I don't know how to do it with nix-shell, but I imagine it's something similar.

domenkozar:
@chris-martin try launching VSCode from nix-shell.

chris-martin:
@ndmitchell Can I put quotes in the "arguments" field? Like if I set the executable path to `nix-shell` and the arguments to `--run 'ghcide --lsp'`?

Or, can the executable path be relative to the project directory, so I can put the command to run ghcide in a bash script, set the executable path to `./ghcide.sh`, and the arguments to nothing?

chris-martin:
@domenkozar That doesn't seem to help either. ghcid works in this project, but I still have red squiggles under all the imports in vscode, even when I launch it from the nix shell.

![Screenshot from 2019-09-26 11-49-12](https://user-images.githubusercontent.com/399718/65712211-ba0f8f00-e053-11e9-9a43-75abebe61529.png)

chris-martin:
Nevermind, I don't think this is an editor integration issue. If I just run "ghcide" on the command line inside the nix shell, it can't see any of the packages either. Just a whole lot of this:

```
Message:
  Could not find module ‘Path’
  It is not a module in the current program, or in any known package.
    haskell/typeclasses/src/Typeclasses/SendingEmail/Daemon.hs
Range:    48:17-48:37
Source:   not found
Severity: DsError
Message:
  Could not find module ‘Network.Api.Postmark’
  It is not a module in the current program, or in any known package.
    haskell/typeclasses/src/Typeclasses/SendingEmail/Daemon.hs
Range:    51:17-51:37
Source:   not found
Severity: DsError
Message:
  Could not find module ‘Network.Api.Postmark’
  It is not a module in the current program, or in any known package.
    haskell/typeclasses/src/Typeclasses/SendingEmail/Daemon.hs
Range:    54:7-54:29
Source:   not found
Severity: DsError
Message:
[...]

PR #138, opened 2019-09-26
#137, make the status message clearer about what worked and what didn't

ndmitchell:
See #137 where it wasn't clear because the list of failing files pushed them off the edge.


chris-martin:
:+1: This would have spared me a good bit of initial confusion!

Issue #71, opened 2019-09-16
Go to definition doesn't work for record syntax


newhoggy:
Doesn't work for constructor

![image](https://user-images.githubusercontent.com/63014/64929940-3e515d80-d86f-11e9-8f57-35f925335529.png)

Doesn't work for fields

![image](https://user-images.githubusercontent.com/63014/64929942-414c4e00-d86f-11e9-85db-8b62dba23f74.png)

cocreature:
Thanks for the bug report, that definitely sounds like something that we should fix. If you want to take a look yourself, my guess would be that [getSrcSpanInfos](https://github.com/digital-asset/ghcide/blob/7133f4192f8d11908e5fb4c3f52ca3af54a23459/src/Development/IDE/Spans/Calculate.hs#L45) is the culprit.

jacg:
It seems to work fine here:

![constructor](https://user-images.githubusercontent.com/2570854/65230313-ee240680-dacd-11e9-8442-56cdefa3cd63.png)

... both for constructors and fields (though for fields it still refers to the position of the constructor).

Can you provide a minimal failing example?

newhoggy:
I defined `person0` and `person1`.

The only time I get the popup is when I hover over the `Person` constructor in `person1`.

![image](https://user-images.githubusercontent.com/63014/65380614-6a168c80-dd22-11e9-9050-f170f1b6edf9.png)

And Go To Definition doesn't work anywhere in the `person0` case.

![image](https://user-images.githubusercontent.com/63014/65380622-7c90c600-dd22-11e9-9343-662be129d800.png)

![image](https://user-images.githubusercontent.com/63014/65380628-8dd9d280-dd22-11e9-9581-46315c8499f3.png)

ndmitchell:
I get definitions for the `Person` constructor in both `person0` (due to https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Spans/Calculate.hs#L122) and `person1` (due to https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Spans/Calculate.hs#L121). I don't get information about `name` or `age` because it doesn't have any cases that examine the record fields. This whole piece of code was copied wholesale from Intero, upgraded through a few GHC versions, lightly hacked for "things that didn't work", and could do with someone giving it some love. However, it shouldn't be that hard to make it a lot more powerful. If using SYB was OK then finding all names and grabbing the nearest enclosing Loc would probably be perfect.

[...]

Issue #70, opened 2019-09-14
Type information for operators and arbitrary portions of code


jacg:
`ghcide` gives me information on most alphanumeric symbols (and numeric literals) but there are two features that I'm used to from `dante` and/or `itero`:

+ type information on operators
+ type information on arbitrary selected expressions

Do these features exist, and have I simply not found them?

If not, any hints on how to implement them?

mitchellwrosen:
I'm able to see type information on operators.
![2019-09-14-221120_723x423_scrot](https://user-images.githubusercontent.com/1074598/64915754-9e93b280-d73c-11e9-90d8-c3286e6aeaa6.png)

cocreature:
Operators should definitely work, given that it seems to work in some cases, it would be good if you could provide a specific test case where it fails.

Getting type information of a range is a bit more tricky. It is obviously more complex to implement but there is a second issue here: LSP hover does only give us a position not a range. I’ve taken a quick look at the spec and I also couldn’t find any other method that we could use here either.

jacg:
> Operators should definitely work

Sorry, yes, they do (it's just that in my configuration symbols' types were being shown spontaneously, while it took a bit more effort for symbols).

jacg:
I guess that the `bug` label isn't appropriate here, as everything that is expected to work (information about symbols) seems to work.

jneira:
The type of arbitrary expression can be extracted indirectly in hls, using the eval plugin, that includes `:t`.
Not sure if it is possible right now to translate in the hover, using the lsp protocol as pointed by @cocreature. So we should wait for some spec update to implement it.

@jacq does the workaround using eval works for you?

JacoboDominguez:
That question should be addressed by me nearly twin @jacg :)
Cheers,

alanz:
FYI https://github.com/microsoft/language-server-protocol/issues/377

normenmueller:
> The type of arbitrary expression can be extracted indirectly in hls, using the eval plugin, that includes :t.

@jneira Could you please give some more advice on how to install/ configure?

jneira:
@normenmueller you have to install [haskell-language-server](https://github.com/haskell/haskell-language-server) which includes by default the eval plugin (you can see it in accion [here ](https://github.com/haskell/haskell-language-server#features)).

Then you can write `-- >>>` in the code to evaluate expressions in the module context:

```
-- >>> :t fooExe
fooExe :: [Char]
fooExe = "hi"
```

[...]

PR #139, opened 2019-09-27
Add tests for find-definition and hover

jacg:
These tests document what works or doesn't work in terms of hover and find definition. Tests corresponding to features which do not work are marked as expected to fail, allowing the test suite as a whole to pass.

The current state of play is:

| | find definition | hover |
|-------------|-----------|---------|
| field in record definition | :heavy_check_mark:  | :heavy_check_mark: |
| field in record construction | :exclamation:  | :exclamation: |
| field name as accessor | :heavy_check_mark: | :heavy_check_mark: |
| top level name | :heavy_check_mark: | :heavy_check_mark: |
| record data constructor | :exclamation: | :exclamation:  |
| plain data constructor | :heavy_check_mark: | :heavy_check_mark:  |
| type constructor | :heavy_check_mark: | :exclamation: |
| plain parameter | :heavy_check_mark: | :heavy_check_mark: |
| pattern match name | :heavy_check_mark: | :heavy_check_mark: |
| top level operator | :heavy_check_mark: | :heavy_check_mark: |
| parameter operator | :heavy_check_mark:  | :heavy_check_mark:  |
| name in backticks | :heavy_check_mark: | :heavy_check_mark: |

Curiously, hover over type constructors finds no information, while go-to-definition works. In all other cases tested here, either both hover and go-to-definition work, or neither of them does.

Plenty more test cases can be found, but this is a start. Hopefully the existence of these tests can help focus discussions such as those in #71 and #70, as well as assisting anybody working on fixing, improving or refactoring the relevant code.

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -648,7 +648,7 @@ findDefinitionTests = let
   aaa    = mkRange   6  0    6  3
   vv     = mkRange  15  4   15  6
   op     = mkRange  16  2   16  4
-  opp    = mkRange  17 13   17 15
+  opp    = mkRange  17 13   17 17
   apmp   = mkRange  17 10   17 11
   bp     = mkRange  18  6   18  7
   -- search locations
@@ -661,7 +661,7 @@ findDefinitionTests = let
   tcL5   = Position  5 11
   vvL15  = Position 15 12
   opL15  = Position 15 15
-  opL17  = Position 17 21
+  opL17  = Position 17 22
   aL17   = Position 17 20
   b'L18  = Position 18 13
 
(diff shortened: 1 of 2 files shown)


test/exe/Main.hs
```diff
@@ -585,6 +586,123 @@ addSigActionTests = let
...
+           liftIO $ expected @=? foundRange
+
+  checkHover hover expected = do
+    case hover of
+      Nothing -> liftIO $ "hover found" @=? ("no hover found" :: T.Text)
+      Just Hover{_contents = (HoverContents MarkupContent{_value = v})} ->
+        liftIO $ adjust expected @=? Position l c where
+          found = T.splitOn ":" $ head $ T.splitOn "**" $ last $ T.splitOn "Testing.hs:" v
+          [l,c] =  map (read . T.unpack) found
+          -- looks like hovers use 1-based numbering while definitions use 0-based
```
cocreature: good catch, that sounds like a bug.

jacg: Which do we prefer, 0 or 1 based?

cocreature: LSP expects both lines and columns to be 0-based so we should follow that.

jacg: These 1-based numbers come from GHC itself, specifically [`SrcLoc`](https://hackage.haskell.org/package/ghc-8.6.5/docs/src/SrcLoc.html#RealSrcLoc).

They only appear in the hover message text, rather than being used programatically anywhere (except for these tests), so I wonder to what extent it is important.

PR #141, opened 2019-09-29
Ignore the stack.yaml.lock files

ndmitchell:


jhrcek:
The stack team recommends checking the lock files into version control: https://docs.haskellstack.org/en/stable/lock_files/

Please don't think of this as me opposing this change. I'd just like to understand the motivation behind git-ignoring the lock files in this project :-)

PR #150, opened 2019-10-03
Signal code actions in initialize response

jacg:
Following on from the conversation in #147 ...


jacg:
This resolves the problem of missing code actions from the perspective of Emacs/lsp-ui (https://github.com/emacs-lsp/lsp-ui/issues/319), so I am confident that this change had some positive effect, as I had something to test against beyond the internal tests in #147.

But there are lots of other blanks in those tests. Which of those are obviously just as wrong as the code action blank? Can you think of any external verification mechanism for any of them?

cocreature:
The other ones look fine to me but I might be missing something. Not sure how to verify them.

Issue #126, opened 2019-09-24
Drive letters are not consistent on Windows


ndmitchell:
I see  both `c:\\` and `C:\\`, which i'm sure screws up the cache and makes everything load twice. Do we normalise them? If so, to which? FWIW I see `c:\\` as the initial load, seemingly coming out of VS Code, and then `C:\\` for our recursive imports.

I'm seeing this when trying to load a session by creating a separate session for each file, which I hope will all resolve to the same session. The difference between drive letters stops that happening.

cocreature:
We’ve never normalised drive letters. The only normalisation that we perform at the moment is normalising file path slashes.

ndmitchell:
It seems there are some nuances to doing it there: https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Types/Location.hs#L54-L56

I can't imagine it's not causing us issues somewhere, perhaps double loading on Windows.

cocreature:
That code explicitly does not normalise drive letters. IIRC the issue was that if we do try to normalize them without normalizing them in `haskell-lsp` we get VFS lookup failures. So this needs to change in both `NormalizedUri` in `haskell-lsp` and here or we’ll run into issues.

ndmitchell:
Did this get fixed in #152 ? Or are there more issues here?

cocreature:
This should be fixed by the `haskell-lsp` upgrade which fully normalises file paths and #266 on our side.

PR #152, opened 2019-10-03
Avoid file path normalization in moduleImportPath

cocreature:
This fixes some issues where we used an uppercase drive letter in the
import path even though the LSP client uses lowercase drive letters


cocreature:
This should help a bit with #126

PR #153, opened 2019-10-04
Fix progress reporting

cocreature:
This is only a draft since my patch to `haskell-lsp` isn’t finished yet but it seems to work in my very limited testing.

### DIFF
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -28,12 +28,13 @@ export function activate(context: ExtensionContext) {
 		// Register the server for plain text documents
 		documentSelector: ["haskell"]
 	};
-	let client2 = new LanguageClient( 
+	client = new LanguageClient(
 		'haskell',
 		'Haskell IDE Core',
 		{ args: args, command: cPath, options: {cwd: workspace.rootPath }}, clientOptions, true);
+    client.registerProposedFeatures();
 	
-	client2.start();
+	client.start();
 }
 
 export function deactivate(): Thenable<void> | undefined {
--- a/extension/src/extension.ts
+++ b/extension/src/extension.ts
@@ -30,7 +30,7 @@ export function activate(context: ExtensionContext) {
 	};
 	client = new LanguageClient(
 		'haskell',
-		'Haskell IDE Core',
+		'ghcide',
 		{ args: args, command: cPath, options: {cwd: workspace.rootPath }}, clientOptions, true);
     client.registerProposedFeatures();
 	
--- a/src/Development/IDE/LSP/LanguageServer.hs
+++ b/src/Development/IDE/LSP/LanguageServer.hs
@@ -41,7 +41,7 @@ import Language.Haskell.LSP.Messages
 runLanguageServer
     :: LSP.Options
     -> PartialHandlers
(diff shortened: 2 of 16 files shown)


src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -107,6 +109,8 @@ runLanguageServer options userHandlers getIdeState = do
             , LSP.onStartup = handleInit (signalBarrier clientMsgBarrier ()) clearReqId waitForCancel clientMsgChan
             }
 
+    LSP.setupLogger (Just "/tmp/lsp-hello.log") [] L.DEBUG
```
ndmitchell: I assume this is the code that makes it draft?

cocreature: That’s part of it but the main reason is that the changes to `haskell-lsp` aren’t in a mergable state as I only implemented the part that we rely on.

Anrock: I hope you wouldn't use hardcoded path after merge. On Windows it may produce some strange behaviour, see [similar situation here](https://github.com/baverman/vial/issues/7)

cocreature: I definitely won’t, there is a reason this is a draft PR :slightly_smiling_face:

extension/src/extension.ts
```diff
@@ -28,12 +28,13 @@ export function activate(context: ExtensionContext) {
 		// Register the server for plain text documents
 		documentSelector: ["haskell"]
 	};
-	let client2 = new LanguageClient( 
+	client = new LanguageClient(
 		'haskell',
 		'Haskell IDE Core',
```
domenkozar: Shouldn't this be ghcide?

cocreature: good catch, that’s leftover from before the renaming.

domenkozar:
@cocreature I'd love to give this one a try, but it needs a bump of haskell-lsp commit in stack.yaml

cocreature:
@domenkozar Just pushed a PR that should update that.

I’ve also now cleaned up the PRs to `haskell-lsp` and patched `lsp-test` so once https://github.com/alanz/haskell-lsp/pull/188, https://github.com/alanz/haskell-lsp/pull/189 and https://github.com/bubba/lsp-test/pull/54 have been merged this can go in.

domenkozar:
@cocreature did you test this on vscode, I don't see any process indication (using the latest 1.39.1)

domenkozar:
Ah, it also needs an update of the extension.

domenkozar:
Doesn't seem to be supported in vscode yet: https://github.com/microsoft/vscode-languageserver-node/issues/528#issuecomment-541600565

cocreature:
I’ve tested this in VSCode yesterday and it worked fine. We don’t rely on the fields mentioned in that issue.

domenkozar:
@cocreature can you describe what I'm supposed to see with this PR?

cocreature:
Screenshot from 2019-10-15 14-40-56
Here is a screenshot, the processing thing at the bottom left is new.

domenkozar:
Thanks! Seems like I had to uninstall the stable extension, quit vscode, install the development extension, then launch vscode. Now I see the progress information as well!

Issue #160, opened 2019-10-08
--lsp hangs in nix-shell


willisplummer:
I've been running into an issue where it seems that `ghcide --lsp` has totally hung in a relatively small project that I've been setting up with nix

https://github.com/willisplummer/haskell-nix

It works if I run `ghcide` without the `--lsp` flag.

I'm at a loss about what's going wrong. It would be helpful if there was a verbose flag I could pass to get more info about what's happening, but I'd also really appreciate if anyone has any other ideas what I should be doing differently?

cocreature:
Can you explain what exactly you are trying to do? `ghcide --lsp` is the command that your LSP client (e.g. VSCode) should be using. It isn’t supposed to be run in a terminal and if you do run it in a terminal it won’t print any output until you send it valid LSP commands.

willisplummer:
Ah, sorry - I hadn't realized that. I've been attempting to get this working with vscode by setting the path to ghcide to `./ghcide` but I'm not currently getting any feedback when I edit a type signature to be incorrect and I thought that the lack of feedback when running `ghcide --lsp` was indicative of something.

I actually just identified the issue. Thanks!

ndmitchell:
Should --lsp print a message to stderr to explain its starting a server and listening for messages? I think I've seen someone get confused before.

cocreature:
Sounds like a good idea.

PR #161, opened 2019-10-08
Issue visual warning for humans if --lsp option used

jacg:
Experience shows (e.g. #160) that people sometimes mistakenly start `ghcide` on
the command line with the `--lsp` option (which is intended to be used
only in server/client communication scenarios) and then wonder why
nothing is working..

So let's issue a warning message whenever `--lsp` is used.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -75,6 +75,7 @@ main = do
     if argLSP then do
         t <- offsetTime
         hPutStrLn stderr "Starting LSP server..."
+        hPutStrLn stderr "If you are seeing this in a terminal, you probably should have run ghcidie WITHOUT the --lsp option!"
         runLanguageServer def def $ \event vfs caps -> do
             t <- t
             hPutStrLn stderr $ "Started LSP server in " ++ showDuration t


jacg:
Seemed like a good way of spending 5 minutes before making dinner.

Though I'm a little puzzled by the "Started LSP server in" message never appearing.

jacg:
> maybe you are hiding stderr somehow?

I've modified the code, so that it looks like this:

    if argLSP then do
        t <- offsetTime
        hPutStrLn stderr "Starting LSP server..."
        hPutStrLn stderr "If you are seeing this in a terminal, you probably should have run ghcidie WITHOUT the --lsp option!"
        runLanguageServer def def $ \event vfs caps -> do
            hPutStrLn stderr "ONE ONE ONE ONE ONE ONE ONE ONE ONE ONE ONE ONE "
            hFlush stderr

            t <- t
            hPutStrLn stderr $ "Started LSP server in " ++ showDuration t
            hFlush stderr
            -- very important we only call loadSession once, and it's fast, so just do it before starting
            session <- loadSession dir
            let options = (defaultIdeOptions $ return session)
                    { optReportProgress = clientSupportsProgress caps }
            initialise (mainRule >> action kick) event (logger minBound) options vfs
        hPutStrLn stderr "TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO "
        hFlush stderr

Then I switch into a really basic shell, without any of my customizations, and do

````sh
sh-4.4$ ghcide --lsp
ghcide version: 0.0.3 (GHC: 8.6.5)
Starting LSP server...
If you are seeing this in a terminal, you probably should have run ghcidie WITHOUT the --lsp option!
 ````

... and it sits waiting there (NOTE, there is a single space at the beginning of the last line), until I press any key, at which point I get 
````
TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO TWO 
````
and the process exits.

Any ideas?

cocreature:
That looks fine? The server is exiting because you are sending an invalid LSP message.

jacg:
My point being that the effect of `hPutStrLn stderr $ "Started LSP server in " ++ showDuration t` is never seen.

cocreature:
oh I see, the callback is only executed once the client sends the initialize request which in your terminal never occurs.

jacg:
That's what I was missing. Thanks.

PR #167, opened 2019-10-15
Add goto/hover tests for type constructors from other package

jacg:
Add *go to defirition* and *hover* tests for type constructors which are defined in a different package. This adds a new row, *external type constructor*, to the table from #139:

| | find definition | hover |
|-------------|-----------|---------|
| field in record definition | :heavy_check_mark:  | :heavy_check_mark: |
| field in record construction | :x:  | :x: |
| field name as accessor | :heavy_check_mark: | :heavy_check_mark: |
| top level name | :heavy_check_mark: | :heavy_check_mark: |
| record data constructor | :x: | :x: |
| plain data constructor | :heavy_check_mark: | :heavy_check_mark:  |
| type constructor | :heavy_check_mark: | :x: |
| external type constructor | :heavy_minus_sign: | :x: |
| plain parameter | :heavy_check_mark: | :heavy_check_mark: |
| pattern match name | :heavy_check_mark: | :heavy_check_mark: |
| top level operator | :heavy_check_mark: | :heavy_check_mark: |
| parameter operator | :heavy_check_mark:  | :heavy_check_mark:  |
| name in backticks | :heavy_check_mark: | :heavy_check_mark: |

*go to definition* (for type constructors from other packages) probably can't be made to work at all, although one might argue that you should be taken to the line that imports the type.

*hover* should probably display the module from which the name is imported.

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -740,62 +740,66 @@ findDefinitionTests = let
   yes, broken :: (TestTree -> Maybe TestTree)
   yes    = Just -- test should run and pass
   broken = Just . (`xfail` "known broken")
+  cant   = Just . (`xfail` "cannot be made to work")
 --  no = const Nothing -- don't run this test at all
 
   source = T.unlines
     -- 0123456789 123456789 123456789 123456789
     [ "{-# OPTIONS_GHC -Wmissing-signatures #-}" --  0
     , "module Testing where"                     --  1
-    , "data TypeConstructor = DataConstructor"   --  2
-    , "  { fff :: String"                        --  3
-    , "  , ggg :: Int }"                         --  4
-    , "aaa :: TypeConstructor"                   --  5
-    , "aaa = DataConstructor"                    --  6
-    , "  { fff = \"\""                           --  7
-    , "  , ggg = 0"                              --  8
-    , "  }"                                      --  9
+    , "import Data.Text (Text)"                  --  2
+    , "data TypeConstructor = DataConstructor"   --  3
+    , "  { fff :: Text"                          --  4
+    , "  , ggg :: Int }"                         --  5
+    , "aaa :: TypeConstructor"                   --  6
+    , "aaa = DataConstructor"                    --  7
+    , "  { fff = \"\""                           --  8
+    , "  , ggg = 0"                              --  9
     -- 0123456789 123456789 123456789 123456789
-    , "bbb :: TypeConstructor"                   -- 10
-    , "bbb = DataConstructor \"\" 0"             -- 11
-    , "ccc :: (String, Int)"                     -- 12
-    , "ccc = (fff bbb, ggg aaa)"                 -- 13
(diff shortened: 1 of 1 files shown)


jacg:
Maintaining the source code on which these tests operate, in `Text` within the test source code, is getting rather unwieldy. So I tried to move it out to a separate source file, and use `openDoc` instead of `openDoc'` but I had no success. Even though the file is read in correctly, no definitions or hovers are ever found at all. Maybe there's something I'm not understanding about the effects of the `--cwd` option to `ghcide`. Any thoughts?

I'd like to have a go at turning some of these :x:s into  :heavy_check_mark:s but I get lost among the thousands of type constructors in the GHC API, with which I am not at all familiar. Any suggestions?

aherrmann-da:
> Maintaining the source code on which these tests operate, in `Text` within the test source code, is getting rather unwieldy. So I tried to move it out to a separate source file, and use `openDoc` instead of `openDoc'` but I had no success. Even though the file is read in correctly, no definitions or hovers are ever found at all. Maybe there's something I'm not understanding about the effects of the `--cwd` option to `ghcide`. Any thoughts?

Do you have that trial available in a branch somewhere?

cocreature:
> Conceivably, it could do something like IntelliJ and download sources. But, that's not available right now.

We already have most of the infrastructure for this in place assuming there are `.hie` files available for the dependencies. Locating the sources and hie files is parametrized so we could fairly easily swap this out. Cabal usually has the sources somewhere in `~/.cabal` so it would be feasible to point there.

jacg:
@aherrmann-da 

> Do you have that trial available in a branch somewhere?

[here](https://github.com/jacg/ghcide/tree/external-test-source)

aherrmann-da:
@jacg I took a look at the `openDoc` issue. The confusing bit is that the LSP session runs in a temporary directory, but the test runner still runs in the repository root. One way to deal with this is to refer to the source file by absolute path, like so:
``` diff
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -692,7 +692,8 @@ findDefinitionTests :: TestTree
 findDefinitionTests = let

   tst (get, check) pos targetRange title = testSession title $ do
-    doc <- openDoc' "Testing.hs" "haskell" source
+    sourceFile <- liftIO (makeAbsolute "test/data/GotoHover.hs")
+    doc <- openDoc sourceFile "haskell"
     found <- get doc pos
     check found targetRange
```
In that case you'll want to make sure that the test case doesn't modify the source file. Copying the file into the temporary session directory might be a safer option.

You should also add the source file to the `extra-source-files` section in the Cabal file, to make sure that it's included in source distributions.

jacg:
@aherrmann-da Thanks. This approach was still giving me problems, and in the end it seemed simpler and safer to copy the file into the temporary session directory with a combination of `readFile` and `openDoc'`, and I have submitted this approach in #175.

Issue #170, opened 2019-10-17
Type errors not shown when using cabal


kosmikus:
Under certain circumstances that I have not yet been able to determine completely, ghcide seems to accept erroneous files without complaining when using a `cabal` cradle. (I cannot reproduce this with either direct or stack cradles.)

This seems to happen with both ghcide-0.0.3 from Hackage and the current master, built using ghc-8.6.5.

### To reproduce

1. Create a new directory `ghcide-test`.

2. Create a file `cabal.project`:
    ```
    packages: .
    ```

3. Create a file `ghcide-test.cabal`:
    ```
    name: ghcide-test
    version: 0.1
    build-type: Simple
    cabal-version: >= 1.2

    library
      default-language: Haskell2010
      hs-source-dirs: .
      build-depends:
        base
      exposed-modules:
        Foo
    ```

4. Create a file `Foo.hs`:
    ```haskell
    module Foo where

    x :: Int
    x = "foo"

    y :: Int
    y = _
    ```

5. Run `ghcide`.

This is the output I'm getting:
```
Starting ghcide (GHC v8.6)
/home/andres/repos/ghcide-test
[1/6] Finding hie-bios cradle
Cradle {cradleRootDir = "/home/andres/repos/ghcide-test", cradleOptsProg = CradleAction: cabal}

[2/6] Converting Cradle to GHC session

[3/6] Initialising IDE session

[4/6] Finding interesting files
Found 1 files

[5/6] Setting interesting files
[DEBUG] Set files of interest to: [NormalizedFilePath "./Foo.hs"]
[DEBUG] Starting shakeRun (aborting the previous one took 0.00s)

[6/6] Loading interesting files
[DEBUG] Finishing shakeRun (took 0.00s, exception: thread killed)
[DEBUG] Starting shakeRun (aborting the previous one took 0.00s)
[DEBUG] Finishing shakeRun (took 0.16s, completed)
Files that worked: 1
Files that failed: 0
Done
```

I would expect ghcide to report the errors. In the editor, no errors in `Foo.hs` are shown either.

---

Things seem to become better once I enable more warnings in `ghc-options` such as `-Wall`, `-Wdeferred-type-errors` and/or `-Wtyped-holes`.

I have the feeling that the output is somewhat non-deterministic even in such cases though, with not always the same errors being shown and occasionally the file still being accepted without errors being reported (but rarely). Could there be a race condition involved?

kosmikus:
[...]

PR #172, opened 2019-10-19
#170, make sure the right warnings are set when deferring errors to warnings

ndmitchell:


cocreature:
Thanks a lot for tracking this down @kosmikus and for turning it into a PR @ndmitchell. I completely agree that this patch makes sense.

PR #173, opened 2019-10-20
Prepare for new releases

cocreature:


CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+  this.
+* Properly declare that we should support code actions. This helps
+  with some clients that rely on this information to enable code
+  actions (thanks @jacg).
+* Fix a race condition caused by sharing the finder cache between
+  concurrent compilations.
+* Avoid normalizing include dirs. This avoids issues where the same
+  file ends up twice in the module graph, e.g., with different casing
+  for drive letters.
+* More code actions.
```
ocharles: Could we be more precise?

cocreature: I’ll try to expand it tomorrow.

cocreature: Actually it turns out that all of those changes were before the last release so I’ve removed this. Should have looked closer at the commits initially.

PR #175, opened 2019-10-20
Move test sample code out to external file

jacg:
This has been written on top of #168 which has not been merged yet: only the last 2 commits belong to this PR.

The sample code used in the goto def / hover tests, is becoming to unwieldy to be kept in strings in the test sources. So it has been moved out into a separate file in a new directory `test/data`.

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -173,6 +173,7 @@ test-suite ghcide-tests
         ghcide:ghcide
     build-depends:
         base,
+        bytestring,
         containers,
         directory,
         extra,
@@ -185,6 +186,7 @@ test-suite ghcide-tests
         -- which works for now.
         ghc,
         --------------------------------------------------------------
+        ghcide,
         haskell-lsp-types,
         lens,
         lsp-test >= 0.8,
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -15,6 +15,7 @@ homepage:           https://github.com/digital-asset/ghcide#readme
 bug-reports:        https://github.com/digital-asset/ghcide/issues
 tested-with:        GHC==8.6.5
 extra-source-files: include/ghc-api-version.h README.md CHANGELOG.md
+                    test/data/GotoHover.hs
 
 source-repository head
     type:     git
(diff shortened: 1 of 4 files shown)


jacg:
Hlint suggests
````
Found:
  T.readFile
Perhaps:
  Data.Text.Extended.readFileUtf8
````

but I'm not having much luck finding a provider of `Data.Text.Extended`.

**Comment by ndmitchell** [comment]
@jacg it's recommending https://github.com/digital-asset/daml/blob/99ea93168d84d456060bbdad4d7922e19eacf4f8/libs-haskell/da-hs-base/src/Data/Text/Extended.hs#L44 - an artefact of before we split ghcide out from daml. The advice is good though - always better to avoid readFile which has confusing semantics regarding locale.

**Comment by jacg** [comment]
How should I deal with it?

1. Just implement the equivalent of `readFileUtf8` at the place where I'm using `readFile`.
2. Implement the equivalent and provide it as a utility within `ghcide`
   a) in a subsequent PR
   b) in this PR
   c) in a PR on which this one should be rebased.
3. Something else.

**Comment by ndmitchell** [comment]
In place is a bad idea, since with the name its super obvious what it does. Doing it in a separate commit just makes more work for you, so I wouldn't do that. As to where to put it, Development.IDE.GHC.Util seems the least silly place for it, and creating a module just for this seems pretty silly, so I'd put it in Development.IDE.GHC.Util. The real ideal place for it is in Data.Text.IO itself, since calling readFile is a bad idea, and would be great if readFileUtf8 was next to it.

**Comment by jacg** [comment]
I've put `readFileUtf8` in `Development.IDE.GHC.Util` ... but given that the implementation uses `readFile` hlint will complain ...

**Comment by jacg** [comment]
... no, apparently it doesn't mind that (because it's the bytestring version?)

In which case, I think that this is ready.

File: ghcide.cabal [hunk 1]
```diff
@@ -194,7 +197,7 @@ test-suite ghcide-tests
         text
     hs-source-dirs: test/cabal test/exe test/src
     include-dirs: include
-    ghc-options: -threaded -Wall -Wno-name-shadowing
+    ghc-options: -threaded -Wall -Wno-name-shadowing -Wincomplete-patterns
```
cocreature: This should be part of `-Wall`, no?

jacg:
> Would be great if you could rebase this on master now that we’ve merged the other PR.

Done.

PR #188, opened 2019-11-07
Adapt code to use hie-bios's new CradleLoadResult

afcowie:
Unclear why **hie-bios** changed `getCompilerOptions` from its Either return type, but they did. This patch adapts the code in Main to handle this new type.

Bumps dependency on **hie-bios** to >= 0.3.0

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -154,7 +154,12 @@ showEvent lock e = withLock lock $ print e
 
 cradleToSession :: Cradle -> IO HscEnvEq
 cradleToSession cradle = do
-    opts <- either throwIO return =<< getCompilerOptions "" cradle
+    result <- getCompilerOptions "" cradle
+    opts <- case result of
+        CradleSuccess x -> return x
+        CradleFail e -> throwIO e
+        CradleNone -> error "Not only did it not work, but there's no error either"
+
     libdir <- getLibdir
     env <- runGhc (Just libdir) $ do
         _targets <- initSession opts
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -158,7 +158,7 @@ cradleToSession cradle = do
     opts <- case result of
         CradleSuccess x -> return x
         CradleFail e -> throwIO e
-        CradleNone -> error "Not only did it not work, but there's no error either"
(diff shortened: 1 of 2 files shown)


exe/Main.hs
```diff
@@ -154,7 +154,12 @@ showEvent lock e = withLock lock $ print e
 
 cradleToSession :: Cradle -> IO HscEnvEq
 cradleToSession cradle = do
-    opts <- either throwIO return =<< getCompilerOptions "" cradle
+    result <- getCompilerOptions "" cradle
+    opts <- case result of
+        CradleSuccess x -> return x
+        CradleFail e -> throwIO e
+        CradleNone -> error "Not only did it not work, but there's no error either"
```
cocreature: Looking at the docs at https://github.com/mpickering/hie-bios/blob/8d73b5cb416a1f9b8ecd95d25c094e0f43f0123a/src/HIE/Bios/Cradle.hs#L133, the `none` cradle is intended to be used to ignore a directory so we should probably output something like “Refusing to load directory due to none cradle”. It would also be nice to use `fail` instead of `error` given that we are in `IO` anyway.

afcowie: Done

afcowie:
I should point out that I'm testing by building with the following _stack.yaml_ file:

```yaml
resolver: lts-14.13
packages:
- .
extra-deps:
- haskell-lsp-0.17.0.0
- haskell-lsp-types-0.17.0.0
- lsp-test-0.8.0.0
- git: https://github.com/mpickering/hie-bios
  commit: 55da0d88710686e47009c627ee2123b000ba62b7

nix:
  packages: [zlib]
```

which is valid now that the patch to expose CradleLoadResult is merged.

cocreature:
Oh I didn’t realize that the new version of `hie-bios` is not yet released. In that case, I think it would probably be better to guard this against a `MIN_VERSION` pragma. I would like to avoid depending on unreleased versions.

mpickering:
We are going to do the 0.3 release shortly. It contains a lot of improvements and changes that will need to be adapted for.

cocreature:
Closing in favor of #255

PR #192, opened 2019-11-13
Support plugins

jinwoo:
Call initializePlugins before running typecheck.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -99,19 +99,22 @@ typecheckModule (IdeDefer defer) packageState deps pm =
     runGhcEnv packageState $
         catchSrcErrors "typecheck" $ do
             setupEnv deps
-#if MIN_GHC_API_VERSION(8,6,0)
-            session <- getSession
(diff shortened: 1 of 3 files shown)


jinwoo:
I was going to ask about tests :) Any pointers to how to test this?

src/Development/IDE/Core/Compile.hs
```diff
@@ -95,8 +99,17 @@ typecheckModule (IdeDefer defer) packageState deps pm =
     runGhcEnv packageState $
         catchSrcErrors "typecheck" $ do
             setupEnv deps
+#if MIN_GHC_API_VERSION(8,6,0)
```
ndmitchell: Any way to abstract the key steps out into a function with a single signature that is the same on all versions (merely a noop on one version), then have this code just call that? Would make managing the CPP a bit easier.

jinwoo:
Thanks, @ndmitchell . I'm working on tests at the moment by adding some 3rd-party plugin. Will update this once it's done.

jinwoo:
Added a test. PTAL.

cocreature:
Nice work, thank you!

Issue #25, opened 2019-09-11
Error "expectJust getLinkDeps"


ocharles:
I'm trying `ghcide` on just a single Cabal package from our work codebase. I get the following:

```
6/6] Loading interesting files
[DEBUG] Finishing shakeRun (took 0.00s, exception: thread killed)
[DEBUG] Starting shakeRun (aborting the previous one took 0.00s)
File:     ../entities/Model/Progress.hs
Range:    1:0-100001:0
Source:   compiler
Severity: DsError
Message: 
  expectJust getLinkDeps
  CallStack (from HasCallStack):
  error, called at compiler/utils/Maybes.hs:55:27 in ghc:Maybes
  expectJust, called at compiler/ghci/Linker.hs:719:28 in ghc:Linker
File:     ../entities/Model/Common.hs
Range:    1:0-100001:0
Source:   compiler
Severity: DsError
Message: 
  expectJust getLinkDeps
  CallStack (from HasCallStack):
  error, called at compiler/utils/Maybes.hs:55:27 in ghc:Maybes
  expectJust, called at compiler/ghci/Linker.hs:719:28 in ghc:Linker
File:     ../entities/Model/OctopartV3.hs
Range:    1:0-100001:0
Source:   compiler
Severity: DsError
Message: 
  expectJust getLinkDeps
  CallStack (from HasCallStack):
  error, called at compiler/utils/Maybes.hs:55:27 in ghc:Maybes
  expectJust, called at compiler/ghci/Linker.hs:719:28 in ghc:Linker
File:     ../entities/Model/User.hs
Range:    1:0-100001:0
Source:   compiler
Severity: DsError
Message: 
  expectJust getLinkDeps
  CallStack (from HasCallStack):
  error, called at compiler/utils/Maybes.hs:55:27 in ghc:Maybes
  expectJust, called at compiler/ghci/Linker.hs:719:28 in ghc:Linker
[DEBUG] Finishing shakeRun (took 35.84s, completed)
Files that worked: 0
Files that failed: 1
 * app/main.hs
```

Any idea what this `expectJust getLinkDeps` business is about?

ndmitchell:
You aren't the first person to report that. Do you have TemplateHaskell or ANN involved?

ocharles:
Yes, all of these modules use `persistent-template` to derive `persistent` models using Template Haskell.

ndmitchell:
I imagine that's certainly related. It would be great to get a simple example that exhibited the problem.

[...]

Issue #34, opened 2019-09-11
Support TemplateHaskell


ndmitchell:
Trying to synthesise one ticket that describes all the TH work. My understanding of the current state is:

* If you have TH or ANN pragmas, and they all refer to code that is imported by packages or in this module, it works.
* If you refer to other modules in this package it breaks, perhaps with #25.

The solution is likely to make sure we have compiled the dependencies to bytecode and send them at the linker before type checking a module that uses TH or ANN.

However, we don't have tests for the current state, and it works by luck rather than deliberately.

cocreature:
I have also seen some issue around async exceptions and TH which seemed to make GHC very unhappy (I don’t recall the precise error message).

pwm:
Hi, anyone is in a position to estimate how much work this would entail? Since my team started using ghcide we can’t imagine going back to the stone age ever. You guys did an amazing job! On our end TH support seems to be the last big issue that’s in the way of full adoption.

ndmitchell:
My guess is you need to:

* Figure out which modules require TH, nice and easy.
* For those that do, first build all the dependencies to object code (moderate amount of work, probably < 100 lines, fits fairly well with the graph approach).
* Then send that object code to the linker (dunno how hard this is, might have problems with state management).

So probably not obviously hard, but I'd almost be surprised if something didn't get in your way.

domenkozar:
@DavidEichmann in case you were using ghcide-nix, that has been fixed now.

serras:
I would like to tackle this issue, since out team uses TH in quite some places.
Is there any further information that I can read to work on this?

ndmitchell:
None that I'm aware of @serras - but best of luck!

serras:
This seems to be related https://gitlab.haskell.org/ghc/ghc/issues/8025, at least it ends up with the same error as #212

serras:
I have started working on the issue at https://github.com/serras/ghcide/commit/90c784b6f9503a2686f3e585e678ee65f60e3d9a in a quite crude way, asking to compile all dependencies even if the file does not use Template Haskell.
Now I get errors which look the ones in #212, even though now I am loading some bytecode:
```
    ./src/B.hs
Range:    1:0-100001:0
Source:   typecheck
Severity: DsError
Message:
  Program error:
  ByteCodeLink.lookupCE
  During interactive linking, GHCi couldn't find the following symbol:
  A_a_closure
  This may be due to you not asking GHCi to load extra object files,
[...]

PR #212, opened 2019-11-27
Fix "expectJust getLinkDeps" error

jinwoo:
The linker crashes because the hm_linkable field of HomeModInfo is
empty. Fill it with correct data.

Fixes https://github.com/digital-asset/ghcide/issues/25.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -191,7 +191,11 @@ mkTcModuleResult
 mkTcModuleResult tcm = do
     session <- getSession
     (iface, _) <- liftIO $ mkIfaceTc session Nothing Sf_None details tcGblEnv
-    let mod_info = HomeModInfo iface details Nothing
+    let summary = pm_mod_summary . tm_parsed_module $ tcm
+        -- This is a fake Linkable with an empty list as the last argument.
+        -- As we're not generating code, it is probably fine.
+        linkable = LM (ms_hs_date summary) (ms_mod summary) []
+        mod_info = HomeModInfo iface details (Just linkable)
     return $ TcModuleResult tcm mod_info
   where
     (tcGblEnv, details) = tm_internals_ tcm
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -41,6 +41,7 @@ main = defaultMain $ testGroup "HIE"
   , codeActionTests
   , findDefinitionAndHoverTests
   , pluginTests
+  , linkerTests
   ]
 
 initializeResponseTests :: TestTree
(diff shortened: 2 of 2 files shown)


cocreature:
Awesome thanks a lot! Can we get some test for this?

jinwoo:
Sorry I don't have a simple repro and so am not sure how to test this.

jinwoo:
Both our case and the repro mentioned in https://github.com/digital-asset/ghcide/issues/25 involve `GeneralizedNewtypeDeriving` and `TemplateHaskell`, and I tried to create a simple repro using them. But no, I couldn't.

cocreature:
If there’s no easy way to reproduce this, I’m happy to go without a testcase for now. Looking at the docs of [`Linkable`](https://hackage.haskell.org/package/ghc-8.6.5/docs/HscTypes.html#v:LM) they state `INVARIANT: A valid linkable always has at least one Unlinked item. If this list is empty, the Linkable represents a fake linkable, which is generated in HscNothing mode to avoid recompiling modules.`. We seem to be using an empty list here which might be fine since we aren’t doing code generation but it isn’t quite clear to me. Do you know more about this and can help me understand this?

jinwoo:
I'm not really an expert in GHC internals and, as you said, I assumed an empty list is ok because we're not generating code here. I added a comment there.

ndmitchell:
https://github.com/digital-asset/ghcide/issues/25#issuecomment-530289283 has a minimal reproducible example, and you commented on that ticket that you have a fix for it? If so, could that minimal example become the test? Even if it can't (the dependency on external packages is annoying) could you test that this patch fixes it?

jinwoo:
I finally succeeded to create a test case. Let me clean it up and upload soon.

jinwoo:
Added a test, which fails with "expectJust getLinkDeps" without the fix.

But with the fix, I'm getting a different error and I'm not sure how to fix that. Probably it is a separate issue. This same patch did fix our internal code.

jinwoo:
I also tried from the command-line the repro case in https://github.com/digital-asset/ghcide/issues/25#issuecomment-530289283, and I'm getting a new build error there too as below. Maybe that's because I'm passing an empty list as the `linkableUnlinked` field of `Linkable`.

```
  Program error:
  ByteCodeLink.lookupCE
  During interactive linking, GHCi couldn't find the following symbol:
  Settings_settings_closure
  This may be due to you not asking GHCi to load extra object files,
  archives or DLLs needed by your current session. Restart GHCi, specifying
  the missing library using the -L/path/to/object/dir and -lmissinglibname
  flags, or simply by naming the relevant files on the GHCi command line.
  Alternatively, this link failure might indicate a bug in GHCi.
  If you suspect the latter, please send a bug report to:
  glasgow-haskell-bugs@haskell.org
```

jinwoo:
I tried filling the `linkableUnlinked` field as GHC does in https://github.com/ghc/ghc/blob/master/compiler/main/DriverPipeline.hs#L219. But because ghcide doesn't generate `.o` files, I get errors like `Settings.o: getModificationTime:getFileStatus: does not exist (No such file or directory)`.

I'm not sure whether this new problem always happens with this "fix". If not, the fix would be worth it, but I don't have a good way to prove that.

merijn:
This does fix the original error, but I run into the `ByteCodeLink.lookupCE` now too.

merijn:
I've seen similar errors from ghci in the past, which at the time needed `-fobject-code` to resolve. On the other hand, ghci seems to be able to load these modules just fine without `-fobject-code`...

jinwoo:
I think the new error `ByteCodeLink.lookupCE` is related to https://github.com/digital-asset/ghcide/issues/34.

jinwoo:
I'm going to revert this change. This seems to only fix the first symptom. I think the proper fix should be done together with https://github.com/digital-asset/ghcide/issues/34; by generating bytecode for things that are used by Template Haskell, and passing the bytecode as part of the `Linkable`.

PR #225, opened 2019-12-08
Fix horrible path normalisation issue

pepeiborra:
Not sure why sometimes the path to the cradle is absolute and other times it's relative. Using Linux and direct cradles, happens both on Emacs and Code. 

Because of this issue, ghcide v0.4 is broken for me on certain files both on personal projects and at work.


ndmitchell:
Would canonicalizeFilePath end up doing both these things, potentially more robustly?

cocreature:
`canonicalizeFilePath` also resolves symlinks which doesn’t necessarily make sense.

ndmitchell:
Good point!

Issue #211, opened 2019-11-25
Add completion support


maralorn:
This is an obvious feature request.

Are there already any thoughts on how to implement it and how hard it‘s gonna be?

ndmitchell:
haskell-ide-engine has some form of completion, so probably best to see what they do. That probably also gives a good idea of how hard it is.

cocreature:
Shouldn’t be too hard, I’ve looked at the code in `haskell-ide-engine` in the past and it should translate fairly easily.

serras:
I see that the license of `haskell-ide-engine` is different from `ghcide`'s. How much can be copied/slightly-modified from [their implementation](https://github.com/haskell/haskell-ide-engine/blob/master/src/Haskell/Ide/Engine/LSP/Completions.hs) into `ghcide`? I would be really happy to help, but I don't want to introduce license problems.

ndmitchell:
haskell-ide-engine is licensed under BSD. ghcide is licensed under BSD+Apache. You can freely intermingle the two, e.g. see https://softwareengineering.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache. There are no license problems in copying any haskell-ide-engine code (although, not in this case, but in general, there are cabal helper interdependencies in HIE that are license issues due to GPL)

cocreature:
Added in #227

PR #227, opened 2019-12-09
[WIP] Completion support

serras:
This PR adds completion support (as per #211). Right now it's mostly stripped out parts of `haskell-ide-engine`, put back together into the rule-based system of `ghcide`. It seems to work for value level things, but not so well for type-level things, so further investigation is needed.

The information shown is also very basic. In the future the idea is to also return the documentation of the functions, if available.


serras:
PS: right now there's a problem about how I published my branch, and it seems that the TH-related changes have also slipped in this branch. Hopefully once the TH branch is merged, this problem will go away.

mpickering:
I thought the idea was to *not* implement completion support directly in `ghcide` but as some kind of plugin. Has that plan changed?

ndmitchell:
@mpickering I think the plan was always to implement completions inside ghcide. It was things like Hoogle/HLint/LiquidHaskell we thought were best handled by plugins.

serras:
I was not aware on any plans on this respect, I just found `ghcide` to be a nice project to contribute to! But if you prefer to do it in other way, I'm happy too :)

ocharles:
Just wanted to say I tried this out and it works very nicely!

fendor:
Just mentioning that you are probably inheriting https://github.com/haskell/haskell-ide-engine/issues/1434 from me.

ndmitchell:
Thanks for the heads up @fendor - would be great if we can eventually share these (either because hie and ghcide merge, or because there is a common underlying library everyone uses)

cocreature:
Thanks for the PR, I’m hoping to get around to taking a look at this tomorrow but from an initial look, it looks great :+1:

serras:
There's still a problem I need to solve. The completions only work if the chnages make the problem still typecheck. For example, if I have

```haskell
thing :: String -> I
```

and I want to obtain completions for `I`, then nothing happens. But if I have

```haskell
thing :: String -> Int
```

this is type-correct code and `Int32`, `Int64` and so on are shown.

The question here is: is there any way to "save" the latest type correct `TypecheckedModule` and use it if the new one gives problems? That would be the solution to these problems.

ndmitchell:
The core of the IDE was designed to support getting stale values, e.g. `useWithStale` https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Core/Shake.hs#L432. I'm not sure where that is poked out, or even if it still works (or in fact ever worked), but that was the original design, precisely for this purpose.

serras:
So if I changed my usage of `use TypeCheck` with `useWithStale TypeCheck`, everything should in theory work? Thay would be great!

cocreature:
`useWithStale` definitely works (at least in some circumstances), we use this in DAML. But for `ghcide`, I would have expected that the `-fdefer-type-errors` stuff takes care of making sure that you still get a typechecking result even if typechecking fails.

ndmitchell:
defer type errors will defer type errors, but not parse or name errors, which tends to be when you are asking for completions, and typically I want the best completions that can be found in < 0.1s. I think this is the right place to use deferred type checking.

cocreature:
The code looks good to me, thanks a lot! I’ve also tested it on an internal codebase and it doesn’t seem to increase memory usage significantly which I was a bit afraid off. I am trying to figure out what we need to make sure that this doesn’t cause any licensing problems. I’ll get back to you once I know what steps need to be taken.

serras:
It would be great if you could test the code. It seems that everything works fine except for qualified imports, but I cannot find the reason why. Also, the new version with `useWithStale` works quite well, but not in a completely reliable way.

Some other issues before considering merging:
- The code from HIE uses something called `withSnippets`, but I cannot completely understand why. I am leaning towards removing that part.
- The code which obtains the type for elements doesn't work for imported functions (for example, the completion of `head` only shows `Prelude`, but not the type). My gut feeling is that it's related to `getComplsForOne` (around line 300 of `IDE/Core/Completions`), but any help would be great.

src/Development/IDE/Core/Completions.hs
```diff
@@ -0,0 +1,565 @@
...
+-- The supported languages and extensions
+languagesAndExts :: [T.Text]
+languagesAndExts = map T.pack DynFlags.supportedLanguagesAndExtensions
+  
+-- ---------------------------------------------------------------------
+-- helper functions for pragmas
+-- ---------------------------------------------------------------------
+
+validPragmas :: [(T.Text, T.Text)]
+validPragmas =
```
cocreature: At least these require the `withSnippets` logic so let’s keep this for now.

serras: 👍

src/Development/IDE/LSP/Completions.hs
```diff
@@ -0,0 +1,44 @@
...
+
+-- | Generate code actions.
+getCompletionsLSP
+    :: LSP.LspFuncs ()
+    -> IdeState
+    -> CompletionParams
+    -> IO CompletionResponseResult
+getCompletionsLSP lsp ide CompletionParams{_textDocument=TextDocumentIdentifier uri,_position=position} = do
+    -- disable logging as its quite verbose
+    -- logInfo (ideLogger ide) $ T.pack $ "Code action req: " ++ show arg
```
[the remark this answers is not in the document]
serras: 👍

src/Development/IDE/LSP/Completions.hs
```diff
@@ -0,0 +1,44 @@
...
+    contents <- LSP.getVirtualFileFunc lsp $ toNormalizedUri uri
+    case (contents, uriToFilePath' uri) of
+      (Just cnts, Just path) -> do
+        pfix <- VFS.getCompletionPrefix position cnts
+        let npath = toNormalizedFilePath path
+        tm  <- fmap fst <$> runAction ide (useWithStale TypeCheck npath)
+        cci <- fmap fst <$> runAction ide (useWithStale ProduceCompletions npath)
```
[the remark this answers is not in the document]
serras: Done in the new commit

src/Development/IDE/LSP/Completions.hs
```diff
@@ -0,0 +1,44 @@
...
+    -> IO CompletionResponseResult
+getCompletionsLSP lsp ide CompletionParams{_textDocument=TextDocumentIdentifier uri,_position=position} = do
+    -- disable logging as its quite verbose
+    -- logInfo (ideLogger ide) $ T.pack $ "Code action req: " ++ show arg
+    contents <- LSP.getVirtualFileFunc lsp $ toNormalizedUri uri
+    case (contents, uriToFilePath' uri) of
+      (Just cnts, Just path) -> do
+        pfix <- VFS.getCompletionPrefix position cnts
+        let npath = toNormalizedFilePath path
+        tm  <- fmap fst <$> runAction ide (useWithStale TypeCheck npath)
```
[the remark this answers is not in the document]
serras: Do you have some code or docs I could check to see how to do this? I would rather have a working PR.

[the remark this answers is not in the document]
cocreature: We have an example in the DAML codebase, see https://github.com/digital-asset/daml/blob/c65f9181459cadc3da79687571788bdbd74c4484/compiler/damlc/daml-ide/src/DA/Daml/LanguageServer/CodeLens.hs#L36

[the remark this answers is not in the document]
serras: I have taken care of this in the latest commit :)

cocreature:
Thanks for fixing the positions, I would still like to see some tests before merging this. Doesn’t need to be comprehensive but at least one basic integration test and one that exercises the `useWithStale` code path and checks that the positions are correctly remapped.

serras:
I've written some preliminary tests. However, I cannot get my head around how to check the stale file; I've tried several ways but all of them lead to a stuck test.

cocreature:
I’ve pushed a fix for the tests. The two expected failures where caused by the fact that those did not typecheck nor did you have a stale result. I changed them to first produce something that typechecks so you have a stale result. Given that this already exercises the stale codepath I’ve removed the separate test for that.

Do you have anything else that you want to address? Otherwise I think this should be ready to land (I’ll be on vacation so it might take me longer to respond).

serras:
No, the PR looks perfect to me. I guess we now should wait for bug reports and other problems that may arise.

src/Development/IDE/Core/Completions.hs
```diff
@@ -0,0 +1,565 @@
...
+  | otherwise = error "namespace not recognized"
+
+parseNs :: Value -> Parser NameSpace
+parseNs (String "v") = pure Name.varName
+parseNs (String "c") = pure dataName
+parseNs (String "t") = pure tcClsName
+parseNs (String "z") = pure tvName
+parseNs _ = mempty
+
+instance FromJSON NameDetails where
```
wz1000: All this JSON stuff is not needed here because it is only used for the `resolveCompletion` LSP request, where the types of non local names and documentation is looked up. It doesn't seem like you added support for the resolve completion request.

serras: Nope, I didn't implement the resolver completion request, since all information is already sent in the original response. Is this wrong?

wz1000: I don't think you are going to get types for things not defined in the active module, since they are looked up in the local type env.

wz1000: If you don't plan to implement the resolve completion request, you can remove everything to do with JSON - which is everything from line 42 to 77.

src/Development/IDE/Core/Completions.hs
```diff
@@ -0,0 +1,565 @@
...
+getCompletions CC { allModNamesAsNS, unqualCompls, qualCompls, importableModules }
+               tm prefixInfo caps withSnippets = do
+  let VFS.PosPrefixInfo { VFS.fullLine, VFS.prefixModule, VFS.prefixText } = prefixInfo
+      enteredQual = if T.null prefixModule then "" else prefixModule <> "."
+      fullPrefix  = enteredQual <> prefixText
+
+      -- default to value context if no explicit context
+      context = fromMaybe ValueContext $ getCContext pos (tm_parsed_module tm)
+
+      {- correct the position by moving 'foo :: Int -> String ->    '
```
wz1000, line 361: @serras sorry for digging up this old PR, but do you remember the rationale for this?

serras: Most of this code was taken from `haskell-ide-engine`, and I think that's where this originated.

wz1000: I know, I was one of the original implementors! Guess I'll have to do some more digging...

PR #229, opened 2019-12-11
Retry hDuplicateTo

cocreature:
We have seen a bunch of failures on CI where this failed with
EBUSY. I find the hDuplicateTo here to be quite useful for debugging
since you don’t have to worry about corrupting the JSON-RPC stream to
instead of getting rid of it, we now retry 10 times which should
hopefully at least reduce CI issues.


cocreature:
Doesn’t actually seem to help :cry:

PR #232, opened 2019-12-11
Enhancements to top-level signatures

serras:
I am trying to implement #231. However, I think I have done something very wrong, because I get a `BadDependency` error message.
@cocreature @ndmitchell could you provide some help? Thanks in advance :)

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -102,12 +102,20 @@ typecheckModule (IdeDefer defer) packageState deps pm =
         catchSrcErrors "typecheck" $ do
             setupEnv deps
             let modSummary = pm_mod_summary pm
+                dflags = ms_hspp_opts modSummary
             modSummary' <- initPlugins modSummary
             (warnings, tcm) <- withWarnings "typecheck" $ \tweak ->
                 GHC.typecheckModule $ enableTopLevelWarnings
                                     $ demoteIfDefer pm{pm_mod_summary = tweak modSummary'}
             tcm2 <- mkTcModuleResult tcm
-            return (map unDefer warnings, tcm2)
+            let errorPipeline = unDefer
+                              . (if wopt Opt_WarnMissingSignatures dflags
+                                    then id
+                                    else degradeError Opt_WarnMissingSignatures)
+                              . (if wopt Opt_WarnMissingLocalSignatures dflags
+                                    then id
+                                    else degradeError Opt_WarnMissingLocalSignatures)
+            return (map errorPipeline warnings, tcm2)
 
 initPlugins :: GhcMonad m => ModSummary -> m ModSummary
 initPlugins modSummary = do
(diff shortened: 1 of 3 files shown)


src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -53,14 +54,16 @@ codeLens
     -> CodeLensParams
     -> IO (List CodeLens)
 codeLens _lsp ideState CodeLensParams{_textDocument=TextDocumentIdentifier uri} = do
-    diag <- getDiagnostics ideState
+    -- diag <- getDiagnostics ideState
     case uriToFilePath' uri of
       Just (toNormalizedFilePath -> filePath) -> do
+        _ <- runAction ideState $ use_ TypeCheck filePath
```
cocreature: `use_` throws a `BadDependency` exception if the file does not typecheck. You probably want `useE` which will give you a `Maybe`.

src/Development/IDE/Core/Compile.hs
```diff
@@ -104,7 +104,8 @@ typecheckModule (IdeDefer defer) packageState deps pm =
             let modSummary = pm_mod_summary pm
             modSummary' <- initPlugins modSummary
             (warnings, tcm) <- withWarnings "typecheck" $ \tweak ->
-                GHC.typecheckModule $ demoteIfDefer pm{pm_mod_summary = tweak modSummary'}
+                GHC.typecheckModule $ enableTopLevelWarnings
```
cocreature: I’ve brought this up a couple of times before but I really dislike the idea of enabling warnings by default. If users care about these warnings, they should enable them in their cabal file. I don’t want to see a bunch of warnings in my IDE that I haven’t enabled. If we manage to make this work such that we only show the codelenses by default but not the warnings, I could be convinced that this is fine but given that the codelenses are based on diagnostics this doesn’t seem trivial and I think the complexity of trying to make this work is probably not worth it.

serras: I am happy both ways. I can attempt to write something which remembers whether these warning were already in effect at the beginning, so that we can filter out diagnostics if needed. But I agree: it might be quite complicated.
@ndmitchell you seemed to have an idea of how to achieve this in #231. Any suggestions?

src/Development/IDE/LSP/CodeAction.hs
serras: Resolved in a new commit.

src/Development/IDE/Core/Compile.hs
ndmitchell: Code which doesn't have a warning is more likely to have top-level signatures omitted, and more likely to benefit from having them as code lenses, since I see them as helpful docs, rather than a mechanism to easily add them. I separately think that turning on some warnings in the IDE vs compiler is reasonable, but see that as unrelated to this patch.

I'm not convinced the code is that complex. You turn on the warning always, but demote the warning to an info if errMsgReason is missing top level signatures. At some point (if we want) we ditch info level diagnostics before sending them to the LSP client. A few lines here, a few lines (that will be generally useful) in the diagnostic sending code.

cocreature: I am admittedly a bit biased here, I don’t recall the last time I worked on a project that didn’t have these warnings enabled and that seems to be quite common in the ecosystem in general, so I’m somewhat hesitant to add any complexity (even if it’s probably not too much) to cater to that usecase.

That said, the argument that we’ll need this for other warnings eventually that are less popular is compelling so I’m fine with adding it.

ndmitchell: None of my projects have it turned on :)

cocreature: That’s your loss :trollface:

ndmitchell: My loss is not getting to work with you awesome guys every day anymore!

cocreature: :hugs:

serras: It took a while, but finally I managed to "degrade" warnings about missing signatures to info level if they were not explicitly enabled by the user. They look great in VSCode, since they are shown with a blue underline.

src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -53,14 +55,16 @@ codeLens
     -> CodeLensParams
     -> IO (List CodeLens)
 codeLens _lsp ideState CodeLensParams{_textDocument=TextDocumentIdentifier uri} = do
-    diag <- getDiagnostics ideState
+    -- diag <- getDiagnostics ideState
```
cocreature: It looks like this line can be removed?

PR #235, opened 2019-12-12
Workaround hDuplicateTo issues

cocreature:
We have seen a bunch of failures on CI where this failed with
EBUSY. I find the hDuplicateTo here to be quite useful for debugging
since you don’t have to worry about corrupting the JSON-RPC stream to
instead of getting rid of it, we add a somewhat ugly workaround.

There is an explanation in an inline comment on why this helps but
admittedly I am somewhat guessing since I don’t understand what is
actually allocating the file descriptor that turns out to be
stdout. That said, I am not guessing on the results: Without this PR I
am able to make this fail in roughly 50% of the cases on CI whereas
with this PR, I’ve now run it 60 times on CI without a single failure.

### DIFF
--- a/src/Development/IDE/GHC/Util.hs
+++ b/src/Development/IDE/GHC/Util.hs
@@ -20,11 +20,15 @@ module Development.IDE.GHC.Util(
     moduleImportPath,
     HscEnvEq, hscEnv, newHscEnvEq,
     readFileUtf8,
+    hDuplicateTo,
     cgGutsToCoreModule
     ) where
 
 import Config
+import Control.Concurrent
 import Data.List.Extra
+import Data.Maybe
+import Data.Typeable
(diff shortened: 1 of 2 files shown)


ndmitchell:
Can you raise this with GHC HQ too? I have a failure in the extra library very very occasionally, and its always near hDuplicateTo, and this seems an entirely plausible explanation of it.

cocreature:
I was planning to do that. I tried getting a response in IRC on why the `hClose` is there in the first place but haven’t gotten one so I’ll just open an issue/a PR.

cocreature:
@ndmitchell In case you wan to follow along, I’ve opened https://gitlab.haskell.org/ghc/ghc/merge_requests/2318/diffs to see if this can be upstreamed.

Issue #237, opened 2019-12-13
Hover on names bound by do-notation shows type spuriously wrapped in monad


jacg:
For example, in `do xxx <- Just True` the type of `xxx` is `Bool`, but hover reports it as `Maybe Bool`.

The spurious wrapping only appears at the bidning site; hover shows the correct type where the name is used.

Tests demonstrating this issue are available in #238.

jacg:
A similar problem occurs in list comprehensions.

Tests showing this have been added to #238.

PR #244, opened 2019-12-14
Cleaner Show instance for SpanInfo

jacg:
This helped with debugging #237, so maybe it's worth keeping.

### DIFF
--- a/src/Development/IDE/Spans/Type.hs
+++ b/src/Development/IDE/Spans/Type.hs
@@ -13,9 +13,9 @@ module Development.IDE.Spans.Type(
 
 import GHC
 import Control.DeepSeq
-import Data.Maybe
 import OccName
-
+import Data.List
+import Development.IDE.GHC.Util
 
 -- | Type of some span of source code. Most of these fields are
 -- unboxed but Haddock doesn't show that.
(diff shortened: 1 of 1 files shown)


src/Development/IDE/Spans/Type.hs
```diff
@@ -29,15 +29,17 @@ data SpanInfo =
...
            ,spaninfoSource :: !SpanSource
            -- ^ The actutal 'Name' associated with the span, if
             -- any. This can be useful for accessing a variety of
             -- information about the identifier such as module,
             -- locality, definition location, etc.
            }
 instance Show SpanInfo where
-  show (SpanInfo sl sc el ec t n) = show [show sl, show sc, show el, show ec, show $ isJust t, show n]
+  show (SpanInfo sl sc el ec t n) =
+    intercalate " " ["(SpanInfo", show sl, show sc, show el, show ec
```
ndmitchell: This can be `unwords` rather than `intercalate " "`.

ndmitchell:
Makes a lot of sense - the original Show instance is clearly a quick hack.

src/Development/IDE/Spans/Type.hs
jacg: Of course. Funny how the route you take influences what you end up with.

Fixed.

PR #252, opened 2019-12-15
Fix #246

### LINKED ISSUE #246: Hover only works for first clause in multi-equation function definition
... as demonstrated by the tests in #245.

jacg:
`getTypeLHsBind` returned a single span corresponding to the overall function
binding. The fix drills down into the individual matches and returns a span for
each of them.

Fixes #246.

### DIFF
--- a/src/Development/IDE/Spans/Calculate.hs
+++ b/src/Development/IDE/Spans/Calculate.hs
@@ -98,9 +98,16 @@ getTypeLHsBind :: (GhcMonad m)
                => TypecheckedModule
                -> LHsBind GhcTc
                -> m [(SpanSource, SrcSpan, Maybe Type)]
-getTypeLHsBind _ (L _spn FunBind{fun_id = pid,fun_matches = MG{}}) =
-  return [(Named $ getName (unLoc pid), getLoc pid, Just (varType (unLoc pid)))]
+getTypeLHsBind _ (L _spn FunBind{ fun_id = pid
+                                , fun_matches = MG{mg_alts=(L _ matches)}}) =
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Spans/Calculate.hs
```diff
@@ -98,8 +98,9 @@ getTypeLHsBind :: (GhcMonad m)
                => TypecheckedModule
                -> LHsBind GhcTc
                -> m [(SpanSource, SrcSpan, Maybe Type)]
-getTypeLHsBind _ (L _spn FunBind{fun_id = pid,fun_matches = MG{}}) =
-  return [(Named $ getName (unLoc pid), getLoc pid, Just (varType (unLoc pid)))]
+getTypeLHsBind _ (L _spn FunBind{ fun_id = pid
+                                , fun_matches = MG{mg_alts=(L _ matches)}}) =
+  return [(Named (getName (unLoc pid)), getLoc match, Just (varType (unLoc pid))) | match <- matches ]
```
cocreature: It looks like we no longer record the  location of `pid`. Is that guaranteed to be one of the matches?

jacg: I don't have a language-lawyer-level guarantee, but I think that the location of the first match is the same as the location of the overall `pid`. Experimental evidence in support of this: before this fix, gotoDef only worked for the first clause, and jumped to itself; now gotoDef works for all clauses, and jumps to the first one.

Is that enough, or would you like me to dig a bit more?

cocreature: Definitely good enough, if it turns out to be a problem we can always address it later. Thanks for the response!

PR #254, opened 2019-12-15
Upgrade to haskell-lsp 0.19

cocreature:


src/Development/IDE/Core/FileStore.hs
```diff
@@ -76,7 +76,7 @@ makeVFSHandle = do
               modifyVar_ vfsVar $ \(nextVersion, vfs) -> pure $ (nextVersion + 1, ) $
                   case content of
                     Nothing -> Map.delete uri vfs
-                    Just content -> Map.insert uri (VirtualFile nextVersion (Rope.fromText content)) vfs
+                    Just content -> Map.insert uri (VirtualFile nextVersion 0 (Rope.fromText content)) vfs
```
mpickering: Not sure this line is right. The first field is supposed to be a version number you get from the LSP request and the second number an auto-incrementing version. It seems you are not keeping track of the version number reported by `haskell-lsp` at all?

cocreature: Afaik the second number is only used for `persistVirtualFile` which we don’t use.

cocreature: I’ve checked the `haskell-lsp` source again which seems to confirm this and I’ve added a comment to make it clear that we do not care about this field.

cocreature: Also note that this is the fake VFS handle that we use in some DAML tests that don’t use LSP, the proper VFS handle that you usually get is the one below in `makeLSPVFSHandle`.

pepeiborra:
I've noticed a hover performance degradation in head, and bisected it to this change. Hover takes twice as long for me.

PR #257, opened 2019-12-16
Upgrade to hie-bios 0.3

cocreature:
This does not yet take advantage of any new features but we should at
least be able to benefit from bugfixes.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -156,7 +156,13 @@ showEvent lock e = withLock lock $ print e
 
 cradleToSession :: Cradle -> IO HscEnvEq
 cradleToSession cradle = do
-    opts <- either throwIO return =<< getCompilerOptions "" cradle
+    cradleRes <- getCompilerOptions "" cradle
+    opts <- case cradleRes of
+        CradleSuccess r -> pure r
+        CradleFail err -> throwIO err
+        -- TODO Rather than failing here, we should ignore any files that use this cradle.
+        -- That will require some more changes.
+        CradleNone -> fail "'none' cradle is not yet supported"
     libdir <- getLibdir
     env <- runGhc (Just libdir) $ do
         _targets <- initSession opts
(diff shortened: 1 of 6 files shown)


mpickering:
Note that anything configured with the multi-cradle will also not work with `ghcide` because you pass the empty path when trying to determine which component to use.

cocreature:
Right but that was already the case before. Multicomponent cradles definitely don’t work at the moment.

symbiont-eric-torreborre:
> Right but that was already the case before. Multicomponent cradles definitely don’t work at the moment.

This deserves a bigger note in the release notes, I was wondering what I was doing wrong :-). Is it still possible to load only the modules from one of the components though?

cocreature:
If you set your `hie.yaml` to a cabal cradle that loads one component it should work fine (see  https://github.com/mpickering/hie-bios#explicit-configuration). You can also manually construct a direct cradle that merges the components and sets things like default extensions and include paths manually (getting packages can be a bit tricky but the ghc environment files that cabal writes by default should be picked up).

domenkozar:
I think previously by default stack would load lib and exe components, whereas with the upgrade I can only specify one component (since multi-cradle isn't supported yet). Is that correct observation?

cocreature:
As far as I know `cradle: { stack: {}}` should be equivalent to the stack cradle on older `hie-bios` versions.

domenkozar:
That's what I thought, but hie-bios fails since stack asks to pick the component between lib, exes and tests.

cocreature:
Not sure what is going on then. Maybe worth opening an issue in `hie-bios` for that.

fendor:
We no longer pass in filepaths to `stack repl` but components which likely explains this behaviour. If you are not using a component, just `stack repl` is used, which asks to pick the component if there are multiple exes, for example.

cocreature:
We never did pass a file path in `ghcide`, we just set that to an empty string so I’m still struggling to see what changed here. Could it be that this never worked if you have more than one main module?

PR #260, opened 2019-12-16
Refactor hover and go-to-definition searching

jacg:
The process of searching for definitions is similar to the process of searching
for hover information. In the original code (much of which was written out twice
with occasional stylistic differences) the signal to noise ratio seemed pretty
poor.

Here is a refactoring which aims to make it easier to see the similarities and
differences between these two related functionalities.

I'm not sure about how this should be organized in modules: originally, this code was spread across two modules `Development.IDE.LSP{Hover,Definition}`, half of which was boilerplate (mostly imports) repeated in both. After the rewrite, about 80% of each module would be boilerplate repeated in both, and it wasn't obvious to me where to put the common code. So I have replaced the two original `Definition` and `Hover` modules with a single `HoverDefinition` module. Not sure whether this is the right thing to do.

More generally, I'm not sure whether this sort of change is considered desirable in the culture of this project: please advise.


ndmitchell:
As for culture, my culture (and I think shared by most of DA) is to aggressively refactor until the code is as beautiful as it can be. Don't break externally visible API's unnecessarily, but do so when the benefits are worth it. If it's not an external API and you can do it 5% better, it's worth it.

cocreature:
I completely agree with @neil-da. Refactoring is crucial to keep things maintainable. If we can avoid breaking changes that’s great but if we can’t I’m usually happy to accept them. Even more so in the context of `ghcide` which has basically only one user that uses it as a library (DAML) and we can usually deal with breaking changes reasonably easily.

jacg:
> aggressively refactor until the code is as beautiful as it can be

Excellent! But beauty is in the eye of the beholder, which may be influenced by culture among other things.

> Refactoring is crucial to keep things maintainable.

Great to hear. I'll take that as licence to spam your inboxes with refactoring PRs :-)

On that note, I've been refactoring to try to unify (and understand) the hover and gotoDef functionality as much as possible, and it's this process that led me to the discovery of the solution to the missing-type-level-hover-info problems. I would naively hope that finding the spans present in the code is essentially the same process for both hover and gotoDef, and that the main difference between them is what you do with the information thus found. From this (naive) perspective, I would expect the following two rules to be identical, except for their last lines:

https://github.com/digital-asset/ghcide/blob/7e18f84f81e1d82031007777a567a1d445acea82/src/Development/IDE/Core/Rules.hs#L101-L108

https://github.com/digital-asset/ghcide/blob/7e18f84f81e1d82031007777a567a1d445acea82/src/Development/IDE/Core/Rules.hs#L110-L117

Can you help me understand the reasons for the differences between 

https://github.com/digital-asset/ghcide/blob/7e18f84f81e1d82031007777a567a1d445acea82/src/Development/IDE/Core/Rules.hs#L106-L107

https://github.com/digital-asset/ghcide/blob/7e18f84f81e1d82031007777a567a1d445acea82/src/Development/IDE/Core/Rules.hs#L115-L116

?

cocreature:
The code here shouldn’t make a difference. Typechecking `file` will automatically typecheck all its dependencies. So the only difference seems to be that we pass in the list of typechecked modules to `atPoint` but not to `gotoDefinition`. `atPoint` only uses them for the (currently broken) `getDocumentation` functionality so I don’t think this explains the issue.

PR #261, opened 2019-12-16
Filter not-enabled warnings completely

cocreature:
I’ve tried out the new behavior added in #232 and it is extremely
noisy, in particular, the warnings enabled by
Opt_WarnMissingLocalSignatures which I don’t want to see. I think for
now it makes more sense to disable them completely. We still show the
codelenses which is somewhat on the edge but at least they don’t
clutter the problems tab/CLI output so I’m fine with keeping those
even if they are not enabled.

cc @serras

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -108,14 +108,15 @@ typecheckModule (IdeDefer defer) packageState deps pm =
                 GHC.typecheckModule $ enableTopLevelWarnings
                                     $ demoteIfDefer pm{pm_mod_summary = tweak modSummary'}
             tcm2 <- mkTcModuleResult tcm
-            let errorPipeline = unDefer
-                              . (if wopt Opt_WarnMissingSignatures dflags
-                                    then id
-                                    else degradeError Opt_WarnMissingSignatures)
-                              . (if wopt Opt_WarnMissingLocalSignatures dflags
-                                    then id
-                                    else degradeError Opt_WarnMissingLocalSignatures)
-            return (map errorPipeline warnings, tcm2)
+            let errorPipeline =
+                    fmap unDefer .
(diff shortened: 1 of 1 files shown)


serras:
The behaviour you are proposing seems completely reasonable to me.

ndmitchell:
Why not filter Info diagnostics right at the very edge, in the place we send them to VS code. We never want to sent info level stuff out, and that one place seems much easier to intercept. Plus it may be easier to filter out things like unused variables which we might surface as changing colour.

cocreature:
I don’t think that makes sense. Info level diagnostics are a reasonable thing to sent in general. I just don’t think we should send diagnostics for warnings that the user hasn’t enabled regardless of whether they are info level or some other level.

ndmitchell:
I don't see how this patch works. Don't the code lenses grab the diagnostics for a file to see what needs adding? Haven't you eliminated that diagnostic? Or did the code lens get changed?

cocreature:
ugh, you’re right :cry:  And we don’t have any tests for this :sob: In that case I’m tempted to pull this whole automatic enabling of warnings out completely. In the current state this is unusable for DAML and unusable for me personally and having a global filter goes exactly in the direction of adding too much complexity for little benefit that I was afraid off.

ndmitchell:
Of course we should add tests. And I'm not saying we should send info level warnings - merely that we currently know there are no legit info level warnings, so a simple and effective strategy is to squash info level warnings as they leave the client.

cocreature:
I’d rather pull out automatic enabling of diagnostics than add a hack to filter out info level diagnostics. We can always add it back if we come up with a less hacky solution for this but at the moment I’m not convinced that this brings enough benefits.

cocreature:
Closing since this PR definitely doesn’t make sense in this form.

PR #263, opened 2019-12-16
Filter out completely warnings not enabled by user

serras:
This is another attempt at fixing the problems described by #261. After this patch, warnings which were not enabled explicitly by the user but are needed to support some other functionality (mainly code lenses) are completely filtered out.

The implementation works by keeping two `DiagnosticStore`s in the `IdeState`: the usual one, and a new one of "hidden diagnostics". The means that a call to `getDiagnostics` should do the right thing and show only those warnings which were enabled. You need to call `getHiddenDiagnostics` to obtain the rest of the diagnostics, as done in the code lens code.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -109,12 +109,7 @@ typecheckModule (IdeDefer defer) packageState deps pm =
                                     $ demoteIfDefer pm{pm_mod_summary = tweak modSummary'}
             tcm2 <- mkTcModuleResult tcm
             let errorPipeline = unDefer
-                              . (if wopt Opt_WarnMissingSignatures dflags
-                                    then id
-                                    else degradeError Opt_WarnMissingSignatures)
-                              . (if wopt Opt_WarnMissingLocalSignatures dflags
-                                    then id
-                                    else degradeError Opt_WarnMissingLocalSignatures)
+                              . unShow dflags
             return (map errorPipeline warnings, tcm2)
 
 initPlugins :: GhcMonad m => ModSummary -> m ModSummary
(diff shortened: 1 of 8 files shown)


src/Development/IDE/Core/Compile.hs
```diff
@@ -197,20 +192,16 @@ unDefer (Reason Opt_WarnTypedHoles                 , fd) = upgradeWarningToError
...
-  (nfp, fd{_severity = Just DsError, _message = warn2err $ _message fd}) where
+upgradeWarningToError (nfp, sh, fd) =
+  (nfp, sh, fd{_severity = Just DsError, _message = warn2err $ _message fd}) where
   warn2err :: T.Text -> T.Text
   warn2err = T.intercalate ": error:" . T.splitOn ": warning:"
 
-degradeWarningToError :: FileDiagnostic -> FileDiagnostic
-degradeWarningToError (nfp, fd) =
-  (nfp, fd{_severity = Just DsInfo})
+unShow :: DynFlags -> (WarnReason, FileDiagnostic) -> (WarnReason, FileDiagnostic)
```
cocreature: why call this `unShow` instead of `Hide`?

src/Development/IDE/Types/Diagnostics.hs
```diff
@@ -39,14 +41,20 @@ ideErrorText fp msg = (fp, LSP.Diagnostic {
     _relatedInformation = Nothing
     })
 
+data ShowDiagnostic
+    = ShowDiag | HideDiag
```
cocreature: A comment on what `HideDiag` represents would be great.

src/Development/IDE/Core/Shake.hs
```diff
@@ -656,17 +665,26 @@ updateFileDiagnostics ::
...
         -- publishDiagnosticsNotification.
         newDiags <- modifyVar diagnostics $ \old -> do
-            let newDiagsStore = setStageDiagnostics fp (vfsVersion =<< modTime) (T.pack $ show k) current old
+            let newDiagsStore = setStageDiagnostics fp (vfsVersion =<< modTime)
+                                  (T.pack $ show k) (map snd currentShown) old
+            let newDiags = getFileDiagnostics fp newDiagsStore
+            _ <- evaluate newDiagsStore
+            _ <- evaluate newDiags
+            pure $! (newDiagsStore, newDiags)
+        _newHiddenDiags <- modifyVar hiddenDiagnostics $ \old -> do
```
cocreature: We don’t seem to care about the result so we might as well use `modifyVar_`.

src/Development/IDE/Core/Shake.hs
```diff
@@ -600,7 +609,7 @@ defineOnDisk act = addBuiltinRule noLint noIdentity $
                     pure $ RunResult ChangedNothing (fromMaybe "" current) (isJust current)
                   else do
                     (diags, mbHash) <- runAct
-                    updateFileDiagnostics file (Key key) extras $ map snd diags
+                    updateFileDiagnostics file (Key key) extras $ map (\(_,y,z) -> (y,z)) diags
```
cocreature: Probably better for another PR but this seems to hint at the fact that a record type would be a better choice for `FileDiagnostic` than a 3-tuple.

serras:
I've implemented the suggested changes, except for the tests, which I cannot make run correctly.
In particular, if I use `waitForDiagnostics`, the tests fail with time-out when no warning is shown. On the other hand, if I don't add it, then no code lens in obtained!

serras:
I've written some tests for the code lenses funcionality, but they fail with a very bad message:

```
ghcide-tests: fd:14: hGetLine: illegal operation (handle is closed)
FAIL
Exception: /private/var/folders/hd/mq6r7hc959q33jjgw5xbmvmm0000gn/T/extra-dir-31741191348/Sigs.hs: openFile: does not exist (No such file or directory)
```

Could somebody who knows more about `lsp-test` help? The relevant test is [here](https://github.com/digital-asset/ghcide/pull/263/commits/ca8893471cfb8d6d23bc4438bb5e9d91a4ff5e1e#diff-15add744e546094261d92580aa638eabR718).

jacg:
> if I use `waitForDiagnostics`, the tests fail with time-out when no warning is shown. On the other hand, if I don't add it, then no code lens in obtained!

Have you tried running the test in [`testSessionWait`](https://github.com/digital-asset/ghcide/blob/master/test/exe/Main.hs#L905) rather than `testSession`?

cocreature:
I’ve pushed a fix for the tests. The issue was that you did not wait for the `ApplyWorkspaceEdit` message. That resulted in `lsp-test` shutting down which cleared the VFS map. While shutting down, we received the `ApplyWorkspaceEdit` message (so there was a race condition). At this point, the VFS map was empty so we tried to lookup the file on disc which failed since it is never written to disk.

PR #264, opened 2019-12-16
Define "__GHCIDE__" on CPP

serras:
This patch comes from my experience in using ghcide in combination with quasi-quoters which read from files. If you have something like:

```haskell
$(doSomething "from.file")
```

then build tools would take the relative path to start with the source folder of the corresponding project, but ghcide takes it from the root of the project. Until a better solution is found, my proposal is to have this additional define in CPP, so we can do:

```haskell
#if __GHCIDE__
$(doSomething "src/from.file")
#else
$(doSomething "from.file")
#endif
```

### DIFF
--- a/src/Development/IDE/GHC/CPP.hs
+++ b/src/Development/IDE/GHC/CPP.hs
@@ -114,6 +114,8 @@ doCpp dflags raw input_fn output_fn = do
                     ++ map SysTools.Option sse_defs
                     ++ map SysTools.Option avx_defs
                     ++ mb_macro_include
+        -- Define a special macro "__GHCIDE__"
+                    ++ [ SysTools.Option "-D__GHCIDE__"]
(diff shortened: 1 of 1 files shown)


cocreature:
Do you have an idea of how hard it would be to fix this properly? I would very much like to avoid adding more GHCIDE specific hacks and rather invest into making sure that we agree with GHC.

serras:
My gut feeling is that once we have multi-cradle support from `hie-bios`, we should be able to configure it there.

ndmitchell:
I think it's pretty reasonable to have a macro for ghcide. It's standard practice for most tools - eg HLint defines one. Would be great if no one ever needed it, but that's probably unrealistic.

cocreature:
HLint doesn’t try to be a Haskell compiler so I don’t think this is a fair comparison. That said, I tend to agree that this macro is reasonable so let’s merge this for now.

PR #266, opened 2019-12-17
Normalise filepaths to match haskell-lsp 0.19

cocreature:
haskell-lsp 0.19 has started to normalise file paths completely so we
need to make sure that NormalizedFilePath agrees with that, otherwise,
we get a bunch of test failures on the daml repo (they are not
specific to DAML, but atm ghcide CI does not run windows).

### DIFF
--- a/src/Development/IDE/Types/Location.hs
+++ b/src/Development/IDE/Types/Location.hs
@@ -30,7 +30,6 @@ import Data.Maybe as Maybe
 import Data.Hashable
 import Data.String
 import System.FilePath
-import System.Info.Extra
 import qualified Language.Haskell.LSP.Types as LSP
 import Language.Haskell.LSP.Types as LSP (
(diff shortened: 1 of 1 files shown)


src/Development/IDE/Types/Location.hs
```diff
@@ -50,24 +50,7 @@ instance IsString NormalizedFilePath where
 
 toNormalizedFilePath :: FilePath -> NormalizedFilePath
 toNormalizedFilePath "" = NormalizedFilePath ""
```
aherrmann-da: Is this special case still correct?

cocreature: Yes, otherwise this will be normalized to `"."`which is not what we want.

cocreature: I’ve added a comment to that effect.

PR #267, opened 2019-12-17
Fix #248 and #250

### LINKED ISSUE #248: Hover does not work for type constructors
... as demonstrated by the test titled *type constructor*.

### LINKED ISSUE #249: Hover / goto def do not work for symbols from external packages
... as documented by the tests titled *type constructor from other package* and *value from other package*.

### LINKED ISSUE #250: Hover does not work for classes
... as documented by the tests named *class in instance declaration* and *class in signature*.

jacg:
This fixes hover for types, classes and type variables.

Information about spans includes a `Maybe Type` which is `Just` for data-level
expressions and `Nothing` for type-level expressions.

`AtPoint.atPoint` which is the oddly-named function responsible for constructing
hover information, runs in the `Maybe` monad, and aborted at the first sight of
a `Nothing`, thus producing no hover information for type-level spans.

In the process of fixing this, I have refactored the function to

+ separate the construction of data-level and type-level hover info

+ make the components that make up the hover info (and their construction) more
  clear

I can see plenty little improvements that could be made to the functionality of
the code (and lots that could be made to its organization), but the most
important fixes of the basic missing functionality are here.

Fix #248
Fix #250

### DIFF
--- a/src/Development/IDE/Spans/AtPoint.hs
+++ b/src/Development/IDE/Spans/AtPoint.hs
@@ -46,7 +46,7 @@ gotoDefinition
   -> Position
   -> m (Maybe Location)
 gotoDefinition getHieFile ideOpts pkgState srcSpans pos =
-  locationsAtPoint getHieFile ideOpts pkgState pos srcSpans
+  listToMaybe <$> locationsAtPoint getHieFile ideOpts pkgState pos srcSpans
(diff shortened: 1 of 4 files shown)


jacg:
Issues #273 and #274 opened as requested.

Issue #273, opened 2019-12-18
Display kinds in hover info for type-level symbols


jacg:
In other words, the TODO at https://github.com/digital-asset/ghcide/blob/bbb75c2f95ca209db32defb2c10972968d4c829d/src/Development/IDE/Spans/AtPoint.hs#L69

jneira:
The kind is already displayed in type info so closing

PR #275, opened 2019-12-18
Add tests for #237: kinds in hover info

jacg:


jacg:
The original title incorrectly identified the issue as 237, but in reality it is #273!

PR #286, opened 2019-12-19
Require hie-bios 0.3.2 or above

### LINKED ISSUE #281: hie-bios-0.3 has Windows issues
I've run into at least https://github.com/mpickering/hie-bios/issues/110 and https://github.com/mpickering/hie-bios/pull/108. It's completely screwed, and I'd be surprised if this didn't appear as ghcide bug reports in the near future, so having a ticket to dedupe them all to, and to inform the ghcide people.

ndmitchell:
Lower versions are pretty horribly broken on Windows. Require a fixed hie-bios to avoid support questions. Closes #281.

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -163,7 +163,7 @@ executable ghcide
         ghc-paths,
         ghc,
         haskell-lsp,
-        hie-bios >= 0.3 && < 0.4,
+        hie-bios >= 0.3.2 && < 0.4,
         ghcide,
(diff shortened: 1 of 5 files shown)


cocreature:
Thanks! You need to update the `stack.yaml` files to include the new version.

ndmitchell:
Urk. I might wait until hie-bios-0.3.2 makes it onto Stackage, which is hopefully soon.

PR #288, opened 2019-12-20
Better docs for completions

serras:
This PR makes completion search for documentation using the [`getDocs` function](https://downloads.haskell.org/~ghc/latest/docs/html/libraries/ghc-8.8.1/GHC.html#v:getDocs).


serras:
There's something very strange: the new search for documentation works when I run `ghcide` in VS Code, but tests fail, and I cannot find a reason why.

serras:
Once this is solved, the next step is [reading the Haddock markup](http://hackage.haskell.org/package/pandoc-2.9/docs/Text-Pandoc-Readers-Haddock.html) and [producing Markdown](http://hackage.haskell.org/package/pandoc-2.9/docs/Text-Pandoc-Writers-Markdown.html) with the help of `pandoc`.

serras:
I'll do it on Monday, once I'm back in action.

For the Haddock to Markdown conversion: do you know of any alternative to
Markdown? Would it be OK to depend on Haddock-the-library for parsing the
docs and then turning them to Markdown by hand?

cocreature:
Depending on the haddock library would be fine. Alternatively, we could also go for something a bit more hacky and just do a bunch of search and replace statements to approximate the conversion.

serras:
The part about removing unused JSON instances is now #305. In this PR I have reverted those changes, and also moved from Pandoc to a hand-written Haddock-to-Markdown conversion.

It should now be ready for review. Still, I would suggest to merge #305 first, and then work on merging both branches as part of this PR, since there are some conflicts in the tests between them.

test/exe/Main.hs
```diff
@@ -1061,31 +1061,33 @@ completionTests
...
   where
-    complItem label xdata kind = CompletionItem
+    dropDocs :: CompletionItem -> CompletionItem
+    dropDocs ci = ci { _documentation = Nothing }
+    complItem label xdata kind ty = CompletionItem
       { _label = label
       , _kind = kind
-      , _detail = Just "Prelude"
-      , _documentation = Just (CompletionDocMarkup (MarkupContent {_kind = MkMarkdown, _value = ""}))
+      , _detail = (":: " <>) <$> ty
```
cocreature: This should also check `optNewColonConvention`.

test/exe/Main.hs
```diff
@@ -1061,31 +1061,33 @@ completionTests
...
         expectDiagnostics [ ("A.hs", [(DsWarning, (2, 0), "not used")]) ]
         changeDoc docId [TextDocumentContentChangeEvent Nothing Nothing $ T.unlines ["{-# OPTIONS_GHC -Wunused-binds #-}", "module A () where", "f = Prelude.hea"]]
         compls <- getCompletions docId (Position 2 15)
-        liftIO $ compls @?= [complItem "head" ["GHC.List", "base", "v", "head"] (Just CiFunction)]
+        liftIO $ map dropDocs compls @?= [complItem "head" ["GHC.List", "base", "v", "head"] (Just CiFunction) (Just "[a] -> a")]
     ]
   where
-    complItem label xdata kind = CompletionItem
+    dropDocs :: CompletionItem -> CompletionItem
+    dropDocs ci = ci { _documentation = Nothing }
```
cocreature: Instead of dropping docs, why don’t we check them in the tests? That seems much more useful.

src/Development/IDE/Core/Completions.hs
```diff
@@ -187,8 +195,8 @@ occNameToComKind oc
 
 mkCompl :: CompItem -> CompletionItem
 mkCompl CI{origName,importedFrom,thingType,label,isInfix,docs} =
-  CompletionItem label kind (Just $ maybe "" (<>"\n") typeText <> importedFrom)
-    (Just $ CompletionDocMarkup $ MarkupContent MkMarkdown $ T.intercalate sectionSeparator docs)
+  CompletionItem label kind ((":: "<>) <$> typeText)
```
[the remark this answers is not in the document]
serras: Done in the last commit

test/exe/Main.hs
serras: Since the tests would get very big, I'm checking that the first 100 characters match.

test/exe/Main.hs
serras: I am not sure how to do this, since I don't know where the tests set this flag.

cocreature: Oh I missed that this is a test. Fine to leave it in that case since the test will always use double colons.

cocreature:
Looks like GHC 8.4 is failing due to `-Werror` and GHC 8.8 is failing due to `haddock-library` not being in stackage which can hopefully be solved by adding that to `stack88.yaml`.

serras:
I'll try to fix both problems tonight. Thanks for the review!

serras:
Finally everything is green!

test/exe/Main.hs
```diff
@@ -1243,7 +1243,7 @@ completionTests
         compls <- getCompletions docId (Position 1 7)
         liftIO $ map dropDocs compls @?= 
           [complItem "head" (Just CiFunction) (Just "[a] -> a")
-                     "*Defined in 'Prelude'*\n*\t*\t*\n\n\nExtract the first element of a list, which must be non-empty."]
```
cocreature: This doesn’t actually check anything about the docs. The “Defined in Prelude” part is completely independent of this PR (although this PR slightly changes the wording). Let’s please check that the docs actually work. Some CPP to not check this on older GHCs is totally fine.

serras: New tests are in!

Issue #45, opened 2019-09-12
bug in "remove import" code action when some symbols used


hsenag:
This should be a fairly easy beginner task, at least to stop the incorrect action from being shown.

If you have a redundant symbol imported from a module where the module is still needed, e.g. the obvious code generating this error:
```
C:\msys64\home\ganesh\temp\x\Bar.hs:3:1: warning:
    The import of `func3' from module `Foo' is redundant
```
Then the code action suggests deleting the entire import of `Foo`, which is wrong.

The simplest fix would be to simply tighten up the recognition [here](https://github.com/digital-asset/ghcide/blob/f4553757ddf92b77cfbca9f4daa2bcc9e8e86d28/src/Development/IDE/LSP/CodeAction.hs#L46-L52).

But ideally we'd instead get the appropriate action in this case, namely to just delete the unused symbols. That would probably be tricky to get perfect, because of things like data constructors imported with a type etc, but at least some simple cases could be covered. It might involve making a rudimentary effort to parse the import declaration.

It would also be good to have some tests of this action. There weren't any tests of code actions when it was initially written, but there are some now, [here](https://github.com/digital-asset/ghcide/blob/f4553757ddf92b77cfbca9f4daa2bcc9e8e86d28/test/exe/Main.hs#L211).

hsenag:
There are also now tests of the code actions, so they just need updating rather than writing from scratch: see fdefb6bc.

chshersh:
If I understood correctly from the code, the redundant import is detected by parsing the text of the error message. I'm a bit concerned by this solution because I had enough problems in my life with parsing the somewhat fragile output of error messages and maintaining this parser through multiple versions. I'm working on the `smuggler` tool which removes all unused imports automatically and does it quite successfully. And this tool is implemented via Source Plugins and reuses as much logic from GHC as possible regarding detecting unused imports (and also contains a lot of tests):

* https://github.com/kowainik/smuggler

I believe that using Source Plugins is a better architectural solution for detecting and removing unused imports. This plugin in the future can be extended to support actions like:

1. Convert implicit imports to explicit
2. Compress multiple import definitions into fewer ones
3. Remove duplicate imports
4. Add suggested imports

[...]

PR #290, opened 2019-12-22
Fix for #45 - remove redundant symbols from imports

pepeiborra:
To parse and edit module imports I went with a very low tech solution, instead of reusing the GHC parse tree or similar, in the interest of preserving as much of the user text as possible.

To match on the GHC error output I went with *regex-tdfa*, which is able to run regular expressions on `Text` values, because it seems the right tool for the job. While it has an upper bound on `base` that rules out 8.8, it worked for me with `allow-newer`.

### DIFF
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -327,7 +327,7 @@ dropBindingsFromImportLine bindings_ importLine =
 
       filtering x = case () of
         () | x `elem` bindings -> Nothing
-        () | x `elem` (map (<> ")") bindings) -> Just ")"
+        () | x `elem` map (<> ")") bindings -> Just ")"
         _                      -> Just x
 
       joinCloseParens (x : ")" : rest) = (x <> ")") : joinCloseParens rest
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -88,7 +88,7 @@ executeAddSignatureCommand _lsp _ideState ExecuteCommandParams{..}
 suggestAction :: Maybe T.Text -> Diagnostic -> [(T.Text, [TextEdit])]
 suggestAction contents diag@Diagnostic{_range=_range@Range{..},..}
 --     The qualified import of ‘many’ from module ‘Control.Applicative’ is redundant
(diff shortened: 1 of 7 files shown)


src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -293,6 +299,47 @@ textInRange (Range (Position startRow startCol) (Position endRow endCol)) text =
...
+      importRest' =
+        T.intercalate ","
+          $ joinCloseParens
+          $ mapMaybe (filtering . T.strip)
+          $ T.splitOn ","
+          $ T.tail importRest
+
+      filtering x = case () of
+        () | x `elem` bindings -> Nothing
+        () | x `elem` (map (<> ")") bindings) -> Just ")"
```
aherrmann-da:
```suggestion
        () | x `elem` map (<> ")") bindings -> Just ")"
```
Redundant parentheses

stack.yaml
```diff
@@ -7,5 +7,6 @@ extra-deps:
 - lsp-test-0.9.0.0
 - hie-bios-0.3.0
 - fuzzy-0.1.0.0
+- regex-tdfa-1.3.1.0
```
aherrmann-da:
```suggestion
- regex-base-0.94.0.0
- regex-tdfa-1.3.1.0
```

stack84.yaml
```diff
@@ -12,6 +12,7 @@ extra-deps:
 - js-dgtable-0.5.2
 - hie-bios-0.3.0
 - fuzzy-0.1.0.0
+- regex-tdfa-1.3.1.0
```
aherrmann-da:
```suggestion
- regex-base-0.94.0.0
- regex-tdfa-1.3.1.0
```

stack88.yaml
```diff
@@ -7,6 +7,7 @@ extra-deps:
 - lsp-test-0.9.0.0
 - hie-bios-0.3.0
 - fuzzy-0.1.0.0
+- regex-tdfa-1.3.1.0
```
aherrmann-da:
```suggestion
- regex-base-0.94.0.0
- regex-tdfa-1.3.1.0
```

pepeiborra:
Suggestions applied, thanks!

aherrmann-da:
@pepeiborra Thank you for fixing all these and for implementing the feature.

Issue #101, opened 2019-09-20
Performance of hover seems very poor


mpickering:
Compared to hover in `haskell-ide-engine` I think the performance of hovering over a symbol seems very poor. It seems to take up to 1s to process the request when the information can be precomputed for an entire module.

mpickering:
Reload performance is very good though :+1:

cocreature:
Hover works in two steps:
1. We have a rule for getting a list of source spans from a module https://github.com/digital-asset/ghcide/blob/32a049cc4c61961088cdd74ea0ebaa9d4ce596ff/src/Development/IDE/Core/Rules.hs. This rule is cached so if is slow on the second hover in the same module (without modifications) this shouldn’t be the bottleneck.
2. Based on that list, we then filter it to the ones at a given location, see https://github.com/digital-asset/ghcide/blob/32a049cc4c61961088cdd74ea0ebaa9d4ce596ff/src/Development/IDE/Spans/AtPoint.hs#L58. That filtering is an incredibly stupid search that is linear in the number of source spans. A very easy improvement here would be to replace the list by an interval map.

ndmitchell:
The code is also used for goto definition, and there it doesn't seem great at finding all the relevant definitions either. This piece seems something that hie, intero and ghcide should all be able to share - something like:

```
findIdentifierAtLocation :: Module -> (Location -> [Identifier])
```

Where you do meaningful caching after seeing the module. I think the ghcide code originally came in from Intero, but then got tweaked a fair bit. Is there scope for sharing? Who does it best? Could the hie code become hie-spans and then be shared/improved by all?

mpickering:
Here is the code which generates the `TypeMap` in `haskell-ide-engine`. 

https://github.com/haskell/haskell-ide-engine/blob/ff9a5b2e85a0910205cb781bf630503efe6bd227/hie-plugin-api/Haskell/Ide/Engine/TypeMap.hs#L1

Definition map:

https://github.com/haskell/haskell-ide-engine/blob/6254c09fbc909dca2673ad7a27fe082dc75d6fc9/hie-plugin-api/Haskell/Ide/Engine/ArtifactMap.hs#L119

mpickering:
Also another implementation of a definition map from my branch with less CPP using pattern synonyms.

https://github.com/mpickering/haskell-ide-engine/blob/hie-bios/hie-plugin-api/Haskell/Ide/Engine/ArtifactMap.hs#L88

[...]

PR #292, opened 2019-12-29
Speed up span lookups

### LINKED ISSUE #247: Hover / goto def do not work for data constructors in record construction
... as demonstrated by the test with title *record data constructor*.

pepeiborra:
Following the first part of the discussion in #101, 
and reusing some code from https://github.com/mpickering/ghcide/commits/wip/type-map.

An interval tree is used to speed up span lookups, which benefits hovers and
definition lookups. The computation of spans is still done using the same
algorithm as previously.

Fixes #247 by sorting matching spans by relevance.

### DIFF
--- a/src/Development/IDE/Spans/AtPoint.hs
+++ b/src/Development/IDE/Spans/AtPoint.hs
@@ -142,12 +142,16 @@ spansAtPoint :: Position -> [SpanInfo] -> [SpanInfo]
 spansAtPoint pos = filter atp where
   line = _line pos
   cha = _character pos
-  atp SpanInfo{..} =    spaninfoStartLine <= line
-                     && spaninfoEndLine >= line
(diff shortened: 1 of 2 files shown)


domenkozar:
@pepeiborra this is great! Do you have some measurements of improvement?

pepeiborra:
Very hard to measure, due to https://github.com/digital-asset/ghcide/pull/254#issuecomment-569924242

domenkozar:
@pepeiborra seems like that's fixed.

pepeiborra:
No real speed up observed. Fail.

Looking up spans in a list is (unsurprisingly) way less expensive than type checking; I suspect that a module with enough spans to make the lookup operation noticeably slow would run out of memory during type checking.

I'll revert the interval tree change and keep just the fix for #247

PR #293, opened 2019-12-29
Document symbols provider

pepeiborra:
Low hanging fruit, worth doing as VS Code makes document outlines actually useful by adorning them with warnings and errors markers.

### DIFF
--- a/src/Development/IDE/LSP/Outline.hs
+++ b/src/Development/IDE/LSP/Outline.hs
@@ -1,0 +1,185 @@
+{-# LANGUAGE DuplicateRecordFields #-}
+module Development.IDE.LSP.Outline
+  ( setHandlersOutline
+  )
+where
+
+import qualified Language.Haskell.LSP.Core     as LSP
+import           Language.Haskell.LSP.Messages
+import           Language.Haskell.LSP.Types
+import           Data.Functor
+import           Data.Generics
+import           Data.Maybe
+import           Data.Text                      ( Text
+                                                , pack
+                                                )
+import qualified Data.Text                     as T
+import           Development.IDE.Core.Rules
+import           Development.IDE.Core.Shake
+import           Development.IDE.GHC.Error      ( srcSpanToRange )
+import           Development.IDE.LSP.Server
+import           Development.IDE.Types.Location
+import           GHC
+import           Outputable                     ( Outputable
+                                                , ppr
+                                                , showSDocUnsafe
+                                                )
+
+setHandlersOutline :: PartialHandlers
+setHandlersOutline = PartialHandlers $ \WithMessage {..} x -> return x
+  { LSP.documentSymbolHandler = withResponse RspDocumentSymbols moduleOutline
+  }
+
+moduleOutline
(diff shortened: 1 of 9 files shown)


src/Development/IDE/LSP/Outline.hs
```diff
@@ -0,0 +1,194 @@
...
+      pure $ case mb_decls of
+        Nothing -> DSDocumentSymbols (List [])
+        Just (ParsedModule { pm_parsed_source = L _ltop HsModule { hsmodName, hsmodDecls, hsmodImports } })
+          -> let
+               declSymbols  = mapMaybe documentSymbolForDecl hsmodDecls
+               moduleSymbol = hsmodName <&> \(L l m) ->
+                 (defDocumentSymbol l :: DocumentSymbol)
+                   { _name  = pprText m
+                   , _kind  = SkFile
+                   , _range = Range (Position 0 0) (Position 10000000 0) -- _ltop is 0 0 0 0
```
aherrmann-da:
```suggestion
                   , _range = Range (Position 0 0) (Position maxBound 0) -- _ltop is 0 0 0 0
```
I'm guessing 10000000 is just meant to be larger than any actual file. Would `maxBound` be more meaningful?

src/Development/IDE/LSP/Outline.hs
```diff
@@ -0,0 +1,194 @@
...
+    , _kind = SkInterface
+    }
+documentSymbolForDecl (L l (DerivD DerivDecl { deriv_type })) =
+  gfindtype deriv_type <&> \(L (_ :: SrcSpan) name) ->
+    (defDocumentSymbol l :: DocumentSymbol) { _name = pprText @(HsType GhcPs)
+                                              name
+                                            , _kind = SkInterface
+                                            }
+documentSymbolForDecl (L l (ValD x)) =
+  gfindtype x <&> \(L (_ :: SrcSpan) name) ->
```
aherrmann-da: Bindings like `foo :: Int = 42` or `Just foo = Just 21` seem to be skipped. Maybe that's due to the use of `gfindtype`?

src/Development/IDE/LSP/Outline.hs
```diff
@@ -0,0 +1,194 @@
...
+    (defDocumentSymbol l :: DocumentSymbol)
+      { _name   = showRdrName name
+      , _kind   = case x of
+                    PatBind{} -> SkConstant
+                    _         -> SkFunction
```
aherrmann-da: I'm not sure the distinction between `SkConstant` and `SkFunction` makes sense at this point. As I understand, a binding like `Just plus2 = Just (+ 2)` would be qualified as `SkConstant`, while `two = 2` would be qualified as `SkFunction`. Sadly, LSP doesn't seem to have `SkValue`, but it seems more to correct to consistently use either `SkFunction` or `SkConstant`.

src/Development/IDE/LSP/Outline.hs
pepeiborra: Good point, thanks!

src/Development/IDE/LSP/Outline.hs
pepeiborra: Should be fixed now

PR #295, opened 2019-12-30
Insert imports code action

pepeiborra:
Initial stab at a code action to insert imports based on GHC suggestions. Outstanding issues  (to my knowledge):

- the suggestions for missing constructors lead to compile errors when applied, see below. To deal with this,  a second code action is provided to fix these imports.
```
    In module ‘Data.Maybe’:
      ‘Just’ is a data constructor of ‘Maybe’
    To import it use
      import Data.Maybe( Maybe( Just ) )
    or
      import Data.Maybe( Maybe(..) )
```
- If a missing constructor has the same name as the defining data type (e.g. `Identity`), neither code action can deal with it properly.

I'd say that this PR gets ghcide 80% there. I will add tests for these outstanding issues later today.

### DIFF
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -92,6 +92,7 @@ suggestAction text diag = concat
     , suggestExtendImport text diag
     , suggestFillHole diag
     , suggestFillTypeWildcard diag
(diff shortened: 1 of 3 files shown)


pepeiborra:
This is ready for review

PR #296, opened 2019-12-30
Collect CPP error logs into diagnostics.

jinwoo:
Fixes https://github.com/digital-asset/ghcide/issues/87

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE CPP #-}
-{-# LANGUAGE ScopedTypeVariables #-}
 #include "ghc-api-version.h"
 
 module Main (main) where
@@ -1024,7 +1023,10 @@ cppTests =
     -- Some give the column number and others don't (hence -1). Assert either
     -- of them.
     (run $ expectError content (2, -1))
-      `catch` (\(_ :: HUnitFailure) -> run $ expectError content (2, 1))
+      `catch` ( \e -> do
+                  let _ = e :: HUnitFailure
+                  run $ expectError content (2, 1)
+              )
   where
     expectError :: T.Text -> Cursor -> Session ()
     expectError content cursor = do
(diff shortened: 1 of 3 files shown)


jinwoo:
The source location of the CPP error from CI is different from what I see locally. It is (2, 1) locally but (2, -1) in CI. I'll change the column number to `-1` but I'm worried that it'd make `cppTests` fail locally.

test/exe/Main.hs
```diff
@@ -977,6 +978,30 @@ pluginTests = testSessionWait "plugins" $ do
...
+          ]
+  _ <- openDoc' "Testing.hs" "haskell" content
+  expectDiagnostics
+    [ ( "Testing.hs",
+        [ ( DsError,
+            (2, -1),
+            T.unlines
+              [ " error: unterminated #ifdef",
+                " #ifdef FOO",
+                " ^"
```
cocreature: My CPP doesn’t seem to have the carret in the output here so it would probably be good to make this test a bit more fuzzy and only assert on `error: unterminated #ifdef` which is hopefully present in the output of most if not all CPP implementations.

test/exe/Main.hs
```diff
@@ -977,6 +978,30 @@ pluginTests = testSessionWait "plugins" $ do
...
+          [ "{-# LANGUAGE CPP #-}",
+            "module Testing where",
+            "#ifdef FOO",
+            "foo = 42"
+          ]
+  _ <- openDoc' "Testing.hs" "haskell" content
+  expectDiagnostics
+    [ ( "Testing.hs",
+        [ ( DsError,
+            (2, -1),
```
cocreature: I get the same line numbers as CI but a different failure as mentioned below. My guess is that this is because we’re using different CPP implementations (I’m using whatever the default is on linux when invoked via stack, I guess you might be on Windows or MacOS). I would like to avoid this test being flaky locally so maybe we can just test against both `(2,-1)` and `(2,0)` (by catching `HUnitFailure` on the first try) and succeed if either of them succeeds?

test/exe/Main.hs
jinwoo: Reduced to `error: unterminated` because the error message on my mac is `error: unterminated conditional directive`.

test/exe/Main.hs
jinwoo: Yeah, I'm doing this on my mac at the moment. What about this?

PR #303, opened 2020-01-03
Fix performance regression introduced by filepath normalisation

### LINKED ISSUE #298: performance degradation after upgrade to haskell-lsp 0.19
I've noticed a hover performance degradation in head, and bisected it to this change. Hover takes twice as long for me.

_Originally posted by @pepeiborra in https://github.com/digital-asset/ghcide/pull/254#issuecomment-569924242_

cocreature:
We already normalise filepaths in NormalizedFilePath. haskell-lsp
changed things such that the conversion from Uri to NormalizedUri
normalises the filepath again which caused a significant slowdown in
GetFileExists.

We already have a wrapper for converting from NormalizedFilePath to
NormalizedUri so this PR changes this wrapper to inline the definition
without the additional layer of normalisation.

fixes #298

The corresponding `haskell-lsp` PR that caused the regression was https://github.com/alanz/haskell-lsp/pull/205

Before
Screenshot from 2020-01-03 15-35-32
After
Screenshot from 2020-01-03 15-35-56

### DIFF
--- a/src/Development/IDE/Types/Location.hs
+++ b/src/Development/IDE/Types/Location.hs
@@ -29,7 +29,12 @@ import Data.Binary
 import Data.Maybe as Maybe
 import Data.Hashable
 import Data.String
+import qualified Data.Text as T
+import Network.URI
 import System.FilePath
+import qualified System.FilePath.Posix as FPP
+import qualified System.FilePath.Windows as FPW
+import System.Info.Extra
 import qualified Language.Haskell.LSP.Types as LSP
 import Language.Haskell.LSP.Types as LSP (
     filePathToUri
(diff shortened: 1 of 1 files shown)


jneira:
Sorry for the inconveniences. I added the extra normalization step to ensure the file uris were correct even if they come from the aeson instance. I assumed naively that the performance would not be so worse.
Check if uris are correct had suffer a similar penalty, i think.
To avoid the duplication we could expose another function in haskell-lsp without the extra step, if you agree.

cocreature:
No worries, these things happen and once #101 is addressed this shouldn’t be a bottleneck anymore. I wonder if it might make sense to move `NormalizedFilePath` to `haskell-lsp`. Then we could expose `NormalizedFilePath -> NormalizedUri` which can safely skip normalization.

PR #307, opened 2020-01-05
Add a new flag --shake-profiling DIR

pepeiborra:
The flag provides a way to enable Shake profiling reports without recompiling.
Debug output prints links to the profiling artifacts for convenience.

### DIFF
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -53,9 +53,10 @@ import qualified Data.ByteString.Internal as BS
 import           Data.Dynamic
 import           Data.Maybe
 import Data.Map.Strict (Map)
-import           Data.List.Extra
+import           Data.List.Extra (foldl', partition, takeEnd)
 import qualified Data.Set as Set
 import qualified Data.Text as T
+import Data.Traversable (for)
 import Data.Tuple.Extra
 import Data.Unique
 import Development.IDE.Core.Debouncer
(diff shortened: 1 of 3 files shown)


src/Development/IDE/Core/Shake.hs
```diff
@@ -392,9 +395,12 @@ shakeRun IdeState{shakeExtras=ShakeExtras{..}, ..} acts =
                   let res' = case res of
                           Left e -> "exception: " <> displayException e
                           Right _ -> "completed"
+                      profile = case res of
+                          Right (_, Just fp) -> ", profile saved at file://" <> fp
```
cocreature: This isn’t going to produce a valid URI in general, if you want the URI (I guess to be able to click on it in VSCode?), you can use `filePathToUri'` or alternatively, just print out the filepath.

src/Development/IDE/Core/Shake.hs
```diff
@@ -227,14 +227,17 @@ data IdeState = IdeState
...
-shakeRunDatabaseProfile :: Maybe FilePath -> ShakeDatabase -> [Action a] -> IO [a]
+shakeRunDatabaseProfile :: Maybe FilePath -> ShakeDatabase -> [Action a] -> IO ([a], Maybe FilePath)
 shakeRunDatabaseProfile mbProfileDir shakeDb acts = do
         (time, (res,_)) <- duration $ shakeRunDatabase shakeDb acts
-        whenJust mbProfileDir $ \dir -> do
-            count <- modifyVar profileCounter $ \x -> let !y = x+1 in return (y,y)
-            let file = "ide-" ++ profileStartTime ++ "-" ++ takeEnd 5 ("0000" ++ show count) ++ "-" ++ showDP 2 time <.> "html"
-            shakeProfileDatabase shakeDb $ dir </> file
-        return res
+        proFile <- case mbProfileDir of
```
cocreature: this can be simplified a bit with `for` or `traverse`.

pepeiborra:
Thanks for the feedback, I have applied it, rebased on master, and force-pushed.

cocreature:
Thanks for addressing the comments!

PR #309, opened 2020-01-06
Code action to insert new definitions

pepeiborra:
GHC error messages are quite awesome. If only there was a typed API ...

### DIFF
--- a/src/Development/IDE/LSP/CodeAction.hs
+++ b/src/Development/IDE/LSP/CodeAction.hs
@@ -104,7 +104,7 @@ suggestAction ideOptions parsedModule text diag = concat
     , suggestReplaceIdentifier text diag
     , suggestSignature True diag
     ] ++ concat
-    [  suggestNewDefinition ideOptions pm diag
+    [  suggestNewDefinition ideOptions pm text diag
     ++ suggestRemoveRedundantImport pm text diag
     | Just pm <- [parsedModule]]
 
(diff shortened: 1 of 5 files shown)


src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -134,6 +135,22 @@ suggestReplaceIdentifier contents Diagnostic{_range=_range@Range{..},..}
...
+suggestNewDefinition :: Maybe T.Text -> Diagnostic -> [(T.Text, [TextEdit])]
+suggestNewDefinition mb_contents Diagnostic{_message, _range}
+--     * Variable not in scope:
+--         suggestAcion :: Maybe T.Text -> Range -> Range
+    | Just contents <- mb_contents
+    , Just [name, typ] <- matchRegex (unifySpaces _message) "Variable not in scope: ([^ ]*) :: ([^*•]*)"
+    , sig <- name <> " :: " <> T.dropWhileEnd isSpace typ
+    , rangeLine <- _line (_start _range)
+    , contentLines <- T.lines contents
+    , nextEmptyLine <- findIndex T.null (drop rangeLine contentLines)
```
aherrmann-da: Worth noting that this can fail. E.g. in
```
foo :: Bool -> String
foo True = bar "true"

foo False = "false"
```
it will insert the definition of `bar` before the second case of `foo`, which is illegal.

Do we have easy access to the name of the current definition (here `foo`)? If so, you could skip until the next top-level definition with a different name and insert before there.

test/exe/Main.hs
```diff
@@ -1489,6 +1512,25 @@ openTestDataDoc path = do
...
+        [ listToMaybe
+          [ action
+          | CACodeAction action@CodeAction { _title = actionTitle } <- actions
+          , actionTitle == expectedTitle ]
+        | expectedTitle <- expectedTitles]
+  let msg = show $
+            [ actionTitle
+            | CACodeAction CodeAction { _title = actionTitle } <- actions
+            ]
+  liftIO $ assertBool msg (isJust matches)
```
aherrmann-da: This slightly changes the meaning of some of the tests. Before they were checking that the actual list of actions equals the expected list of actions. This seems to check that the actual list of actions is a super set of the expected actions. Also the assertion message now only shows the actual actions and not the expected actions.

Why not keep checking for equality with `@=?`?

pepeiborra: The change is motivated by the list of suggested code actions now including actions to introduce new definitions. I assume that this test doesn't care about those extra actions, and therefore enumerating them here, in order to be able to use `@=?`, did not seem appropriate. 

Thanks for the suggestion to improve the error message, I'll take a look at that.

src/Development/IDE/LSP/CodeAction.hs
pepeiborra: I don't think that the definition is immediately available, but we can extract it from the AST using the range. That should be easy enough, but perhaps I'll wait for #308 to merge to avoid conflicts.

src/Development/IDE/Types/Location.hs
```diff
@@ -129,6 +130,24 @@ noRange =  Range (Position 0 0) (Position 100000 0)
...
 
+isInsideRange :: Position -> RealSrcSpan -> Bool
+p `isInsideRange` r = sp <= p && p <= ep
+  where (sp, ep) = unpackRealSrcSpan r
+
+-- | Converts from 1-based tuple
+toPos :: (Int,Int) -> Position
+toPos (l,c) = Position (l-1) (c-1)
+
+unpackRealSrcSpan :: GHC.RealSrcSpan -> (Position, Position)
```
cocreature: This seems to be pretty much the same as `Development.IDE.GHC.Error.realSrcSpanToRange`.

test/exe/Main.hs
```diff
@@ -1538,6 +1589,27 @@ openTestDataDoc path = do
...
+          | CACodeAction action@CodeAction { _title = actionTitle } <- actions
+          , actionTitle == expectedTitle ]
+        | expectedTitle <- expectedTitles]
+  let msg = show
+            [ actionTitle
+            | CACodeAction CodeAction { _title = actionTitle } <- actions
+            ]
+            ++ "is not a superset of "
+            ++ show expectedTitles
+  liftIO $ assertBool msg (isJust matches)
```
cocreature:
```suggestion
  case matches of
    Nothing -> assertFailure msg
    Just matches -> pure matches
```

jneira:
> GHC error messages are quite awesome. If only there was a typed API ...

@pepeiborra some work in progress: https://github.com/bgamari/ghc-pretty-errors/pull/5/files 😄

Issue #310, opened 2020-01-07
Hover on inner signature gives confusing information


jacg:
Hovering on the *signature* of `inner` in the following sample, gives the signature of `outer`. Hovering on the *definition* of `inner` gives the correct signature of `inner`.

````haskell
outer :: Bool
outer = undefined where

  inner :: Char
  inner = undefined
````

Tests coming up.

jacg:
As discussed in the PR with the tests (#311), hovering on top-level signatures currently does nothing, so for the sake of consistency we should do one of:

+ make hovering on inner signatures also do nothing at all
+ fix the inner signature behaviour to refer to the inner function, but in that case we should make the behaviour on top-level signatures work analogously.

jacg:
In summary, the conclusion we reached in #311 is: hover and gotoDef on both inner and top-level signatures should point to the first clasue of their corresponding definitions. The tests merged in #311 reflect this decision.

PR #311, opened 2020-01-07
Tests for issue 310 (misleading hover on inner signature)

jacg:
Tests for #310.

The most important pair of tests here is the "inner signature" pair. The others
serve mainly to document, compare and contrast what is happening in related
situations.

In summary, hover and gotoDef

+ on inner signatures: give type and location information for the outer
  definition; this is misleading,

+ on outer signatures: give no information at all,

+ on inner definitions: give correct information for the inner definition,

+ on outer definitions: give correct information for the outer definition.

Should hover and gotoDef do anything at all for signatures? or is the current
behaviour for outer signatures (doing nothing at all) what we want?

### DIFF
--- a/test/data/GotoHover.hs
+++ b/test/data/GotoHover.hs
@@ -42,3 +42,9 @@ documented :: Monad m => Either Int (m a)
 documented = Left 7518
 
 listOfInt = [ 8391 :: Int, 6268 ]
+
+outer :: Bool
+outer = undefined where
+
+  inner :: Char
+  inner = undefined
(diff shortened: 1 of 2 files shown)


jacg:
> I'd agree that hover/gotoDef on the signature should point to their corresponding definition (i.e. outer->outer and inner->inner).

Which leaves one small detail: what *exactly* do we mean by "pont to their definition"? Should it point to

1. The signature (which is guaranteed to exist, given that we are considering hovering on a signature)
2. The first clause in the definition

?

aherrmann-da:
> Which leaves one small detail: what _exactly_ do we mean by "pont to their definition"? Should it point to
> 
> 1. The signature (which is guaranteed to exist, given that we are considering hovering on a signature)
> 2. The first clause in the definition
> 
> ?

Right, thanks for pointing that out. In other instances "gotoDef" does (2). In a snippet like the following (2) also seems more useful.
```
a, b :: Int
a = 1
b = 2
```
So, to me (2) seems like the better option.

jacg:
> (2) seems like the better option.

Agreed. I'll adjust the tests to reflect this choice.

jacg:
> I'll adjust the tests to reflect this choice.

Done.

I'm not sure whether it is worth keeping the tests which compare the (working) definition behaviour to the (broken) signature behaviour, so I have added a separate commit which removes the definition tests: we can keep or revert this commit depending on how we feel about those tests.

cocreature:
Thanks! Happy to remove the tests since as you pointed out they are somewhat redundant.

PR #313, opened 2020-01-09
Trigger completion after dot

serras:
This PR makes `ghcide` request a completion every time a dot is pressed. So if you do `Prelude.`, you get the entire list of functions defined in that module.

There's some additional logic to check for the case `id .`, where you want to use `(.)` as the function composition operator. Simply stated, if before the dot we find nothing, and the completion is triggered by that same press of the dot, we give nothing back in return.


serras:
We could be a list more clever, but in cases like `id.` (without a space before the dot) -- where `id` would be taken as the "module" -- we just return no completions because there's no module named `id`. So the logic seems enough to me.

PR #317, opened 2020-01-10
Enhancements to hover

serras:
* Show kinds of known types (part of #273, type variables still missing)
* Show documentation
* Show some literals (part of #274, literal `Text` and lists still missing)


serras:
Some functions had to be moved to a new module to prevent cyclic dependencies.

domenkozar:
Is there some special configuration for showing documentation?

serras:
Not in theory, although you might need to install your libraries with
documentation. Try some Prelude functions, they should work.

Also, the corresponding GHC functionality is only available from 8.4.

serras:
The documentation for `:doc` in GHCi (which uses the same code as we do here to find docs is [here](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/ghci.html#ghci-cmd-:doc).

serras:
@cocreature could we try to get this for release 0.6? This solves a bunch of hover bugs, and complements the showing of docs in completions with showing them on hover.

Unfortunately, it seems that there's some conflict with the solution of #318. For some reason, the fixes break, and I guess this comes from my code for getting the type of literals. Maybe the tests are wrong once we have hovers for literals?

serras:
It turns out that some tests had to be updated. Now everything should work!

cocreature:
Sorry this has to wait for the next release. Postponing releases to get in more stuff only results in them getting delayed further and further and this isn’t a regression compared to 0.0.5 so I don’t think there is a particular reason why this has to be in 0.0.6.

As for the change, I’ll take a detailed look at this PR next week.

serras:
That makes complete sense!

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -54,36 +53,39 @@ atPoint
   -> [SpanInfo]
   -> Position
   -> Maybe (Maybe Range, [T.Text])
-atPoint IdeOptions{..} tcs srcSpans pos = do
+atPoint IdeOptions{..} _ srcSpans pos = do
```
cocreature: If we never use the typechecked modules, there is no point in passing them as an argument.

src/Development/IDE/Spans/Type.hs
```diff
@@ -34,9 +35,11 @@ data SpanInfo =
             -- any. This can be useful for accessing a variety of
             -- information about the identifier such as module,
             -- locality, definition location, etc.
+           ,spaninfoDocs :: ![T.Text]
+           -- ^ Documentation for the element
            }
 instance Show SpanInfo where
-  show (SpanInfo sl sc el ec t n) =
+  show (SpanInfo sl sc el ec t n _docs) =
```
cocreature: Given that the `Show` instance is intended for debugging, we should include all fields in it.

src/Development/IDE/Spans/Type.hs
```diff
@@ -34,9 +35,11 @@ data SpanInfo =
             -- any. This can be useful for accessing a variety of
             -- information about the identifier such as module,
             -- locality, definition location, etc.
+           ,spaninfoDocs :: ![T.Text]
```
cocreature: `SpanInfo` is cached so everything that we store here is retained in memory. The way this is setup currently, you convert the docs for a symbol separately for each module. This means that if module `A` defines symbol `a` and module `B` and `C` reference it, you will end up storing the docs for `a` twice. For such a small example, this is probably fine but on larger projects this seems like something that could cause problems.

I would suggest that we only cache the `HsDocString` here which should be shared and then do the conversion to `Text` (and markdown) on the fly (probably in `atPoint`).

src/Development/IDE/Spans/Calculate.hs
```diff
@@ -107,31 +109,44 @@ ieLNames _ = []
...
+  case mbe of
+    Just expr -> do
+      let ss = getSpanSource' e
+      docs <- case ss of
+                Named n -> getDocumentationTryGhc' tms n
+                _       -> return []
+      return $ Just (ss, getLoc e, Just (CoreUtils.exprType expr), docs)
+    Nothing -> return Nothing
   where
+    getSpanSource' :: LHsExpr GhcTc -> SpanSource
```
cocreature: I think this code would be easier to read if you merge `getSpanSource'` and `getSpanSource`.

src/Development/IDE/Spans/Type.hs
```diff
@@ -47,13 +50,15 @@ instance NFData SpanInfo where
 -- we don't always get a name out so sometimes manually annotating source is more appropriate
 data SpanSource = Named Name
                 | SpanS SrcSpan
+                | Lit SrcSpan String
```
cocreature: I couldn’t figure out where we actually make use of the `SrcSpan` that we store here. Can we just remove it?

We should probably also replace `String` by `Text`.

src/Development/IDE/Spans/Calculate.hs
```diff
@@ -157,40 +175,36 @@ getTypeLPat _ pat =
...
+          return $ name' >>= safeTyThingType
+  let ty' = case ty of
+              Right (Just x) -> Just x
+              _ -> Nothing
+  pure [(Named n, spn, ty', docs)]
 getLHsType _ _ = pure []
 
 importInfo :: [(Located ModuleName, Maybe NormalizedFilePath)]
-           -> [(SpanSource, SrcSpan, Maybe Type)]
+           -> [(SpanSource, SrcSpan, Maybe Type, [T.Text])]
```
cocreature: Given that the docs are always empty, I would prefer to add the empty list at the call site instead of returning it from here.

test/exe/Main.hs
```diff
@@ -1132,11 +1132,11 @@ findDefinitionAndHoverTests = let
   , test yes    yes    mclL37 mcl    "top-level fn 2nd clause         #246"
   , test yes    yes    spaceL37 space "top-level fn on space #315"
   , test no     broken docL41 doc    "documentation                     #7"
-  , test no     broken eitL40 kindE  "kind of Either                  #273"
-  , test no     broken intL40 kindI  "kind of Int                     #273"
+  , test no     yes    eitL40 kindE  "kind of Either                  #273"
```
cocreature: I’m struggling to figure out the part of this PR that adds kind information, could you help me out here?

serras: It starts here https://github.com/digital-asset/ghcide/pull/317/files#diff-9e4907094a15701d642e91bb4b7d5bb7R181

src/Development/IDE/Spans/Type.hs
serras: I'll remove it. I was storing it as we store any other `SrcSpan`, but it turns out literals have to be excluded from all "go to definition" and "go to point" searches, so we end up not using it.

src/Development/IDE/Spans/Type.hs
serras: Good idea! I'll do this in this PR, and then do something similar for completions, where the string may be also shared.

src/Development/IDE/Spans/Calculate.hs
```diff
@@ -157,40 +173,36 @@ getTypeLPat _ pat =
...
+    => [TypecheckedModule]
     -> LHsType GhcRn
-    -> m [(SpanSource, SrcSpan, Maybe Type)]
-getLHsType _ (L spn (HsTyVar U _ v)) = pure [(Named $ unLoc v, spn, Nothing)]
+    -> m [(SpanSource, SrcSpan, Maybe Type, SpanDoc)]
+getLHsType tms (L spn (HsTyVar U _ v)) = do
+  let n = unLoc v
+  docs <- getDocumentationTryGhc tms n
+  ty <- catchSrcErrors "completion" $ do
+          name' <- lookupName n
```
cocreature: `lookupName` doesn’t work on `ghc-lib`. See the CPP for completions to disable this.

moodmosaic:
@serras, this looks great, but some parts of the `:doc` are (incorrectly?) wrapped inside `<code></code>`

image

serras:
@moodmosaic thanks for the catch! The algorithm for transforming Haddock docs into Markdown is quite naïve, so it's quite easy to get a problem. I guess the best next step is to write a bug report and I'll try to have a look at it :)

moodmosaic:
>algorithm for transforming Haddock docs into Markdown is quite naïve

That's fine :+1: If you want, I can open an issue for it, or even try to fix it, just let me know where that (issue) should be opened.

serras:
I'm not a maintainer, but I guess that using the issue tracker for the project is the best thing.

PR #318, opened 2020-01-10
Fix source spans for multi-clause definitions

### LINKED ISSUE #315: unexpected server reply to "textDocument/definition" method when definition can't be found
Thanks for all the work on ghcide, it's really helped my workflow.

When i hover on a function that can't be found with go to definition (e.g. anything external to the project), and I go to definition, my cursor jumps to the outer definition I'm currently at, rather than staying at the same location. I'm using ghcide with emacs and eglot, and I don't fully understand how lsp works, so apologies if this is either expected behavior, or actually a problem on eglot's side (e.g. sending the wrong requests to the server).

### Reproduce
Open a file that works with ghcide, and type `f = head`; move cursor over `head`, and attempt to go to definition. On client side this sends to the server:

```
(:jsonrpc "2.0" :id 16 :method "textDocument/definition" :params
	  (:textDocument
	   (:uri "file:///home/adam/haskell/bug-test/Test.hs")
	   :position
	   (:line 0 :character 6)))
```
(apologies for the elisp rendering rather than raw json - I am grabbing this from eglot)

### Expected behavior
Expected behavior: cursor stays over `head`. On ghcide side, return a null result e.g.:
```
(:result
 []
 :jsonrpc "2.0" :id 16)
```

### Actual behavior
Actual behavior: cursor moves to outer definition (`f`). On ghcide side:
```
(:result
 (:uri "file:///home/adam/haskell/bug-test/Test.hs" :range
       (:start
	(:line 0 :character 0)
	:end
	(:line 0 :character 1)))
 :jsonrpc "2.0" :id 16)
```
### ghcide version
```
ghcide version: 0.0.5 (GHC: 8.6.5) (PATH: /home/adam/.local/bin/ghcide) (GIT hash: a0aa013e3362d6ae0a963639d7845ec9990350fe)
```

cocreature:
Currently, we use the source span of the match which corresponds to
the whole clause instead of just the function identifier. This
resulted in us pointing every goto definition request within a clause
to the function if there is no other information (either because it
failed because it came from an external package or simply because you
are not on an identifier).

This PR fixes this by getting the proper source spans frmo the
HsMatchContext. Somewhat annoyingly, we have to get it from the parsed
module since GHC messes this up during typechecking but it’s
reasonably simple.

fixes #315

### DIFF
--- a/src/Development/IDE/Spans/Calculate.hs
+++ b/src/Development/IDE/Spans/Calculate.hs
@@ -23,6 +23,7 @@ import           Desugar
 import           GHC
 import           GhcMonad
 import           FastString (mkFastString)
+import           OccName
 import           Development.IDE.Types.Location
 import           Development.IDE.Spans.Type
 import           Development.IDE.GHC.Error (zeroSpan)
@@ -63,7 +64,8 @@ getSpanInfo mods tcm =
          es  = listifyAllSpans  tcs :: [LHsExpr GhcTc]
          ps  = listifyAllSpans' tcs :: [Pat GhcTc]
          ts  = listifyAllSpans $ tm_renamed_source tcm :: [LHsType GhcRn]
-     bts <- mapM (getTypeLHsBind tcm) bs -- binds
+     let funBinds = funBindMap $ tm_parsed_module tcm
+     bts <- mapM (getTypeLHsBind funBinds) bs -- binds
      ets <- mapM (getTypeLHsExpr tcm) es -- expressions
      pts <- mapM (getTypeLPat tcm)    ps -- patterns
      tts <- mapM (getLHsType tcm)     ts -- types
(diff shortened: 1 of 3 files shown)


src/Development/IDE/Spans/Calculate.hs
```diff
@@ -76,6 +79,14 @@ getSpanInfo mods tcm =
           | b `isSubspanOf` a = GT
           | otherwise         = compare (srcSpanStart a) (srcSpanStart b)
 
+-- | The locations in the typechecked module are as precise as the ones in the parsed module.
+-- Therefore we build up a map from OccName to the corresponding definition in the parsed module
+-- to lookup precise locations for things like multi-clause function definitions.
```
aherrmann-da: This comment confuses me. I thought the issue was that the locations are messed up in the typechecked module and that's why you're using the parsed module instead.

cocreature: oops, my wording was broken there. Should be fixed now.

aherrmann-da: :+1: Thanks!

PR #322, opened 2020-01-12
Fix performance of getFileExists

pepeiborra:
This is a minimal solution that does not entirely follow the plan in #101, focusing on GetFileExists instead of GetLocatedImports. 

In my experiments it solves the problem, hopefully without breaking anything else. Checked with VSCode only. 

Adding tests is tricky since lsp-test does not send DidChangeWatchedFiles notifications. I'm open to suggestions on what to do.

Leveraging DidChangeWatchedFiles notifications for other purposes, e.g. in the GetModificationTime rule, is left for future PRs.

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -121,6 +121,7 @@ library
         Development.IDE.Core.Debouncer
         Development.IDE.Core.Compile
         Development.IDE.Core.Preprocessor
+        Development.IDE.Core.FileExists
         Development.IDE.GHC.Compat
         Development.IDE.GHC.CPP
         Development.IDE.GHC.Error
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -222,4 +222,16 @@ test-suite ghcide-tests
         Development.IDE.Test
         Development.IDE.Test.Runfiles
     default-extensions:
+        BangPatterns
+        DeriveFunctor
+        DeriveGeneric
+        GeneralizedNewtypeDeriving
+        LambdaCase
+        NamedFieldPuns
         OverloadedStrings
+        RecordWildCards
+        ScopedTypeVariables
+        StandaloneDeriving
+        TupleSections
+        TypeApplications
+        ViewPatterns
--- a/src/Development/IDE/Core/FileExists.hs
+++ b/src/Development/IDE/Core/FileExists.hs
@@ -10,13 +10,15 @@ where
 
 import           Control.Concurrent.Extra
 import           Control.Exception
-import           Control.Monad
+import           Control.Monad.Extra
 import qualified Data.Aeson                    as A
 import           Data.Binary
 import qualified Data.ByteString.Lazy          as BS
 import           Data.Map.Strict                ( Map )
 import qualified Data.Map.Strict               as Map
+import           Data.Maybe
 import qualified Data.Text                     as T
+import           Development.IDE.Core.FileStore
 import           Development.IDE.Core.Shake
 import           Development.IDE.Types.Location
 import           Development.Shake
@@ -25,6 +27,7 @@ import           GHC.Generics
 import           Language.Haskell.LSP.Messages
 import           Language.Haskell.LSP.Types
 import           Language.Haskell.LSP.Types.Capabilities
+import qualified System.Directory as Dir
 
 -- | A map for tracking the file existence
 type FileExistsMap = (Map NormalizedFilePath Bool)
@@ -75,7 +78,7 @@ getFileExists fp = use_ GetFileExists fp
 -- | Installs the 'getFileExists' rules.
 --   Provides a fast implementation if client supports dynamic watched files.
 --   Creates a global state as a side effect in that case.
-fileExistsRules :: IO LspId -> ClientCapabilities -> (NormalizedFilePath -> Action Bool) -> Rules ()
+fileExistsRules :: IO LspId -> ClientCapabilities -> VFSHandle -> Rules ()
 fileExistsRules getLspId ClientCapabilities{_workspace}
   | Just WorkspaceClientCapabilities{_didChangeWatchedFiles} <- _workspace
   , Just DidChangeWatchedFilesClientCapabilities{_dynamicRegistration} <- _didChangeWatchedFiles
@@ -84,8 +87,8 @@ fileExistsRules getLspId ClientCapabilities{_workspace}
   | otherwise = fileExistsRulesSlow
 
 --   Requires an lsp client that provides WatchedFiles notifications.
-fileExistsRulesFast :: IO LspId -> (NormalizedFilePath -> Action Bool) -> Rules ()
-fileExistsRulesFast getLspId getFileExists = do
+fileExistsRulesFast :: IO LspId -> VFSHandle -> Rules ()
+fileExistsRulesFast getLspId vfs = do
   addIdeGlobal . FileExistsMapVar =<< liftIO (newVar [])
   defineEarlyCutoff $ \GetFileExists file -> do
     fileExistsMap <- getFileExistsMapUntracked
@@ -93,7 +96,7 @@ fileExistsRulesFast getLspId getFileExists = do
     case mbFilesWatched of
       Just fv -> pure (createKey fv, ([], Just fv))
       Nothing -> do
-        exist                   <- getFileExists file
+        exist                   <- liftIO $ getFileExistsVFS vfs file
         ShakeExtras { eventer } <- getShakeExtras
 
         -- add a listener for VFS Create/Delete file events,
@@ -123,15 +126,23 @@ fileExistsRulesFast getLspId getFileExists = do
 
     void $ eventer $ ReqRegisterCapability req
 
-fileExistsRulesSlow:: (NormalizedFilePath -> Action Bool) -> Rules ()
-fileExistsRulesSlow getFileExists = do
+fileExistsRulesSlow:: VFSHandle -> Rules ()
+fileExistsRulesSlow vfs = do
   defineEarlyCutoff $ \GetFileExists file -> do
     alwaysRerun
-    exist                   <- getFileExists file
+    exist <- liftIO $ getFileExistsVFS vfs file
     pure (createKey exist, ([], Just exist))
  where
   createKey = Just . BS.toStrict . encode
 
+getFileExistsVFS :: VFSHandle -> NormalizedFilePath -> IO Bool
+getFileExistsVFS vfs file = do
+    -- we deliberately and intentionally wrap the file as an FilePath WITHOUT mkAbsolute
+    -- so that if the file doesn't exist, is on a shared drive that is unmounted etc we get a properly
+    -- cached 'No' rather than an exception in the wrong place
+    handle (\(_ :: IOException) -> return False) $
+        (isJust <$> getVirtualFile vfs (filePathToUri' file)) ||^
+        Dir.doesFileExist (fromNormalizedFilePath file)
 
 --------------------------------------------------------------------------------------------------
 -- The message definitions below probably belong in haskell-lsp-types
--- a/src/Development/IDE/Core/FileExists.hs
+++ b/src/Development/IDE/Core/FileExists.hs
@@ -60,7 +60,7 @@ modifyFileExists state changes = do
     modifyVar_ var $ evaluate . Map.union changesMap
     let flushPreviousValues = do
           ShakeExtras { state } <- getShakeExtras
-          liftIO $ mapM_ (resetValue state GetFileExists . fst) changes
+          liftIO $ mapM_ (deleteValue state GetFileExists . fst) changes
     void $ shakeRun state [flushPreviousValues]
 
 -------------------------------------------------------------------------------------
--- a/src/Development/IDE/Core/FileExists.hs
+++ b/src/Development/IDE/Core/FileExists.hs
@@ -13,7 +13,7 @@ import           Control.Exception
 import           Control.Monad.Extra
 import qualified Data.Aeson                    as A
 import           Data.Binary
-import qualified Data.ByteString.Lazy          as BS
+import qualified Data.ByteString               as BS
 import           Data.Map.Strict                ( Map )
 import qualified Data.Map.Strict               as Map
 import           Data.Maybe
@@ -104,7 +104,7 @@ fileExistsRulesFast getLspId vfs = do
     fileExistsMap <- getFileExistsMapUntracked
     let mbFilesWatched = Map.lookup file fileExistsMap
     case mbFilesWatched of
-      Just fv -> pure (createKey fv, ([], Just fv))
+      Just fv -> pure (summarizeExists fv, ([], Just fv))
       Nothing -> do
         exist                   <- liftIO $ getFileExistsVFS vfs file
         ShakeExtras { eventer } <- getShakeExtras
@@ -122,9 +122,8 @@ fileExistsRulesFast getLspId vfs = do
               -- if the key was already there, do nothing
               return x
 
-        pure (createKey exist, ([], Just exist))
+        pure (summarizeExists exist, ([], Just exist))
  where
-  createKey = Just . BS.toStrict . encode
   addListener eventer fp = do
     reqId <- getLspId
     let
@@ -142,12 +141,15 @@ fileExistsRulesFast getLspId vfs = do
 
     eventer $ ReqRegisterCapability req
 
+summarizeExists :: Bool -> Maybe BS.ByteString
+summarizeExists x = Just $ if x then BS.singleton 1 else BS.empty
+
 fileExistsRulesSlow:: VFSHandle -> Rules ()
 fileExistsRulesSlow vfs = do
   defineEarlyCutoff $ \GetFileExists file -> do
     alwaysRerun
     exist <- liftIO $ getFileExistsVFS vfs file
-    pure (Just $ BS.toStrict $ encode exist, ([], Just exist))
+    pure (summarizeExists exist, ([], Just exist))
 
 getFileExistsVFS :: VFSHandle -> NormalizedFilePath -> IO Bool
 getFileExistsVFS vfs file = do
--- a/src/Development/IDE/Core/FileExists.hs
+++ b/src/Development/IDE/Core/FileExists.hs
@@ -54,11 +54,14 @@ modifyFileExists :: IdeState -> [(NormalizedFilePath, Bool)] -> IO ()
 modifyFileExists state changes = do
   FileExistsMapVar var <- getIdeGlobalState state
   changesMap           <- evaluate $ Map.fromList changes
-  modifyVar_ var $ evaluate . Map.union changesMap
-  let flushPreviousValues = do
-        ShakeExtras { state } <- getShakeExtras
-        liftIO $ mapM_ (resetValue state GetFileExists . fst) changes
-  void $ shakeRun state [flushPreviousValues]
+
+  -- Mask to ensure that the previous values are flushed together with the map update
+  mask $ \_ -> do
+    modifyVar_ var $ evaluate . Map.union changesMap
+    let flushPreviousValues = do
+          ShakeExtras { state } <- getShakeExtras
+          liftIO $ mapM_ (resetValue state GetFileExists . fst) changes
+    void $ shakeRun state [flushPreviousValues]
 
 -------------------------------------------------------------------------------------
 
--- a/src/Development/IDE/Core/FileExists.hs
+++ b/src/Development/IDE/Core/FileExists.hs
@@ -75,17 +75,17 @@ getFileExists fp = use_ GetFileExists fp
 -- | Installs the 'getFileExists' rules.
 --   Provides a fast implementation if client supports dynamic watched files.
 --   Creates a global state as a side effect in that case.
-fileExistsRules :: ClientCapabilities -> (NormalizedFilePath -> Action Bool) -> Rules ()
-fileExistsRules ClientCapabilities{_workspace}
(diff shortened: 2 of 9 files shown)


src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,161 @@
...
+          Nothing -> do
+            addListener eventer file
+            return x
+        pure (createKey exist, ([], Just exist))
+ where
+  createKey = Just . BS.toStrict . encode
+  addListener eventer fp = do
+    let
+      req = RequestMessage "2.0" reqId ClientRegisterCapability regParams
+      reqId        = IdString fpAsId
```
pepeiborra: Using the filepath as an id is perhaps not the best thing to do, but I could not figure out how to get access to the `IO LspId` command. Maybe it could be added to `ShakeExtras` ?

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,161 @@
...
+
+-- | Returns True if the file exists
+getFileExists :: NormalizedFilePath -> Action Bool
+getFileExists fp = use_ GetFileExists fp
+
+-- | Installs the 'getFileExists' rules.
+--   Provides a fast implementation if client supports dynamic watched files.
+--   Creates a global state as a side effect in that case.
+fileExistsRules :: ClientCapabilities -> (NormalizedFilePath -> Action Bool) -> Rules ()
+fileExistsRules ClientCapabilities{_workspace}
```
ndmitchell: I find it weird to have _ variables that are then used.

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,161 @@
...
+    case mbFilesWatched of
+      Just fv -> pure (createKey fv, ([], Just fv))
+      Nothing -> do
+        exist                   <- getFileExists file
+        ShakeExtras { eventer } <- getShakeExtras
+
+        -- add a listener for VFS Create/Delete file events,
+        -- taking the FileExistsMap lock to prevent race conditions
+        -- that would lead to multiple listeners for the same path
+        modifyFileExistsAction $ \x -> case Map.lookup file x of
```
ndmitchell: I'd have written the logic for this as `\x -> when (not $ Map.exists file x) $ addListener; return x`. Makes it clear what and why we are looking up, and that we always return x.

ndmitchell:
Nice work. To attempt to summarise the code (so I can check I am reading it right), if the IDE supports dynamic registration of files to watch, then any time we look at a file, we register a listener for it. Seems sensible and smart, and if it fixes the problem, a nice and targeted fix.

We should definitely fix lsp-test to follow the LSP spec here (maybe not immediately, but one day). It's exactly the kind of thing that is good to test!

aherrmann-da:
Thank you! I did a quick test run in vscode on the ghcide repository and this does greatly speed-up hover for me. On ghcide master hover takes between 0.50-0.90s to complete, and with this PR between 0.04-0.07s. In my case the reason for the slow hover on master are many file lookups in the `.stack-work` directory tree.

ghcide.cabal
```diff
@@ -221,4 +222,16 @@ test-suite ghcide-tests
         Development.IDE.Test
         Development.IDE.Test.Runfiles
     default-extensions:
+        BangPatterns
```
cocreature: This change looks unrelated to this PR and you barely make use of it (judging from the fact that you didn’t remove any extensions and you only added one module so you could just add all extensions there.

src/Development/IDE/Core/Shake.hs
```diff
@@ -257,6 +259,16 @@ setValues state key file val = modifyVar_ state $ \vals -> do
     -- Force to make sure the old HashMap is not retained
     evaluate $ HMap.insert (file, Key key) (fmap toDyn val) vals
 
+-- | Delete the value stored for a given ide build key
+resetValue
```
cocreature:
```suggestion
deleteValue
```

test/exe/Main.hs
```diff
@@ -1532,7 +1533,12 @@ run s = withTempDir $ \dir -> do
   -- HIE calls getXgdDirectory which assumes that HOME is set.
   -- Only sets HOME if it wasn't already set.
   setEnv "HOME" "/homeless-shelter" False
-  runSessionWithConfig conf cmd fullCaps { _window = Just $ WindowClientCapabilities $ Just True } dir s
+  let lspTestCaps =
```
cocreature: `lsp-test` has a generic `sendNotification` function that we should be able to use to send `didChangeWatchedFiles` events.

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,161 @@
...
+modifyFileExistsAction f = do
+  FileExistsMapVar var <- getIdeGlobalAction
+  liftIO $ modifyVar_ var f
+
+-- | Modify the global store of file exists
+modifyFileExists :: IdeState -> [(NormalizedFilePath, Bool)] -> IO ()
+modifyFileExists state changes = do
+  FileExistsMapVar var <- getIdeGlobalState state
+  changesMap           <- evaluate $ Map.fromList changes
+  modifyVar_ var $ evaluate . Map.union changesMap
```
cocreature: This looks like it might need to be masked. Otherwise, you can end up modifying `changesMap` without deleting the values if you get interrupted in between.

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,161 @@
...
+  , Just True <- _dynamicRegistration
+  = fileExistsRulesFast
+  | otherwise = fileExistsRulesSlow
+
+--   Requires an lsp client that provides WatchedFiles notifications.
+fileExistsRulesFast :: (NormalizedFilePath -> Action Bool) -> Rules ()
+fileExistsRulesFast getFileExists = do
+  addIdeGlobal . FileExistsMapVar =<< liftIO (newVar [])
+  defineEarlyCutoff $ \GetFileExists file -> do
+    fileExistsMap <- getFileExistsMapUntracked
```
cocreature: It looks like we could avoid the need for deleting values if we set this to `alwaysRerun`. What’s the performance impact of that? As mentioned above, I believe the current code is not quite right in the presence of async exceptions and more generally this feels a bit ugly so I wonder if we can get away with something simpler.

src/Development/IDE/Core/FileExists.hs
cocreature: :+1:

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,161 @@
...
+                                  , kind        = Just 5 -- Create and Delete events only
+                                  }
+
+    void $ eventer $ ReqRegisterCapability req
+
+fileExistsRulesSlow:: (NormalizedFilePath -> Action Bool) -> Rules ()
+fileExistsRulesSlow getFileExists = do
+  defineEarlyCutoff $ \GetFileExists file -> do
+    alwaysRerun
+    exist                   <- getFileExists file
```
cocreature: I’m confused by this. You are defining `GetFileExists` by calling `getFileExists` which is defined as `use_ GetFileExits` in this module https://github.com/digital-asset/ghcide/pull/322/files#diff-61e4beab82bf9b7df988b9e69ab45ae4R72. Am I missing something here? It looks like this should be calling `getFileExists` from `Development.IDE.Core.FileStore` which should probably move to this module.

src/Development/IDE/Core/FileStore.hs
```diff
@@ -198,20 +189,20 @@ ideTryIOException fp act =
 getFileContents :: NormalizedFilePath -> Action (FileVersion, Maybe StringBuffer)
 getFileContents = use_ GetFileContents
 
-getFileExists :: NormalizedFilePath -> Action Bool
-getFileExists =
+getFileExists :: VFSHandle -> NormalizedFilePath -> IO Bool
```
cocreature: I’m not sure if/where this is still used. It looks like this should probably become internal to the implementation of the `getFileExists` rules.

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,161 @@
...
+    pure (createKey exist, ([], Just exist))
+ where
+  createKey = Just . BS.toStrict . encode
```
cocreature: Key is a bit confusing. This is the value (or a hash of the value) not the key.

src/Development/IDE/Core/FileExists.hs
pepeiborra: This is a record pun on a field name. I don't think that `_workspace=workspace` is going to be any better

ghcide.cabal
pepeiborra: I made it because I wasted a couple of minutes rewriting my test code to work around hlint errors. 
```
  # Extensions enabled by `bazel` and `da-ghci` by default.  We ban them here
  # to avoid useless pragmas piling up on the top of files.
  - {name: BangPatterns, within: []}
  - {name: DeriveDataTypeable, within: []}
  - {name: DeriveFoldable, within: []}
  - {name: DeriveFunctor, within: []}
  - {name: DeriveGeneric, within: []}
  - {name: DeriveTraversable, within: []}
  - {name: FlexibleContexts, within: []}
  - {name: GeneralizedNewtypeDeriving, within: []}
  - {name: LambdaCase, within: []}
  - {name: NamedFieldPuns, within: []}
  - {name: OverloadedStrings, within: []}
  - {name: PackageImports, within: []}
  - {name: RecordWildCards, within: []}
  - {name: ScopedTypeVariables, within: []}
  - {name: StandaloneDeriving, within: []}
  - {name: TupleSections, within: []}
  - {name: TypeApplications, within: []}
  - {name: ViewPatterns, within: []}
```
It's unrelated to this PR indeed, but also too small to justify a PR on its own.

src/Development/IDE/Core/FileExists.hs
pepeiborra: Yes, you can do that and it's simpler, but the performance savings are not that good.

Part of the problem here is that GetLocatedModules depends on I*L*2 file paths, where I is the number of Imports and L is the number of incLude paths. A very large number of `getFileExists` calls, Shake db lookups, binary serialization and hashing, and finally a second lookup in the `GetFileExistsMap`. I'm seeing 1s for every 2000 calls to getFileExists, so in a project like GHC where there's 5891 calls to `getFileExists` per hover that's 3s, and in a codebase like ours at SCB where there's 20.000 that's 10s, I think.

cocreature: Are those numbers with the changes in this PR + the alwaysRerun change or without any changes from this PR? There is no doubt that there is a higher overhead to the solution I proposed but I would expect that it’s still a measurable improvement over the status quo so I’d like to see some numbers on how big the difference actually is before we commit to something.

pepeiborra: Those are the numbers of this PR + the alwaysRerun change. I would attach a profile but I have already deleted it

src/Development/IDE/Core/FileExists.hs
pepeiborra: That's right, thanks!

pepeiborra: Simpler now

src/Development/IDE/Core/FileExists.hs
pepeiborra: Thanks, I didn't think of that. In order to delete the values from the state though, I need to call `runShake`. Is it safe to call `runShake` masked?

test/exe/Main.hs
pepeiborra: Thanks, I didn't realize that. 

After enabling the WatchedFiles capability in the test client, an existing test for adding a missing module no longer passed, and with the use of `sendNotification` to send a `WatchedFile` file creation event, now does. 

From my perspective the new feature is now sufficiently tested, but I suppose if ghcide really wanted to support clients without WatchedFiles capability, the entire test suite should be run twice with and without the capability, or at least have tests for both cases.

src/Development/IDE/Core/FileExists.hs
cocreature: `runShake` has a nontrivial overhead so given that we only want to modify the var here, maybe it would make sense to refactor this such that we don’t need `runShake` here at all, e.g., by allocating the Var in `initialise` and then passing it on.

pepeiborra:
Is that something for this PR?

src/Development/IDE/Core/FileExists.hs
pepeiborra: Does the overhead matter if we only subscribe to create and delete events?

Is that refactoring small enough to include in this PR?

test/exe/Main.hs
```diff
@@ -230,7 +230,8 @@ diagnosticTests = testGroup "diagnostics"
       _ <- openDoc' "ModuleB.hs" "haskell" contentB
       expectDiagnostics [("ModuleB.hs", [(DsError, (1, 7), "Could not find module")])]
       let contentA = T.unlines [ "module ModuleA where" ]
-      _ <- openDoc' "ModuleA.hs" "haskell" contentA
+      TextDocumentIdentifier newModule <- openDoc' "ModuleA.hs" "haskell" contentA
```
cocreature: This looks like it is no longer sufficient for the client to just send `openDoc'` if it supports `WorkspaceDidChangeWatchedFiles`. I don’t think that is correct. `WorkspaceDidChangeWatchedFiles` is for files on disk that aren’t opened. So the test for this should only write a file to disk and send the event, not call `openDoc'`. On the other hand, `openDoc'` should still be sufficient for this error to go away even without the event since the file might not be written to disk.

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,172 @@
...
+  defineEarlyCutoff $ \GetFileExists file -> do
+    fileExistsMap <- getFileExistsMapUntracked
+    let mbFilesWatched = Map.lookup file fileExistsMap
+    case mbFilesWatched of
+      Just fv -> pure (createKey fv, ([], Just fv))
+      Nothing -> do
+        exist                   <- liftIO $ getFileExistsVFS vfs file
+        ShakeExtras { eventer } <- getShakeExtras
+
+        -- add a listener for VFS Create/Delete file events,
```
cocreature: I don’t think you are preventing any race conditions here. `addListener` adds the listener asynchronously by going through the `eventer` which will basically just do a `writeChan`.

Also if the request to adds the listener fails for whatever reason, you will never notice that. The proper fix for that would probably be to track the request and wait to see if you get a successful response. If it fails, you have to rerun and at least in theory you also have to rerun once after the listener is registered to prevent events between the initial query and the listener being registered from getting lost.

I would suggest to not try to attack this in this PR given that I don’t see why this request should ever fail (ignoring bugs like invalid messages) and the race seems fairly harmless but we should update the comment.

src/Development/IDE/Core/FileExists.hs
cocreature: I think it makes sense to include it in this PR mainly because I’m not entirely comfortable with masking `shakeRun`. As far as I can tell, the only reason you need to do that is to get access to `ShakeExtras` which is already in `IdeState` so if you simply change `deleteValue` to take `IdeState` as the input, you should be able to avoid this.

test/exe/Main.hs
pepeiborra: I agree that `openDoc'` is not needed for the test, but it's still handy to construct the uri. 

I'm not sure what your second statement means. Can you elaborate? The file does not exist on disk initially, so `GetLocatedModules` will not consider after creation it until the event has been received

src/Development/IDE/Core/FileExists.hs
pepeiborra: Even if `eventer` is asyncrhonous, the race is prevented by the key inserted in the `fileExistsMap`. But if the request fails we'll never notice, agreed.

cocreature: the key is inserted once you receive an event, not once the listener is registered. So I don’t think it prevents you from registering the same listener multiple times.

test/exe/Main.hs
cocreature: Given that `openDoc'` affects the VFS which in turn affects the file exists check, I really think we should not use this here or ideally have two tests. One using `openDoc'` and one without.

As for the second question, without your PR a file is considered existent if it is in the VFS and only if it isn’t in the VFS we look it up on disk. So `openDoc` which adds it to the VFS should be sufficient for it to be picked up. With your change that seems to no longer hold.

src/Development/IDE/Core/FileExists.hs
pepeiborra: The code was wrong from the very first version. It was not inserting anything at all in the map. This would result in unnecessary calls to `getFileExistsVFS`. I have fixed it to insert in the map and now it will only ever set up the listener once, iiuc. Recovering from a failed listener install seems more tricky:
- if it failed with an exception, then no insertion would ever happen and we'd be good. But this is not the case
- as it fails asynchronously, we would need a continuation to remove the entry from `getFileExistsMap`. But as far as I can see there's no way to install such continuation.

test/exe/Main.hs
pepeiborra: Iiuc the problem is that `openDoc'` is badly named. It not only "opens" the file, but it also creates it. Therefore it really should be sending watched file creation notifications - if it was, then it would be sufficient. 

Let me know if that makes sense and I can add it to the PR.

cocreature: Sorry I think I’m not making myself clear. Let me try to give a concrete example.

1. You have module `A` on disk and opened in your LSP client (let’s say Emacs since VSCode behaves in a specific way that makes this a non-issue afaik but that’s definitely not mandated by the LSP spec). Module `A` imports module `B` which is neither opened nor does it exist on dis.
2. Now you open a new buffer for module `B`. This will send a `DidOpenTextDocument` notification to the server (i.e. ghcide). This will result in the module being added to the VFS. However, the module does not necessarily exist on disk at this point. (This is where it matters that we use Emacs and not VSCode since opening a buffer for a non-existent file does not write it to disk until you save whereas I don’t know how to open a buffer for a non-existent file in VSCode since the new file function seems to create it on disk).
3. The client will not send a `DidChangeWatchedFilesParams` notification since the file does not yet exist on disk.
4. Without your PR we consider a file existent if it exists in the VFS so at this point `B` exists and the import in `A` can be resolved.
5. With your PR you will not consider `B` existent as far as I can tell.

pepeiborra: Thanks for the detailed explanation, I see what you mean now. Indeed, until B is saved to disk, it will not be considered as existing; this is a behaviour change, albeit only noticeable in some editors. But then again, the user only needs to save the file to make things work.

The issue is subtle - care needs to be taken in order to ensure that every new buffer is covered by a `DidChangeWatchedFile` subscription, otherwise we will miss renames/deletions. 

Since the behaviour change is minimal and the solution is tricky, I would vote for waiting until we have a user opened issue before trying to fix.

cocreature: Sounds reasonable. Let’s please add a comment (probably to the `GetFileExists` rule) that mentions this. Maybe just copy the whole example that I posted above.

pepeiborra: Done!

mpickering:
Thank you @pepeiborra. Once this is released I can start to advertise using ghcide on GHC's code base more.

pepeiborra:
@cocreature is there anything else needed here?

cocreature:
Should be good to go I think, I’ll do a final review tomorrow when I’m back at work.

src/Development/IDE/Core/FileExists.hs
```diff
@@ -0,0 +1,187 @@
...
+modifyFileExists state changes = do
+  FileExistsMapVar var <- getIdeGlobalState state
+  changesMap           <- evaluate $ Map.fromList changes
+
+  -- Mask to ensure that the previous values are flushed together with the map update
+  mask $ \_ -> do
+    modifyVar_ var $ evaluate . Map.union changesMap
+    let flushPreviousValues = do
+          liftIO $ mapM_ (deleteValue state GetFileExists . fst) changes
+    void $ shakeRun state [flushPreviousValues]
```
cocreature:
```suggestion
    flushPreviousValues
```
You don’t need to call `shakeRun` here. You can just call `flushPreviousValues` which avoids potential issues from masking `shakeRun` (not entirely sure if that causes dissues but I’d rather avoid it if we can).

src/Development/IDE/LSP/Notifications.hs
```diff
@@ -52,4 +56,16 @@ setHandlersNotifications = PartialHandlers $ \WithMessage{..} x -> return x
...
+            let events =
+                    mapMaybe
+                        (\(FileEvent uri ev) ->
+                            (, ev /= FcDeleted) . toNormalizedFilePath
+                            <$> LSP.uriToFilePath uri
+                        )
+                        ( F.toList fileEvents )
+            let msg = Text.pack $ show events
+            logInfo (ideLogger ide) $ "Files created or deleted: " <> msg
+            modifyFileExists ide events
```
cocreature:
```suggestion
            modifyFileExists ide events
            setSomethingModified ide
```
If we follow my suggestion above to remove the call to `shakeRun` we need to call it somewhere to kick things off.

src/Development/IDE/Core/FileExists.hs
pepeiborra: I thought I had done this. Thanks!

cocreature:
Awesome, thank you so much for working on this! Sorry for the long review process.

PR #329, opened 2020-01-17
GC tweaks

pepeiborra:
ghcide uses a lot of memory in large codebases. In ours it starts at 7GB and goes up to 30GB quite easily. Needless to say, GC settings have a huge impact. Disabling the idle GC alone cuts up to 4s for hovers.

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -151,7 +151,18 @@ executable ghcide
       buildable: False
     default-language:   Haskell2010
     hs-source-dirs:     exe
-    ghc-options: -threaded -Wall -Wno-name-shadowing
+    ghc-options:
+                -threaded
+                -Wall
+                -Wno-name-shadowing
+                -- allow user RTS overrides
+                -rtsopts
+                -- disable idle GC
+                -with-rtsopts=-I0
+                -- disable parallel GC
+                -with-rtsopts=-qg
+                -- increase nursery size
+                -with-rtsopts=-A128M
     main-is: Main.hs
     build-depends:
         hslogger,


ndmitchell:
I've used similar GC tweaks on Shake, and find they are usually super helpful.

Can you give any stats about your code base? I'm looking to use ghcide on a pretty huge codebase, and curious to know if it's going to work. I'm wondering if some kind of hi file, paging stuff out, is going to be required at larger sizes?

pepeiborra:
We have ~1300 Haskell modules and 300 packages in the ghc pkg dB. Using ghc 8.6.5 so haven’t tried hie files yet, ghcide takes several minutes to boot

Some of us coming to the Bristol Hackathon are planning to work on reducing the memory usage by increasing sharing.

mpickering:
@pepeiborra Have you done any space profiling to work out what the big memory hogs are?

pepeiborra:
@mpickering I have collected some space profiles in our work repo which show that a large portion of the heap is used by GHC data structures, including Types and Spans. 

Instead of sharing those though, it might be better to collect space profiles done on GHC itself. I've been trying to do that but given up on loading GHC in ghcide. Do you have any insights? See https://gitlab.haskell.org/pepeiborra/ghc/commit/4f4b83d261691fd6872218c9d68976cafb6b3267

mpickering:
@pepeiborra Why do your changes to the files not work?

pepeiborra:
I got ghcide to load GHC successfully after deleting the `.hie-bios` folder. But now ghcide rebuilt with the profiling runtime is unhappy for some other reason.

mpickering:
You probably need to also enable profiling for the project you are trying to profile which would require modifying the hadrian build flavour to build profiling versions of the libraries.

PR #335, opened 2020-01-20
Fix the build for ghc-8.8.2.

### LINKED ISSUE #334: Failed build with ghc 8.8.2
Building with ghc 8.8.2 fails with a couple of errors related to out of scope data constructors, presumably changed in the latest compiler:

```
src/Development/IDE/GHC/CPP.hs:143:22: error:
    Not in scope: data constructor ‘LlvmVersion’
    |
143 |                Just (LlvmVersion n) -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (n,0) ]
    |                      ^^^^^^^^^^^

src/Development/IDE/GHC/CPP.hs:144:22: error:
    Not in scope: data constructor ‘LlvmVersionOld’
    |
144 |                Just (LlvmVersionOld m n) -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m,n) ]
    |                      ^^^^^^^^^^^^^^
```

On 703bb82a1d3b690c7ef2b8cb7d28c0ac5171dd0c, `HEAD` at time of writing.

robrix:
This PR fixes the build for ghc 8.8.2 by using the new `LlvmVersion` API (which no longer exports constructors).

Fixes #334.

### DIFF
--- a/src/Development/IDE/GHC/CPP.hs
+++ b/src/Development/IDE/GHC/CPP.hs
@@ -28,7 +28,9 @@ import Module
 import DynFlags
 import Panic
 import FileCleanup
-#if MIN_GHC_API_VERSION(8,8,0)
+#if MIN_GHC_API_VERSION(8,8,2)
+import LlvmCodeGen (LlvmVersion, llvmVersionList)
+#elif MIN_GHC_API_VERSION(8,8,0)
 import LlvmCodeGen (LlvmVersion (..))
 #endif
 
@@ -139,7 +141,11 @@ getBackendDefs :: DynFlags -> IO [String]
 getBackendDefs dflags | hscTarget dflags == HscLlvm = do
     llvmVer <- figureLlvmVersion dflags
     return $ case llvmVer of
-#if MIN_GHC_API_VERSION(8,8,0)
+#if MIN_GHC_API_VERSION(8,8,2)
+               Just v
+                 | m:n:_ <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m, n) ]
+                 | n:_   <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (n, 0) ]
+#elif MIN_GHC_API_VERSION(8,8,0)
                Just (LlvmVersion n) -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (n,0) ]
                Just (LlvmVersionOld m n) -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m,n) ]
 #else
--- a/src/Development/IDE/GHC/CPP.hs
+++ b/src/Development/IDE/GHC/CPP.hs
@@ -143,8 +143,8 @@ getBackendDefs dflags | hscTarget dflags == HscLlvm = do
     return $ case llvmVer of
 #if MIN_GHC_API_VERSION(8,8,2)
                Just v
-                 | m:n:_ <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m, n) ]
-                 | n:_   <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (n, 0) ]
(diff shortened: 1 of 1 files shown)


src/Development/IDE/GHC/CPP.hs
```diff
@@ -139,7 +141,11 @@ getBackendDefs :: DynFlags -> IO [String]
 getBackendDefs dflags | hscTarget dflags == HscLlvm = do
     llvmVer <- figureLlvmVersion dflags
     return $ case llvmVer of
-#if MIN_GHC_API_VERSION(8,8,0)
+#if MIN_GHC_API_VERSION(8,8,2)
+               Just v
+                 | m:n:_ <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m, n) ]
+                 | n:_   <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (n, 0) ]
```
robrix: I don’t have any projects which use LLVM, so I’d appreciate it if someone could double-check that this is doing what they expect.

Note that despite `llvmVersionList` returning `[Int]`, these branches are in fact exhaustive (under `ghc` 8.8.2), because `LlvmVersion` contains a `NonEmpty` list and converts it to a list otherwise unmodified. So we will always have at least a major version, and thus this will not silently fail (modulo future changes to `ghc`’s API).

src/Development/IDE/GHC/CPP.hs
```diff
@@ -139,7 +141,11 @@ getBackendDefs :: DynFlags -> IO [String]
 getBackendDefs dflags | hscTarget dflags == HscLlvm = do
     llvmVer <- figureLlvmVersion dflags
     return $ case llvmVer of
-#if MIN_GHC_API_VERSION(8,8,0)
+#if MIN_GHC_API_VERSION(8,8,2)
+               Just v
+                 | m:n:_ <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m, n) ]
```
cocreature:
```suggestion
                 | [m] <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m, 0) ]
```

src/Development/IDE/GHC/CPP.hs
cocreature:
```suggestion
                 | m:n:_   <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (m, n) ]
```

src/Development/IDE/GHC/CPP.hs
```diff
@@ -139,7 +141,11 @@ getBackendDefs :: DynFlags -> IO [String]
...
+                 | n:_   <- llvmVersionList v -> [ "-D__GLASGOW_HASKELL_LLVM__=" ++ format (n, 0) ]
+#elif MIN_GHC_API_VERSION(8,8,0)
```
cocreature: This matches the way this is written in GHC and is a bit easier to read imho: https://gitlab.haskell.org/ghc/ghc/blob/ghc-8.8/compiler/main/DriverPipeline.hs#L2174

robrix: Aha, good call 👍

robrix: Shame they didn’t export that.

ndmitchell: The whole doCpp is exported from GHC 8.10 and above, so the whole thing can disappear.

robrix: @ndmitchell Do you mean that we should do that in this PR? Cos I’m not really set up to test against 8.10.* just now.

ndmitchell: No - we should leave it as it is for now. Once 8.10 is actually out we should do some conditional stuff - more just pointing out there is a long term plan for getting rid of it all.

cocreature: No, we need to support older versions so this can’t go away completely in particular not the code for `8.8.2`. We can however, use the definition from GHC in `8.10` but let’s please keep `8.10` support for a separate PR. I expect that there are a bunch of other issues anyway.

PR #339, opened 2020-01-22
Support more kinds of literals in hover

### LINKED ISSUE #274: Display source for literals in hover info
In other words, the TODO at: https://github.com/digital-asset/ghcide/blob/bbb75c2f95ca209db32defb2c10972968d4c829d/src/Development/IDE/Spans/AtPoint.hs#L81

serras:
Fixes #274

### DIFF
--- a/src/Development/IDE/Spans/Calculate.hs
+++ b/src/Development/IDE/Spans/Calculate.hs
@@ -158,7 +158,7 @@ getTypeLHsExpr tms e = do
     Nothing -> return Nothing
   where
     getSpanSource :: HsExpr GhcTc -> SpanSource
-    getSpanSource xpr | isLit True xpr = Lit (showGhc xpr)
+    getSpanSource xpr | isLit xpr = Lit (showGhc xpr)
     getSpanSource (HsVar U (L _ i)) = Named (getName i)
     getSpanSource (HsConLikeOut U (RealDataCon dc)) = Named (dataConName dc)
     getSpanSource RecordCon {rcon_con_name} = Named (getName rcon_con_name)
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Spans/Calculate.hs
```diff
@@ -158,15 +158,39 @@ getTypeLHsExpr tms e = do
...
+-- to please HLint in all versions
+#if MIN_GHC_API_VERSION(8,6,0)
+        isTupLit (Missing U)     = True
+#else
+        isTupLit Missing         = True
+#endif
+        isTupLit _               = False
+    isLit (ExplicitSum U _ _ xpr) = isLit (unLoc xpr)
+    isLit (ExplicitList U _ xprs) = all (isLit . unLoc) xprs
+    isLit (HsWrap U _ xpr) = isLit xpr
```
cocreature: This seems a bit weird, it looks like if I have something like `(1)`, I’ll get two literal spans, one for the wrapped `(1)` and one for just `1`. I think it makes more sense to just have the inner one. You probably still want to recurse below wrappings if you are doing a recursive check on something like `ExplicitList` so we probably want to differentiate two things:

1. Is this the root of a literal span. In this case, we don’t need to match on parentheses and probably also not on `ExprWithTySig`.
2. Can we consider whatever we started with (e.g. `ExplicitTuple`) a literal, if it has this as a child.

src/Development/IDE/Spans/Calculate.hs
```diff
@@ -158,15 +158,39 @@ getTypeLHsExpr tms e = do
...
+    isLit :: HsExpr GhcTc -> Bool
+    isLit (HsLit U _)     = True
+    isLit (HsOverLit U _) = True
+    isLit (ExplicitTuple U args _) = all (isTupLit . unLoc) args
+      where
+        isTupLit (Present U xpr) = isLit (unLoc xpr)
+-- the following #if is there
+-- to please HLint in all versions
+#if MIN_GHC_API_VERSION(8,6,0)
+        isTupLit (Missing U)     = True
```
cocreature: Treating tuple sections as literals feels a bit weird. Is this really what we want here?

serras: Right, this was a mistake.

src/Development/IDE/Spans/Calculate.hs
serras: This has been solved too

PR #344, opened 2020-01-24
Enhancements to Haddock -> Markdown conversion

### LINKED ISSUE #341: Some parts of the doc info (on hover) are incorrectly wrapped inside code tags
The overall feature looks neat, but some parts of the `:doc` are incorrectly wrapped in `<code></code>`

image

As @serras [points out](https://github.com/digital-asset/ghcide/pull/317#issuecomment-577678395), the origin of this may be at the transform of Haddock into Markdown.

---

Should this be reported in https://github.com/haskell/haddock instead, please let me know.

serras:
Fixes #341

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -216,6 +216,7 @@ test-suite ghcide-tests
         --------------------------------------------------------------
         ghcide,
         ghc-typelits-knownnat,
+        haddock-library,
         haskell-lsp-types,
         lens,
         lsp-test >= 0.8,
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -113,6 +113,7 @@ library
         Development.IDE.LSP.LanguageServer
         Development.IDE.LSP.Protocol
         Development.IDE.LSP.Server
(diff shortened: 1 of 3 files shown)


ndmitchell:
Any chance of some tests? I could imagine future changes breaking some of the more subtle bits.

serras:
I can try, but I've already had the problem that the documentation from packages changes from version to version. I'm planning to have a look at #7 and #270, and then we can write tests which reference files in the project itself.

ndmitchell:
I was thinking just a pure function, so take the Show of the DocH for unsafePerformIO as it stands today, paste it in as a test, and paste the current output as a string, and confirm that the conversion takes one to the other.

serras:
That makes sense. I'll write some of them during the weekend.

serras:
@ndmitchell the tests are in. Let me tell you what you think. Right now I've chosen two of the "hardest" ones: `Num` and `unsafePerformIO`, since they involve lists and mixed code.

ndmitchell:
Tests look great! Thanks for doing that.

PR #350, opened 2020-01-24
Enable -haddock at parsing time

### LINKED ISSUE #7: get haddock working
Apologies for the slightly vague bug report, but after spending a little while investigating already, I'd like to get an idea of what's actually supposed to work before digging in more depth.

There's code in [`Documentation.hs`](https://github.com/digital-asset/daml/blob/master/compiler/ghcide/src/Development/IDE/Spans/Documentation.hs) and [`AtPoint.hs`](https://github.com/digital-asset/daml/blob/master/compiler/ghcide/src/Development/IDE/Spans/AtPoint.hs) that appears to be intended to include haddock docs in the hover information. 

However I never see any, even with really trivial test cases. Is this something that is expected to work or is working for others?

One thing I noticed is that there are references in `Documentation.hs` to things happening when `Opt_Haddock` or `Opt_KeepRawTokenStream` are enabled, but I don't see them enabled anywhere.

serras:
Fixes #7

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -69,9 +69,10 @@ parseModule
     -> Maybe SB.StringBuffer
     -> IO ([FileDiagnostic], Maybe ParsedModule)
 parseModule IdeOptions{..} env file =
+    let env' = env { hsc_dflags = hsc_dflags env `gopt_set` Opt_Haddock } in
     fmap (either (, Nothing) (second Just)) .
     -- We need packages since imports fail to resolve otherwise.
-    runGhcEnv env . runExceptT . parseFileContents optPreprocessor file
+    runGhcEnv env' . runExceptT . parseFileContents optPreprocessor file
 
 
 -- | Given a package identifier, what packages does it depend on
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -1183,7 +1183,7 @@ findDefinitionAndHoverTests = let
   , test yes    yes    mclL36 mcl    "top-level fn 1st clause"
   , test yes    yes    mclL37 mcl    "top-level fn 2nd clause         #246"
   , test yes    yes    spaceL37 space "top-level fn on space #315"
-  , test no     broken docL41 doc    "documentation                     #7"
+  , test no     yes    docL41 doc    "documentation                     #7"
   , test no     yes    eitL40 kindE  "kind of Either                  #273"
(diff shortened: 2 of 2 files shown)


serras:
This means that we also get documentation for the modules the user is editing, not only for its dependencies.

src/Development/IDE/Core/Compile.hs
```diff
@@ -69,9 +69,10 @@ parseModule
     -> Maybe SB.StringBuffer
     -> IO ([FileDiagnostic], Maybe ParsedModule)
 parseModule IdeOptions{..} env file =
+    let env' = env { hsc_dflags = hsc_dflags env `gopt_set` Opt_Haddock } in
```
cocreature: GHC has both `Opt_KeepRawTokenStream` and `Opt_Haddock` which are mutually incompatible. We need `Opt_KeepRawTokenStream` in DAML since HLint will otherwise miss some hints and we’ll need the same in `ghcide` if we start integrating `HLint`.

So hardcoding this to `Opt_Haddock` isn’t really an option. Sadly, I don’t have a great answer for what the right option would be. I guess for now, we could add a `ForceHadock` field to `IdeOptions` and then set this to `True` for `ghcide` and if we do integrate `HLint` we accept that we miss some hints since docs are probably more useful. Long-term this should really get fixed in GHC so that the two are not exclusive (iirc the long-term plan is to build haddock on top of `Opt_KeepRawTokenStream`).

serras: Yeah, that seemed too easy to not have been tried earlier.

pepeiborra:
Parse twice, once for docs and another for hlint? Want Haddocks please.

serras:
The problem is that you would essentially double the memory consumption :(

One simple solution is to add `ghc-options: -haddock` to your Cabal file, or `{-# OPTIONS_GHC -haddock #-}` to your files, because then the option is set for your project/file.

pepeiborra:
No, you would double the amount of memory used to store the parsed AST. Even if it is inefficient, the cost would only be paid once/if hlint is integrated. 
Are there any downsides to enabling `-haddock` in my project file? Will that break Hlint?

serras:
As far as I know, if you run hlint independently of ghcide, there should be
no downsides. But maybe @cocreature knows better about this.

cocreature:
Running HLint independently is fine, this is about integrating HLint via Shake rules that rely on `GetParsedModule`.

PR #351, opened 2020-01-24
Keywords in completion

### LINKED ISSUE #347: Completions are missing `do`
When starting a file, and writing `main = do`, I get completions for `drop` etc. I suggest that keywords should be available as completions always. Particularly problematic because after `do` I often want a newline, but in this context enter inserts the word `drop`.

image

serras:
Fixes #347


src/Development/IDE/Core/Completions.hs
```diff
@@ -361,6 +361,7 @@ getCompletions ideOpts CC { allModNamesAsNS, unqualCompls, qualCompls, importabl
       filtImportCompls = filtListWith (mkImportCompl enteredQual) importableModules
       filtPragmaCompls = filtListWithSnippet mkPragmaCompl validPragmas
       filtOptsCompls   = filtListWith mkExtCompl
+      filtKeywrdCompls = if T.null prefixModule then filtListWith mkExtCompl keywords else []
```
fendor:
```suggestion
      filtKeywordCompls = if T.null prefixModule then filtListWith mkExtCompl keywords else []
```
Looks like a typo? Or is that intentional to match length of the other compls?

serras: The latter 🙄

cocreature: Just don’t align things. It only makes diffs noisy and introducing typo-like things to keep alignment is confusing.

serras: 👍 I've commited the suggestion

fendor: I am sorry, but the suggestion is probably not enough, arent you using the function in another location?

src/Development/IDE/Core/Completions.hs
```diff
@@ -380,6 +381,7 @@ getCompletions ideOpts CC { allModNamesAsNS, unqualCompls, qualCompls, importabl
         | otherwise
         = filtModNameCompls ++ map (toggleSnippets caps withSnippets
                                       . mkCompl ideOpts . stripAutoGenerated) filtCompls
+                            ++ filtKeywrdCompls
```
fendor:
```suggestion
                            ++ filtKeywordCompls
```

fendor: Sorry, too slow.

PR #357, opened 2020-01-25
Hide the completions module - they aren't part of the API

ndmitchell:


ndmitchell:
Declining since I think I'm going to move all the completions in a subsequent patch

PR #359, opened 2020-01-25
Move the completions and code actions into Plugin

ndmitchell:
Move the things into Plugin, expose them, and then put them back after. Aim is to migrate them to haskell/ide/plugin as two separate plugins.

Note this will require the DAML compiler to use these plugins.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -29,6 +29,7 @@ import Development.IDE.Types.Options
 import Development.IDE.Types.Logger
 import Development.IDE.GHC.Util
 import Development.IDE.Plugin.Completions
+import Development.IDE.Plugin.CodeAction
 import qualified Data.Text as T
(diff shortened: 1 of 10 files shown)


exe/Main.hs
```diff
@@ -83,11 +85,18 @@ main = do
 
     dir <- getCurrentDirectory
 
+    let handlers = def <>
```
alanz: I don't think you need `def` here, because of the way this are used in `runLanguageServer`.

pepeiborra:
Why not bundle handlers and rules together in a minimal `Plugin` type?

exe/Main.hs
ndmitchell: Thanks, will fix

ndmitchell:
@pepeiborra yep, that's the next patch.

PR #362, opened 2020-01-25
Improve recognition of GHC extension names

sheaf:
Some GHC extensions were not supported in the code actions, for instance `NamedFieldPuns` didn't work because within GHC the internal extension name is still `RecordPuns`. This patch fixes this issue.

### DIFF
--- a/src/Development/IDE/GHC/Compat.hs
+++ b/src/Development/IDE/GHC/Compat.hs
@@ -16,7 +16,6 @@ module Development.IDE.GHC.Compat(
     includePathsGlobal,
     includePathsQuote,
     addIncludePathsQuote,
-    ghcEnumerateExtensions,
     pattern DerivD,
     pattern ForD,
     pattern InstD,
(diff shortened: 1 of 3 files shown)


src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -12,6 +12,7 @@ module Development.IDE.LSP.CodeAction
     ) where
 
 import           Language.Haskell.LSP.Types
+import Control.Arrow((&&&))
```
ndmitchell: I'd have used Data.Tuple.Extra, which provides the same function but specialised to tuples, since no one can understand arrows...

ndmitchell:
Awesome! Any chance of putting a test for this somewhere? It's somewhat subtle, and exactly the kind of thing that will break in future.

src/Development/IDE/LSP/CodeAction.hs
```diff
@@ -364,10 +370,6 @@ extendToWholeLineIfPossible contents range@Range{..} =
         extend = newlineAfter && _character _start == 0 -- takes up an entire line, so remove the whole line
     in if extend then Range _start (Position (_line _end + 1) 0) else range
 
--- | All the GHC extensions
-ghcExtensions :: Set.HashSet T.Text
-ghcExtensions = Set.fromList $ map (T.pack . show) ghcEnumerateExtensions
```
ndmitchell: The ghcEnumerateExtensions function is super awful, and only used in this one place. Can this patch now delete it?

sheaf:
@ndmitchell I followed your suggestions, added tests and rebased. Thanks.

PR #364, opened 2020-01-26
Provide hie.yaml templates for stack and cabal

alanz:
And take hie.yaml out of version control, to be a local developer
preference instead.


cocreature:
Superseded by #522

PR #365, opened 2020-01-26
Add plugin type

ndmitchell:


src/Development/IDE/Plugin.hs
```diff
@@ -0,0 +1,33 @@
...
+import qualified Language.Haskell.LSP.Core as LSP
+import Language.Haskell.LSP.Messages
+
+
+data Plugin = Plugin
+    {pluginRules :: Rules ()
+    ,pluginHandler :: PartialHandlers
+    }
+
+instance Default Plugin where
```
cocreature: Not a big fan of `Data.Default`. Can we just export `defPlugin` instead?

src/Development/IDE/Plugin.hs
```diff
@@ -0,0 +1,33 @@
...
+    def = Plugin mempty def
+
+instance Semigroup Plugin where
+    Plugin x1 y1 <> Plugin x2 y2 = Plugin (x1<>x2) (y1<>y2)
+
+instance Monoid Plugin where
+    mempty = def
+
+
+codeActionPlugin :: (LSP.LspFuncs () -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO [CAResult]) -> Plugin
```
cocreature: Why is `codeActionPlugin` in here instead of `Development.IDE.Plugin.CodeAction`?

ndmitchell: The idea is that `codeActionPlugin` is the general helper for writing code actions, so when there are two code action plugins, they both use this helper, but have different internals. It's closer to the model HIE follows. The aim would eventually for there to be 5 or 6 such helpers in this module.

src/Development/IDE/Plugin.hs
ndmitchell: We're already using Data.Default for the rest of this API. I wanted to stay consistent.

cocreature: Fair enough

src/Development/IDE/Plugin.hs
cocreature: Makes nense.

Issue #325, opened 2020-01-15
Replace GetModificationTime with a file version scheme via WatchedFile events


pepeiborra:
The idea is to replace `GetModificationTime`, which is expensive and called on every change, by a  new rule `GetFileVersion` implemented entirely on top of watched file events. This would reduce I/O, eliminate timestamp serialization and comparison, and hopefully improve performance 

One possible approach is to generalize the `FileExistsMapVar` in order to keep track of file versions.

pepeiborra:
There's two reasons `GetModificationTime` is slow:
1. The non VFS code path is expensive as it needs to hit the file system, which can be addressed with WatchedFiles events
2. The rule is tagged as alwaysRerun, which could be addressed combining WatchedFiles and DidChangeTextDocument events.

It's likely that fixing 1 is good enough.

pepeiborra:
Implemented both 1 and 2, but doesn't seem to pay off massively. 

https://github.com/pepeiborra/ghcide/commits/get-file-modification

ndmitchell:
How slow did you measure getmodificationtime as?

pepeiborra:
I did not measure `getModificationTime` at all! It just seemed reasonable to reuse watched file notifications to avoid filesystem I/O here.

PR #373, opened 2020-01-26
Get file modification faster and cache filepath hashes

pepeiborra:
Implements #325 
Subject to more perf analysis
/cc @mpickering


pepeiborra:
https://github.com/mpickering/ghcide/commit/b7b389e847e985e80ead4f6de04630a2ea0ecc6e is a smaller change with much bigger improvement. 

To revisit after it is merged

PR #375, opened 2020-01-27
Don't support old haddock-library versions

ndmitchell:
I don't see any reason for supporting old versions of haddock-library. Might as well just demand the latest everywhere (and see if the CI disagrees with me).


serras:
That seems like a very sensible idea to me. When I wrote the code I was a bit worried of `haddock` being one of those libraries which cannot be easily updated without breaking GHC, but CI clearly says otherwise.

cocreature:
https://github.com/digital-asset/ghcide/pull/344/files added some more CPP for haddock-library so this should be rebased and remove that as well.

ndmitchell:
All gone now.

PR #378, opened 2020-01-27
See if fakeDynFlags is unnecessary

ndmitchell:
Shoving up for some CI love, and also needs some manual testing locally (which I'm starting now). @mpickering in #360 tries to delete fakeDynFlags. If it works, that's a nice simplification.

### DIFF
--- a/src/Development/IDE/GHC/Util.hs
+++ b/src/Development/IDE/GHC/Util.hs
@@ -1,10 +1,6 @@
 -- Copyright (c) 2019 The DAML Authors. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


cocreature:
I’m pretty sure that I’ve gotten errors when I used `unsafeGlobalDynFlags` just in GHCi for testing. That’s very useful for debugging so even if the tests pass, I think it probably makes sense to keep `fakeDynFlags`. It doesn’t add that much complexity.

ndmitchell:
Locally both `ghcide` on `ghcide` at the command line works, and everything I've managed to test manually in the IDE works (I'd expect if we were to see a failure it would be at the very very start, so even minimal success gives a strong degree of confidence).

Note that in GHC 8.10 the number of things in DynFlags that are strict increases significantly, which means that making a fakeDynFlags value becomes way more effort.

PR #382, opened 2020-01-27
Update to latest hie-bios

fendor:
Updates ghcide to use the latest hie-bios version 0.4.0

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -15,7 +15,6 @@ import Control.Exception
 import Control.Monad.Extra
 import Control.Monad.IO.Class
 import Data.Default
-import Data.Void (Void)
 import System.Time.Extra
 import Development.IDE.Core.FileStore
 import Development.IDE.Core.OfInterest
@@ -176,7 +175,7 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath -> file) diags) =
 showEvent lock e = withLock lock $ print e
 
 
-cradleToSession :: Cradle Void -> IO HscEnvEq
+cradleToSession :: Cradle a -> IO HscEnvEq
 cradleToSession cradle = do
     cradleRes <- getCompilerOptions "" cradle
     opts <- case cradleRes of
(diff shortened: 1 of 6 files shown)


cocreature:
I tested this as part of #361 and it seemed to work fine.

fendor:
What OS does CI run? E.g. I can not rely that it builds on windows only because CI is green, right?

cocreature:
CI only runs Linux but if `hie-bios` compiles on Windows in general I’m not too worried about that. We don’t have any windows-specific code in this part. And if you are worried about actually running rather than just compiling, our test suite isn’t going to be particularly helpful. We don’t test the hie-bios setup. All our tests just rely on `base` and not much else.

fendor:
I remember that I had to add an additional dependency for windows on hie. I will test it on my system later.

fendor:
Tested to build `stack84.yaml` on windows and it succeeded.

exe/Main.hs
```diff
@@ -15,6 +15,7 @@ import Control.Exception
 import Control.Monad.Extra
 import Control.Monad.IO.Class
 import Data.Default
+import Data.Void (Void)
```
ndmitchell: Generally no need to specify imports.

exe/Main.hs
```diff
@@ -175,7 +176,7 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath -> file) diags) =
 showEvent lock e = withLock lock $ print e
 
 
-cradleToSession :: Cradle -> IO HscEnvEq
+cradleToSession :: Cradle Void -> IO HscEnvEq
```
ndmitchell: Could this equally be `Cradle a`, so it was polymorphic in the type of Cradle it receives? That seems slightly simpler.

stack88.yaml
```diff
@@ -3,6 +3,7 @@ packages:
 - .
 extra-deps:
 - fuzzy-0.1.0.0
+- hie-bios-0.4.0
```
ndmitchell: Note this will clash with my changes in #400 - but I'm happy to rebase after this lands.

exe/Main.hs
fendor: Yeah sure, should not be a problem.

PR #383, opened 2020-01-27
Fixed some duplication

Gurkenglas:


cocreature:
#388 adds some test for this logic (we’ve had some before but in the wrong repo). Let’s wait for those tests and then rebase this PR on top of it.

src/Development/IDE/Core/PositionMapping.hs
```diff
@@ -32,53 +32,27 @@ idMapping :: PositionMapping
 idMapping = PositionMapping Just Just
 
 applyChange :: PositionMapping -> TextDocumentContentChangeEvent -> PositionMapping
-applyChange posMapping (TextDocumentContentChangeEvent (Just r) _ t) = PositionMapping
-    { toCurrentPosition = toCurrent r t <=< toCurrentPosition posMapping
-    , fromCurrentPosition = fromCurrentPosition posMapping <=< fromCurrent r t
+applyChange posMapping (TextDocumentContentChangeEvent (Just (Range start end)) _ t) = PositionMapping
+    { toCurrentPosition = morph start end (newEnd t start) <=< toCurrentPosition posMapping
```
cocreature: Probably easier to read if you keep `fromCurrent` and `toCurrent` wrappers for `morph` and then the tests should also compile.

jneira:
@cocreature wow didnt know about `azp run`, i guess you need admin rights to do it

cocreature:
@jneira you just need to be a collaborator on the repo, see https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#comment-triggers

src/Development/IDE/Core/Shake.hs
```diff
@@ -815,6 +815,6 @@ updatePositionMapping IdeState{shakeExtras = ShakeExtras{positionMapping}} Versi
         let uri = toNormalizedUri _uri
         let mappingForUri = Map.findWithDefault Map.empty uri allMappings
         let updatedMapping =
-                Map.insert _version idMapping $
-                Map.map (\oldMapping -> foldl' applyChange oldMapping changes) mappingForUri
+                Map.insert _version id $
+                Map.map ((.) $ foldr (.) id $ fmap toMapping changes) mappingForUri
```
Gurkenglas: oh dang, I think this is in reverse order.

cocreature: Are you saying the old code was incorrect? According to the LSP spec an edit array `[x,y]` is equivalent to first receiving an edit `x` and then an edit `y` so `applyChange (applyChange oldMapping x) y` which is equivalent to `foldl' applyChange oldMapping [x,y]`.

cocreature: I find the `Iso` stuff and the new code quite hard to read. I would prefer if we could stick to the code structure that we’ve had before.

cocreature:
I’m closing this for now to cleanup open PRs since it doesn’t pass CI and my questions haven’t been answered. Please feel free to reopen if you intend to work on this again.

PR #384, opened 2020-01-28
Improve performance by caching conversion to NormalizedUri

mpickering:
Two changes in this patch make performance quite a lot better for all actions.

1. Cache the conversion to `NormalizedUri` from `NormalizedFilePath`
2. Use a hash map for the debouncer to take advantage of the fast hash instance of `NormalizedUri`. 

The cached `Hashable` instance is in this PR: https://github.com/alanz/haskell-lsp/pull/214

My benchmark which ran hover 1000 times went down from 259s to 37s including initialisation time. Hovering on large projects is not noticeably faster (1s vs < 0.1s). 

Reviewer: @pepeiborra 

Note: I removed the Eq/Ord instance from Pepe's original patch because they relied on the hash for comparing equality. There are some hotspots on the latest profile from 

image

image

### DIFF
--- a/src/Development/IDE/Types/Location.hs
+++ b/src/Development/IDE/Types/Location.hs
@@ -58,7 +58,7 @@ import GHC.Generics
 --
 -- This is one of the most performance critical parts of ghcide, do not
 -- modify it without profiling.
-data NormalizedFilePath = NormalizedFilePath NormalizedUriWrapper Int !FilePath
+data NormalizedFilePath = NormalizedFilePath NormalizedUriWrapper !Int !FilePath
     deriving (Generic, Eq, Ord)
 
 instance NFData NormalizedFilePath where
--- a/src/Development/IDE/Types/Location.hs
+++ b/src/Development/IDE/Types/Location.hs
@@ -49,22 +49,55 @@ import Language.Haskell.LSP.Types as LSP (
   )
 import SrcLoc as GHC
 import Text.ParserCombinators.ReadP as ReadP
+import GHC.Generics
 
 
 -- | Newtype wrapper around FilePath that always has normalized slashes.
-newtype NormalizedFilePath = NormalizedFilePath FilePath
-    deriving (Eq, Ord, Show, Hashable, NFData, Binary)
+-- The NormalizedUri and hash of the FilePath are cached to avoided
+-- repeated normalisation when we need to compute them (which is a lot).
+--
+-- This is one of the most performance critical parts of ghcide, do not
+-- modify it without profiling.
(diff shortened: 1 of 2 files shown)


mpickering:
Branch with the change which tracks the haskell-lsp is here - https://github.com/mpickering/ghcide/tree/perf-haskell-lsp

src/Development/IDE/Types/Location.hs
```diff
@@ -49,22 +49,55 @@ import Language.Haskell.LSP.Types as LSP (
...
+instance Hashable NormalizedFilePath where
+  hash (NormalizedFilePath _ h _) = h
+
+-- Just to define NFData and Binary
+newtype NormalizedUriWrapper =
+  NormalizedUriWrapper { unwrapNormalizedFilePath :: NormalizedUri }
+  deriving (Show, Generic, Eq, Ord)
+
+instance NFData NormalizedUriWrapper where
+  rnf = rwhnf
```
pepeiborra: That `NFData` instance isn't correct either. Can we just lose it?

@ndmitchell does Shake care about these `NFData` instances?

pepeiborra: Nevermind, it is fine

pepeiborra:
Looks good to me!

src/Development/IDE/Types/Location.hs
```diff
@@ -49,22 +49,55 @@ import Language.Haskell.LSP.Types as LSP (
...
 
 -- | Newtype wrapper around FilePath that always has normalized slashes.
-newtype NormalizedFilePath = NormalizedFilePath FilePath
-    deriving (Eq, Ord, Show, Hashable, NFData, Binary)
+-- The NormalizedUri and hash of the FilePath are cached to avoided
+-- repeated normalisation when we need to compute them (which is a lot).
+--
+-- This is one of the most performance critical parts of ghcide, do not
+-- modify it without profiling.
+data NormalizedFilePath = NormalizedFilePath NormalizedUriWrapper Int !FilePath
```
cocreature: Similar question as for the `haskell-lsp` PR. Should we make the first two fields or at least the hash strict and unpack it?

mpickering: I can see the argument for making the hash strict but not the `NormalizedUriWrapper` as that is quite expensive to compute if you never use it.

mpickering: I made the hash strict.

cocreature: Thanks!

PR #392, opened 2020-01-31
Only show progress messages if the computation takes > 0.1s

ndmitchell:
At the moment it flickers a progress bar a lot, which is annoying. I first tried 0.3s but that made it almost never show, which felt wrong. 0.1 seems to have reasonable ergonomics.


ocharles:
Theres a rule of thumb in UI that sub 100ms stuff is snappy enough to not need feedback, 100-1000 is noticeable and beyond 1000 is basically "come back later". So 0.1s seems about right here!

ndmitchell:
Thanks. Rebased.

PR #394, opened 2020-01-31
Set targets to the current session.

jinwoo:
It may be useful later. It *is* useful in our internal patched version
of ghcide.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -186,7 +186,8 @@ cradleToSession cradle = do
         CradleNone -> fail "'none' cradle is not yet supported"
     libdir <- getLibdir
     env <- runGhc (Just libdir) $ do
-        _targets <- initSession opts
+        targets <- initSession opts
+        setTargets targets
         getSession
     initDynLinker env
     newHscEnvEq env


mpickering:
Without your other internal changes, this change doesn't make much sense. The information won't be used anywhere.

jinwoo:
I'm fine with abandoning this PR because this is already patched internally. I just thought it'd be no harm to call `setTargets` anyway, so we could reduce the number of patched diffs :) I also thought it'd make a sense to use the return value from `initSession`

If you want to avoid having "unnecessary" code, please just let me know.

mpickering:
Calling `setTargets` is cheap, but it's my opinion that it's confusing to have code which doesn't do anything. Especially as in other Haskell IDEs, it does do something..

Perhaps your other change would be useful to upstream anyway?

jinwoo:
Our related change seems specific to our build system (internal variant of bazel). When we find imported modules (https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Import/FindImports.hs#L49), we shouldn't locate them from random places in the file system, but should only locate them from the target files that are specified in the build rule. Our internal patch uses the `targets` for that purpose. I'm not sure that's useful in other build systems.

I'm closing this PR now.

ndmitchell:
This kind of change is likely to help other Bazel users (eg Digital Asset). Any change of a complete list of changes? I think we'd be happy to figure out the right abstractions that either let you have no patches and instead adjust Options, or perhaps even merge it all.

cocreature:
I don’t think this particular change would help us at Digital Asset, we want to be able to load all of our targets into a single GHCi session so as a general policy we don’t allow colliding module names which means that just searching in the include dirs works fine.

That said, I’m happy for things to be upstreamed but it should be in a somewhat coherent form. For this particular PR I agree with @mpickering that it’s confusing on its own.

PR #396, opened 2020-02-01
Allow the withResponse handler to idiomatically return an error

### LINKED ISSUE #395: How do handlers return errors?
We have

```haskell
data WithMessage = WithMessage
    {withResponse :: forall m req resp . (Show m, Show req) =>
        (ResponseMessage resp -> LSP.FromServerMessage) -> -- how to wrap a response
        (LSP.LspFuncs () -> IdeState -> req -> IO resp) -> -- actual work
        Maybe (LSP.Handler (RequestMessage m req resp))
...
    }
```
But `ResponseMessage resp` defines a response message with a `result` field of type `resp`.  A response message may not have a `result` field, but an `error` field instead. (see https://github.com/alanz/haskell-lsp/issues/191)

If the handler has an error, how is this currently handled?

Should the work part rather be

```haskell
(LSP.LspFuncs () -> IdeState -> req -> IO (Either ResponseError resp))
```
?

alanz:
An LSP response message can have either a result or an error
field. Expose this in the handler by having a return type

    type ResponseBody resp = Either ResponseError resp

Closes #395

### DIFF
--- a/src/Development/IDE/LSP/Server.hs
+++ b/src/Development/IDE/LSP/Server.hs
@@ -6,6 +6,7 @@
 module Development.IDE.LSP.Server
   ( WithMessage(..)
   , PartialHandlers(..)
+  , ResponseBody
   ) where
 
 
@@ -16,17 +17,19 @@ import qualified Language.Haskell.LSP.Core as LSP
 import qualified Language.Haskell.LSP.Messages as LSP
 import Development.IDE.Core.Service
 
+-- | An LSP Response message can either have an error or a result.
+type ResponseBody resp = Either ResponseError resp
 
 data WithMessage = WithMessage
     {withResponse :: forall m req resp . (Show m, Show req) =>
         (ResponseMessage resp -> LSP.FromServerMessage) -> -- how to wrap a response
-        (LSP.LspFuncs () -> IdeState -> req -> IO resp) -> -- actual work
+        (LSP.LspFuncs () -> IdeState -> req -> IO (ResponseBody resp)) -> -- actual work
         Maybe (LSP.Handler (RequestMessage m req resp))
     ,withNotification :: forall m req . (Show m, Show req) =>
         Maybe (LSP.Handler (NotificationMessage m req)) -> -- old notification handler
         (LSP.LspFuncs () -> IdeState -> req -> IO ()) -> -- actual work
         Maybe (LSP.Handler (NotificationMessage m req))
-    ,withResponseAndRequest :: forall m rm req resp newReqParams newReqBody. 
+    ,withResponseAndRequest :: forall m rm req resp newReqParams newReqBody.
         (Show m, Show rm, Show req, Show newReqParams, Show newReqBody) =>
         (ResponseMessage resp -> LSP.FromServerMessage) -> -- how to wrap a response
         (RequestMessage rm newReqParams newReqBody -> LSP.FromServerMessage) -> -- how to wrap the additional req
--- a/src/Development/IDE/LSP/Server.hs
+++ b/src/Development/IDE/LSP/Server.hs
@@ -6,7 +6,6 @@
 module Development.IDE.LSP.Server
   ( WithMessage(..)
   , PartialHandlers(..)
-  , ResponseBody
   ) where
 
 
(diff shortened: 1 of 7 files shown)


alanz:
Note: on my local machine there is one test failure, which also shows up on master.

ndmitchell:
@alanz is it related to cpp and the error message that comes out? I get that one on my machine - not had the time to track it down yet.

src/Development/IDE/LSP/Server.hs
```diff
@@ -16,17 +17,19 @@ import qualified Language.Haskell.LSP.Core as LSP
 import qualified Language.Haskell.LSP.Messages as LSP
 import Development.IDE.Core.Service
 
+-- | An LSP Response message can either have an error or a result.
+type ResponseBody resp = Either ResponseError resp
```
ndmitchell: I'd be tempted to inline this type everywhere. It's somewhat annoying to have, and not vastly shorter than the expanded version.

src/Development/IDE/Plugin.hs
```diff
@@ -26,8 +26,13 @@ instance Monoid Plugin where
...
-codeActionPlugin :: (LSP.LspFuncs () -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO [CAResult]) -> Plugin
+codeActionPlugin :: (LSP.LspFuncs () -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO (ResponseBody [CAResult])) -> Plugin
 codeActionPlugin f = Plugin mempty $ PartialHandlers $ \WithMessage{..} x -> return x{
     LSP.codeActionHandler = withResponse RspCodeAction g
     }
-    where g lsp state (CodeActionParams a b c _) = List <$> f lsp state a b c
+    where
+      g lsp state (CodeActionParams a b c _) = do
+        r <- f lsp state a b c
+        case r of
```
ndmitchell: Isn't this just `return $ List <$> r`?

src/Development/IDE/Plugin.hs
```diff
@@ -26,8 +26,13 @@ instance Monoid Plugin where
     mempty = def
 
 
-codeActionPlugin :: (LSP.LspFuncs () -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO [CAResult]) -> Plugin
+codeActionPlugin :: (LSP.LspFuncs () -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO (ResponseBody [CAResult])) -> Plugin
 codeActionPlugin f = Plugin mempty $ PartialHandlers $ \WithMessage{..} x -> return x{
     LSP.codeActionHandler = withResponse RspCodeAction g
     }
-    where g lsp state (CodeActionParams a b c _) = List <$> f lsp state a b c
```
ndmitchell:
```
where g lsp state (CodeActionParams a b c _) = fmap List <$> f lsp state a b c
```

src/Development/IDE/LSP/Server.hs
alanz: I brought it in to tag the fact that this is a standard reply type. It is basically just documentation.

If you feel it should go, it can go.

src/Development/IDE/Plugin.hs
alanz: Yes. It was more complicated before, then simplified down.

alanz:
@ndmitchell the test failure is related to CPP.

```
  plugins:                                            FAIL (0.10s)
    test/src/Development/IDE/Test.hs:35:
    Could not find (DsError,(8,14),"Variable not in scope: c") in List [Diagnostic {_range = Range {_start = Position {_line = 0, _character = 0}, _end = Position {_line = 100000, _character = 0}}, _severity = Just DsError, _code = Nothing, _source = Just "typecheck", _message = "Unexpected usage error\nCould not find module \8216GHC.TypeLits.KnownNat.Solver\8217\nUse -v to see a list of the files searched for.\n", _relatedInformation = Nothing}]
  preprocessor:                                       OK (0.62s)
```

src/Development/IDE/LSP/Server.hs
ndmitchell: That would be the way I would have done it, but it's a reasonable choice either way.

ndmitchell:
@alanz The error I get is:

```
    cpp-error:                                        FAIL (3.63s)
      test/src\Development\IDE\Test.hs:35:
      Could not find (DsError,(2,1),"error: unterminated") in List [Diagnostic {_range = Range {_start = Position {_line = 0, _character = 0}, _end = Position {_line = 100000, _character = 0}}, _severity = Just DsError, _code = Nothing, _source = Just "compiler", _message = "`gcc.exe' failed in phase `C pre-processor'. (Exit code: 1)", _relatedInformation = Nothing}]
```

The one you get seems to be because the package `ghc-typelits-knownnat` isn't installed/available properly when running the tests. I am running the tests with `stack test` on Windows. How are you doing it?

alanz:
I am running with `cabal new-test`. Will experiment with that too, and possibly add that package to the deps somewhere if it is needed. We have some packages like that in the hie stack files too. Only needed for testing.

src/Development/IDE/LSP/Server.hs
cocreature: I prefer to just inline it. `ResponseBody` isn’t quite accurate since it also includes the “status” (in HTTP terms) of whether it has failed or not and we already have a type called `ResponseMessage` so regardless of what we chose here it will be confusing.

alanz:
@ndmitchell the test fails with `cabal new-test`, but passes with `stack test`. FYI.

ndmitchell:
@alanz I suggest you create a ticket for the test failure with details of how to reproduce, OS etc

cocreature:
The tests rely on `build-depends` of the test leaking into the `ghcide` environment in the test. That seems more like something that accidentally works in Stack rather than something you can rely on so I’m not really surprised this falls apart with cabal.

PR #397, opened 2020-02-01
Allow getting file contents as Text 

alanz:
The VFS already provides it as Text, and getFileContents converts it
to a StringBuffer.

Expose the intermediate step so that tools such as formatters can
access it without a double conversion.

### DIFF
--- a/src/Development/IDE/Core/FileStore.hs
+++ b/src/Development/IDE/Core/FileStore.hs
@@ -14,9 +14,7 @@ module Development.IDE.Core.FileStore(
     makeLSPVFSHandle
     ) where
 
-import           StringBuffer
 import Development.IDE.GHC.Orphans()
-import Development.IDE.GHC.Util
 import           Development.IDE.Core.Shake
 import Control.Concurrent.Extra
 import qualified Data.Map.Strict as Map
@@ -85,7 +83,7 @@ makeLSPVFSHandle lspFuncs = VFSHandle
 
 
 -- | Get the contents of a file, either dirty (if the buffer is modified) or Nothing to mean use from disk.
-type instance RuleResult GetFileContents = (FileVersion, Maybe StringBuffer)
+type instance RuleResult GetFileContents = (FileVersion, Maybe T.Text)
 
 data GetFileContents = GetFileContents
     deriving (Eq, Show, Generic)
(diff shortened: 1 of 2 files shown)


alanz:
Note

a) This Diff looks bigger than it should be, as it is built on top of #396.  It is actually just ca0bf6d801d48f75437aeeb636630e3d574c1e17
b) There is one test that fails locally, on a timeout ("lower-case drive"). It passes for me locally without this commit. I am not sure if I am doing something wrong in the rule dependencies, or if the test is just brittle.

src/Development/IDE/Core/FileStore.hs
```diff
@@ -93,6 +94,16 @@ instance Hashable GetFileContents
 instance NFData   GetFileContents
 instance Binary   GetFileContents
 
+-- | Get the contents of a file as text, either dirty (if the buffer is modified) or Nothing to mean use from disk.
+type instance RuleResult GetFileContentsText = (FileVersion, Maybe T.Text)
```
cocreature: Representing every file twice in the shake graph is rather wasteful. Not just in terms of memory usage but also because those rules will fire fairly often and the diagnostics computations in `updateFileDiagnostics` are not free (iirc that showed up quite a lot in the profiles Matthew took).

Which autoformatters actually need this? If we are worried about efficiency, an autoformatter should probably not read the text at all but instead depend on `GetParsedModule` (that’s how the HLint integration in Daml works). Obviously that’s not an option for autoformatters that use `haskell-src-exts` but given that this has been declared dead, I wouldn’t want to optimize for that. And if the autoformatter does have to reparse the module, I’m not sure that the conversion from `StringBuffer` to `Text` will be the bottleneck.

alanz: I agree that formatting is relatively rare in the scheme of things, and happening at human scale. So will revert and convert from the `StringBuffer` to Text as needed.

cocreature: Another option would be to change `GetFileContents` to give you a `Text`. Given that the only usesite atm of that rule is in `GetParsedModule` we’ll only pay for the conversion once so it doesn’t matter where it happens.

alanz: That makes sense.

ndmitchell: Note that the conversion of Text to StringBuffer currently is a complete mess. I'd be shocked if it never shows up on the profile, so I wouldn't want to do two conversions. I think that the updateFileDiagnostics is slow is really a problem in itself - if we have to pay more than minimal costs for Shake nodes on the graph, we're going to start cheating ourselves out of the nicest graph.

alanz: I updated the PR to only do the StringBuffer conversion for `getParsedModule`, where it is actually needed. So we are only doing it once, and for a guaranteed need.

What could be interesting would be to get the VFS to operate in StringBuffer mode, somehow.

Issue #381, opened 2020-01-27
Progress cancellation sometimes gets lost


ndmitchell:
I had this happen a lot during the weekend (on my too slow for words laptop), and it's just happened on my desktop. I get the progress stuck at nearly complete - which I think is actually the progress completing but failing to send the progress completion message. As yet I have no way to reproduce. Looking at the code, either Shake fails to kill the thread, or killing the thread fails to send the message, or interleaving/VSCode is at fault.

ndmitchell:
I added printing of start and stop messages. Indeed, it sometimes gets lost. Out of 22 progress notifications (while doing random dev work), 3 were lost. A few traces:

```
("PROGRESS Start",ProgressTextToken "8")
[DEBUG] Cancelled request IdInt 19
[DEBUG] Starting shakeRun (aborting the previous one took 0.00s)
[DEBUG] Finishing shakeRun (took 0.00s, exception: AsyncCancelled)
("PROGRESS Start",ProgressTextToken "9")
```

```
("PROGRESS Start",ProgressTextToken "13")
[DEBUG] Cancelled request IdInt 36
[DEBUG] Cancelled request IdInt 37
[DEBUG] Cancelled request IdInt 38
[DEBUG] Finishing shakeRun (took 0.00s, exception: AsyncCancelled)
[DEBUG] Cancelled request IdInt 39
[DEBUG] Starting shakeRun (aborting the previous one took 0.00s)
[INFO] Modified text document: f("PROGRESS Start",ProgressTextToken "14")
```

There is no `PROGRESS Stop` message. They all have an AsyncCancelled in the middle. I am not sure if that is the only case where they get lost, or if all cancellations lead to lost progress stop. The trace didn't show any obvious AsyncCancel where the stop did get sent, but some were ambiguously close due to possible overlapping input.

Next step is to test in isolation that Shake cancels the progress even if the main thread is cancelled, probably as a Shake test.

ndmitchell:
This is a Shake bug which I've replicated outside of ghcide. Hopefully fixed in the next Shake release.

cocreature:
Awesome, looking forward to that!

ndmitchell:
The fix is in https://github.com/ndmitchell/shake/commit/5d991fee03f63020a5f32531b3ffe559b94d2272. The allocate thread cancels the thread, which involves read/writing a Var, and then waiting on a Barrier. Both those things are interruptible, so switching over to an uninterruptibleMask means they get to finish.

[...]

PR #400, opened 2020-02-03
#381, require shake-0.18.5, which ensures progress cancellation is robust

ndmitchell:
Now the progress thread gets killed more robustly, so progress messages always stop.


ndmitchell:
Note this Stackage bump also moves us from GHC 8.8.1 to 8.8.2.

ndmitchell:
Also requires us explicitly not pulling in the latest hie-bios, which is shipped in nightly.

PR #401, opened 2020-02-03
Make ghcide on the command line use absolute file paths

ndmitchell:
Ghcide via LSP uses absolute paths, but via the command line tester uses relative paths. As a consequence the nasty thing I was debugging worked perfectly, so better to make everything work the same if we possibly can.


ndmitchell:
Yep - it's #402 - but it's not infeasible its not the only such issue.

PR #408, opened 2020-02-05
Automatically pick up new dependencies

jinwoo:
hie-bios's componentDependencies returns the dependencies of a cradle
that might change the cradle. Add those deps to the shake graph so that
the GHC session is newly created whenever they change.

For that, add a new rule type, GetHscEnvEq, to cache GHC sessions with
the key of GHC options and dependencies. And delete the optGhcSession
field from IdeOptions.

This is for https://github.com/digital-asset/ghcide/issues/50.

hie-bios's componentDependencies can return files that don't exist yet:
https://github.com/mpickering/hie-bios/blob/master/src/HIE/Bios/Types.hs#L90-L93.
This PR handles changes in the existing dependency files, but doesn't
handle newly created dependency files.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -3,14 +3,19 @@
 {-# OPTIONS_GHC -Wno-dodgy-imports #-} -- GHC no longer exports def in GHC 8.6 and above
 {-# LANGUAGE CPP #-} -- To get precise GHC version
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
 
 module Main(main) where
 
 import Arguments
+import Data.Binary (Binary)
+import Data.Dynamic (Typeable)
+import Data.Hashable (Hashable)
 import Data.Maybe
 import Data.List.Extra
 import System.FilePath
 import Control.Concurrent.Extra
+import Control.DeepSeq (NFData)
 import Control.Exception
 import Control.Monad.Extra
 import Control.Monad.IO.Class
@@ -45,11 +50,12 @@ import System.IO
 import System.Exit
 import Paths_ghcide
 import Development.GitRev
-import Development.Shake (Action, Rules, action, doesFileExist, need)
+import Development.Shake (Action, RuleResult, Rules, action, doesFileExist, need)
 import qualified Data.HashSet as HashSet
 import qualified Data.Map.Strict as Map
 
 import GHC hiding (def)
+import GHC.Generics (Generic)
 import qualified GHC.Paths
 
 import HIE.Bios
@@ -99,12 +105,12 @@ main = do
         runLanguageServer def (pluginHandler plugins) onInitialConfiguration onConfigurationChange $ \getLspId event vfs caps -> do
             t <- t
             hPutStrLn stderr $ "Started LSP server in " ++ showDuration t
-            let options = defaultIdeOptions
+            let options = (defaultIdeOptions $ loadSession dir)
                     { optReportProgress = clientSupportsProgress caps
                     , optShakeProfiling = argsShakeProfiling
                     }
             debouncer <- newAsyncDebouncer
-            initialise caps (loadGhcSessionIO dir >> mainRule >> pluginRules plugins >> action kick)
+            initialise caps (loadGhcSessionIO >> mainRule >> pluginRules plugins >> action kick)
                 getLspId event (logger minBound) debouncer options vfs
     else do
         putStrLn $ "Ghcide setup tester in " ++ dir ++ "."
@@ -137,12 +143,11 @@ main = do
         let grab file = fromMaybe (head sessions) $ do
                 cradle <- Map.lookup file filesToCradles
                 Map.lookup cradle cradlesToSessions
-        let loadGhcSession =
-              defineNoFile $ \GhcSessionIO ->
-                  return $ GhcSessionFun (return . grab)
 
-        let options = defaultIdeOptions { optShakeProfiling = argsShakeProfiling }
-        ide <- initialise def (loadGhcSession >> mainRule) (pure $ IdInt 0) (showEvent lock) (logger Info) noopDebouncer options vfs
+        let options =
+              (defaultIdeOptions $ return $ return . grab)
+                    { optShakeProfiling = argsShakeProfiling }
+        ide <- initialise def (loadGhcSessionIO >> mainRule) (pure $ IdInt 0) (showEvent lock) (logger Info) noopDebouncer options vfs
 
         putStrLn "\nStep 6/6: Type checking the files"
         setFilesOfInterest ide $ HashSet.fromList $ map toNormalizedFilePath files
@@ -183,11 +188,26 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath -> file) diags) =
 showEvent lock e = withLock lock $ print e
 
 
-loadGhcSessionIO :: FilePath -> Rules ()
-loadGhcSessionIO dir = do
+-- Rule type for caching GHC sessions.
+type instance RuleResult GetHscEnv = HscEnvEq
+
+data GetHscEnv = GetHscEnv
+    { hscenvOptions :: [String]        -- componentOptions from hie-bios
+    , hscenvDependencies :: [FilePath] -- componentDependencies from hie-bios
+    }
+    deriving (Eq, Show, Typeable, Generic)
+instance Hashable GetHscEnv
+instance NFData   GetHscEnv
+instance Binary   GetHscEnv
+
+
+loadGhcSessionIO :: Rules ()
+loadGhcSessionIO =
+    -- This rule is for caching the GHC session. E.g., even when the cabal file
+    -- changed, if the resulting flags did not change, we would continue to use
+    -- the existing session.
     defineNoFile $ \(GetHscEnv opts deps) ->
         liftIO $ createSession $ ComponentOptions opts deps
-    defineNoFile $ \GhcSessionIO -> GhcSessionFun <$> loadSession dir
 
 
 getComponentOptions :: Cradle a -> IO ComponentOptions
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -39,13 +39,13 @@ import Language.Haskell.LSP.Types (LspId(IdInt))
 import Linker
 import Data.Version
 import Development.IDE.LSP.LanguageServer
-import System.Directory.Extra as IO
+import qualified System.Directory.Extra as IO
 import System.Environment
 import System.IO
 import System.Exit
 import Paths_ghcide
 import Development.GitRev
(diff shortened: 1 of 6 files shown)


jinwoo:
I'm wondering how I can write a test for this. Is it possible to write a test that simulates changes in files like `ghcide.cabal`?

src/Development/IDE/Core/Rules.hs
```diff
@@ -334,11 +334,21 @@ instance Show GhcSessionFun where show _ = "GhcSessionFun"
 instance NFData GhcSessionFun where rnf !_ = ()
 
 
+-- Rule type for caching GHC sessions.
+type instance RuleResult GetHscEnvEq = HscEnvEq
+
+data GetHscEnvEq = GetHscEnvEq
```
ndmitchell: I'd have called this `GetHscEnv` since that's the real underlying type. The `Eq` is kind of a hack to get equality on it.

hie.yaml
```diff
@@ -1 +1,3 @@
 cradle: {stack: {component: "ghcide:lib"}}
+dependencies:
+  - ghcide.cabal
```
ndmitchell: Hmm, I think hie-bios should auto-detect this as a dependency.

src/Development/IDE/Core/Rules.hs
ndmitchell: Or why not just reuse the `ComponentOptions` type? Since all you ever do is convert it to that. If you want a newtype for the key (which isn't unreasonable) just `newtype GetHscEnv = GetHscEnv ComponentOptions` should be fine.

exe/Main.hs
```diff
@@ -194,19 +206,31 @@ cradleToSession cradle = do
...
 
-loadSession :: FilePath -> IO (FilePath -> Action HscEnvEq)
-loadSession dir = do
+cradleToSession :: Cradle a -> Action HscEnvEq
+cradleToSession cradle = do
+    cmpOpts <- liftIO $ getComponentOptions cradle
+    let opts = componentOptions cmpOpts
+        deps = componentDependencies cmpOpts
+    -- TODO Handle deps that don't exist yet but might be newly created.
+    existingDeps <- liftIO $ filterM doesFileExist deps
```
ndmitchell: If you use the Shake doesFileExist then I think you'll automatically end up tracking the absence of the files that haven't yet been created, thus solving the issue you mention.

jinwoo: Aha, that's convenient. Done. Had to change the import of `System.Directory.Extra` as a qualified import to disambiguate `doesFileExist`, and update several places in this file.

hie.yaml
jinwoo: hie-bios gives package.yaml and stack.yaml, but not .cabal files. I think it makes sense because in most stack-generated projects, the .cabal file is generated from package.yaml.

src/Development/IDE/Core/Rules.hs
jinwoo: Renamed to `GetHscEnv`.

Currently, only the `Main` module (and the `executable` component of this project) depends on hie-bios and I don't want to add that dependency to the `library` component just for this. And  because `ComponentOptions` comes from hie-bios, I'm not sure I can define instances of `Hashable`, `NFData`, and `Binary` for `GetHscEnv` in that case.

hie.yaml
fendor: https://github.com/mpickering/hie-bios/blob/master/src/HIE/Bios/Cradle.hs#L486
Nope, I think it is a bug. You are the first to actually use that feature, thus it makes sense that you are discovering bugs :(

jinwoo: I'm not actually sure I checked this. Let me check now.

jinwoo: I was wrong. ghcide.cabal file is also picked up automatically by hie-bios. Reverted. Thanks, @fendor !

src/Development/IDE/Core/Rules.hs
ndmitchell: Ah, good point about the dependencies. Your approach is definitely the right way to go.

jinwoo:
Rebased and resolved merge conflicts.

src/Development/IDE/Core/Rules.hs
```diff
@@ -334,11 +334,21 @@ instance Show GhcSessionFun where show _ = "GhcSessionFun"
 instance NFData GhcSessionFun where rnf !_ = ()
 
 
+-- Rule type for caching GHC sessions.
```
cocreature: Given that this rule is only used in `Main.hs`, I think we should move it there (or some helper module for the executable). The `ghcide` library explicitly abstracts over how you setup the environment so that consumers like `daml` can plug in something cmpletely custom.

exe/Main.hs
```diff
@@ -134,11 +134,12 @@ main = do
         let grab file = fromMaybe (head sessions) $ do
                 cradle <- Map.lookup file filesToCradles
                 Map.lookup cradle cradlesToSessions
+        let loadGhcSession =
```
cocreature: Why do we setup the environment slightly differently for the CLI mode and the LSP mode? Can we not use `loadGhcSessionIO =<< getCurrentDirectory` here?

exe/Main.hs
```diff
@@ -179,15 +180,26 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath -> file) diags) =
 showEvent lock e = withLock lock $ print e
 
 
-cradleToSession :: Cradle a -> IO HscEnvEq
-cradleToSession cradle = do
+loadGhcSessionIO :: FilePath -> Rules ()
+loadGhcSessionIO dir = do
+    defineNoFile $ \(GetHscEnv opts deps) ->
```
cocreature: Why do we need this separate rule? Is it just to get more caching, i.e., you don’t need to reinitialize the session if the cabal file changed but the resulting flags did not change? Would be good to document this.

src/Development/IDE/Types/Options.hs
```diff
@@ -25,12 +23,6 @@ data IdeOptions = IdeOptions
   { optPreprocessor :: GHC.ParsedSource -> IdePreprocessedSource
     -- ^ Preprocessor to run over all parsed source trees, generating a list of warnings
     --   and a list of errors, along with a new parse tree.
-  , optGhcSession :: IO (FilePath -> Action HscEnvEq)
```
cocreature: Instead of removing this can we simply change it to `Action (FilePath -> Action HscEnvEq)`? By removing this you are turning an explicit dependency into an implicit one (if I don’t define the rule, things will explode).

src/Development/IDE/Core/Rules.hs
jinwoo: I initially added this rule to `Main.hs` and realized that it required adding several new dependencies to the executable component: `hashable`, `deepseq`, and `binary`. I can move this back to `Main.hs` if that makes things cleaner for `daml`. Just let me know.

exe/Main.hs
jinwoo: The difference was already there. The CLI mode used the `grab` function for creating cradles/sessions and my change doesn't change that aspect. Do you want the `grab` function to be deleted? I'm not sure how it affects the behavior in the CLI mode.

jinwoo:
> For tests, I would try to setup a simple test project in a temp directory and then just modify the cabal file in there.

One more question. Do we have any existing tests that do this so that I can steal the code from them? :)

cocreature:
> > 
> 
> One more question. Do we have any existing tests that do this so that I can steal the code from them?

Sadly I think we don’t. All our existing tests don’t have a `.cabal` file. That said, I don’t think it should change much. Basically in addition to the file, you just write out the cabal fie and a `hie.yaml` and hopefully things should work out :crossed_fingers:

exe/Main.hs
jinwoo: I think I know what you meant. Now that `optGhcSession` is back in `IdeOptions`, `loadGhcSession` is gone.

src/Development/IDE/Core/Rules.hs
jinwoo: Moved to `Main.hs`.

src/Development/IDE/Types/Options.hs
jinwoo: Added back as you suggested. That allowed me to reduce the number of changes in this PR. And it also removed the `dir` argument of the `loadGhcSessionIO` rule. Thanks!

ghcide.cabal
```diff
@@ -187,6 +190,7 @@ executable ghcide
         Paths_ghcide
 
     default-extensions:
+        DeriveGeneric
```
jinwoo: This was added here because HLint complains when I added it to `Main.hs`:

```
./exe/Main.hs:7:1: Warning: Avoid restricted extensions
Found:
  {-# LANGUAGE DeriveGeneric #-}
Note: may break the code
```

ndmitchell: Indeed, that's the right thing to do. At Digital Asset we made a list of extensions we felt were "boring", and thus should never occur in a .hs file, only in default-extensions or other places.

jinwoo:
Sorry for the late response. I added a test that verified that the changes in `package.yaml` is picked up and reflected correctly. PTAL.

jinwoo:
Hmm... The new test seems flaky, sometimes timing out. I'm not sure how I can make it stable...

jinwoo:
Updated the test. Hopefully now it's stable.

test/exe/Main.hs
```diff
@@ -1725,13 +1726,77 @@ haddockTests
...
+sessionDepsArePickedUp = testSession'
+  "session-deps-are-picked-up"
+  $ \dir -> do
+    liftIO $ do
+      writeFileUTF8 (dir </> "hie.yaml") "cradle: {stack: {component: \"foo:lib\"}}"
+      writeFileUTF8 (dir </> "stack.yaml") stackYamlContent
+      writeFileUTF8 (dir </> "package.yaml") oldPackageYamlContent
+    -- Open without OverloadedStrings and expect an error.
+    doc <- openDoc' "Foo.hs" "haskell" fooContent
+    expectDiagnostics
```
cocreature: I don’t really want to download a stack GHC bindist in a test and this is also timing out on CI. Can we instead use a `direct-style` cradle and test that changes to that are picked up?

jinwoo: `direct-style` cradle doesn't work at the moment because it doesn't give any files as the session dependencies. I think it should provide the `hie.yaml` file itself as a dependency but it doesn't currently. Let me see if I can fix that before making changes here.

jinwoo: I made a small change in the `cradleToSession` function in `exe/Main.hs` to make this feature work for direct-style cradles. And also updated this test using a direct-style cradle.

jinwoo:
Now all the CI passed! PTAL

Issue #355, opened 2020-01-25
Use interface files (*.hi) when available


pepeiborra:
Interface files could help speed up loading times and reduce memory usage

ocharles:
Wouldn't this risk ghcide using out-of-date information, or am I misunderstanding the suggestion? The only .hi files we have are from `cabal build`, which is possibly much more out of date than what `ghcide` can see. Or is there a way to know when a `.hi` is stale (that's faster than just building a new one)?

pepeiborra:
Interface files would only be used for dependencies, and to determine staleness I was thinking just comparing timestamps with the matching source file.

pepeiborra:
Teamed up with @mpickering and @lazamar working on this during the Bristol Hackathon. Current state: ghcide will use interface files in the typechecking rule when available. On the GHC codebase, loading the GHC module uses "only" 3GB as opposed to 5.6GB as it used to, and starts up significantly more quickly. 
Because interface files carry less information, goto definition is not available for names in modules loaded from interface files. This can be remediated by falling back to `.hie` files, but these only exist in GHC >= 8.8.

Things missing:

- [ ] 1 test failure involving Template Haskell
- [ ] Interface file loading reports errors occasionally, logic needs reviewing
- [ ] Enable feature only if GHC >= 8.8
- [ ] Automatic generation of `.hie` files (if not available)
- [ ] Tests

pepeiborra:
New branch: https://github.com/pepeiborra/ghcide/tree/interface-files-rebased-again

mpickering:
I tried the latest branch and receive the following error when trying to load GHC.

```
Step 6/6: Type checking the files
typecheckIface
Declaration for ComponentId
Constructor ComponentId
Constructor ComponentId arg_tys:
  attempting to use module ‘ghc-8.8.1:FastString’ (compiler/utils/FastString.hs) which is not loaded
File:     compiler/basicTypes/Module.hs
Hidden:   no
Range:    1:0-100001:0
Source:   typecheck
Severity: DsError
Message:  Program error: Cannot continue after interface file error
```

mpickering:
The check to use interface files also needs to check that the interface files are built for the right way I think. I pointed the `hidir` to a random build folder I had and was greeted by many errors. Ideally for me `ghcide` will write the interface files itself.

```
    .hie-bios/stage0/compiler/build/Config.hs
Hidden:   no
Range:    1:0-100001:0
Source:   compiler
Severity: DsError
[...]

PR #412, opened 2020-02-08
Workspace roots and getFileExists

pepeiborra:
This is a correctness fix for `getFileExists` when targetting files out of the workspace - they cannot be watched by the client so we need to fall back to the slow path.

This is a spinout from the effort to switch to interface files (#355).

### DIFF
--- a/src/Development/IDE/Core/IdeConfiguration.hs
+++ b/src/Development/IDE/Core/IdeConfiguration.hs
@@ -1,0 +1,66 @@
+{-# LANGUAGE DuplicateRecordFields #-}
+module Development.IDE.Core.IdeConfiguration
+  ( IdeConfiguration(..)
+  , registerIdeConfiguration
+  , parseConfiguration
+  , parseWorkspaceFolder
+  , isWorkspaceFile
+  , modifyWorkspaceFolders
+  )
+where
+
+import           Control.Concurrent.Extra
+import           Control.Monad
+import           Data.HashSet                   (HashSet, singleton)
+import           Data.Text                      (Text, isPrefixOf)
+import           Development.IDE.Core.Shake
+import           Development.IDE.Types.Location
+import           Development.Shake
+import           Language.Haskell.LSP.Types
+
+-- | Lsp client relevant configuration details
+data IdeConfiguration = IdeConfiguration
+  { workspaceFolders :: HashSet NormalizedUri
+  }
+  deriving (Show)
+
+data IdeConfigurationVar = IdeConfigurationVar {unIdeConfigurationRef :: Var IdeConfiguration}
+
+instance IsIdeGlobal IdeConfigurationVar
+
+registerIdeConfiguration :: ShakeExtras -> IdeConfiguration -> IO ()
+registerIdeConfiguration extras =
+  addIdeGlobalExtras extras . IdeConfigurationVar <=< newVar
+
+getIdeConfiguration :: Action IdeConfiguration
+getIdeConfiguration =
+  getIdeGlobalAction >>= liftIO . readVar . unIdeConfigurationRef
+
+parseConfiguration :: InitializeRequest -> IdeConfiguration
+parseConfiguration RequestMessage { _params = InitializeParams {..} } =
+  IdeConfiguration { .. }
+ where
+  workspaceFolders =
+    foldMap (singleton . toNormalizedUri) _rootUri
+      <> (foldMap . foldMap)
+           (singleton . parseWorkspaceFolder)
+           _workspaceFolders
+
+parseWorkspaceFolder :: WorkspaceFolder -> NormalizedUri
+parseWorkspaceFolder =
+  toNormalizedUri . Uri . (_uri :: WorkspaceFolder -> Text)
+
+modifyWorkspaceFolders
+  :: IdeState -> (HashSet NormalizedUri -> HashSet NormalizedUri) -> IO ()
+modifyWorkspaceFolders ide f = do
+  IdeConfigurationVar var <- getIdeGlobalState ide
+  IdeConfiguration    ws  <- readVar var
+  writeVar var (IdeConfiguration (f ws))
+
+isWorkspaceFile :: NormalizedFilePath -> Action Bool
+isWorkspaceFile file = do
+  IdeConfiguration {..} <- getIdeConfiguration
+  let toText = getUri . fromNormalizedUri
+  return $ any
+    (\root -> toText root `isPrefixOf` toText (filePathToUri' file))
+    workspaceFolders
--- a/src/Development/IDE/Core/IdeConfiguration.hs
+++ b/src/Development/IDE/Core/IdeConfiguration.hs
@@ -24,7 +24,7 @@ data IdeConfiguration = IdeConfiguration
   }
   deriving (Show)
 
-data IdeConfigurationVar = IdeConfigurationVar {unIdeConfigurationRef :: Var IdeConfiguration}
+newtype IdeConfigurationVar = IdeConfigurationVar {unIdeConfigurationRef :: Var IdeConfiguration}
 
 instance IsIdeGlobal IdeConfigurationVar
 
--- a/src/Development/IDE/LSP/Notifications.hs
+++ b/src/Development/IDE/LSP/Notifications.hs
@@ -72,7 +72,9 @@ setHandlersNotifications = PartialHandlers $ \WithMessage{..} x -> return x
             setSomethingModified ide
     ,LSP.didChangeWorkspaceFoldersNotificationHandler = withNotification (LSP.didChangeWorkspaceFoldersNotificationHandler x) $
         \_ ide (DidChangeWorkspaceFoldersParams events) -> do
+            let add       = S.union
+                substract = flip S.difference
             modifyWorkspaceFolders ide
-              $ S.union      (foldMap (S.singleton . parseWorkspaceFolder) (_added   events))
-              . S.difference (foldMap (S.singleton . parseWorkspaceFolder) (_removed events))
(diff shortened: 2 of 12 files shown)


src/Development/IDE/Core/IdeConfiguration.hs
```diff
@@ -0,0 +1,66 @@
...
+  IdeConfigurationVar var <- getIdeGlobalState ide
+  IdeConfiguration    ws  <- readVar var
+  writeVar var (IdeConfiguration (f ws))
+
+isWorkspaceFile :: NormalizedFilePath -> Action Bool
+isWorkspaceFile file = do
+  IdeConfiguration {..} <- getIdeConfiguration
+  let toText = getUri . fromNormalizedUri
+  return $ any
+    (\root -> toText root `isPrefixOf` toText (filePathToUri' file))
```
pepeiborra: Comparing URIs textually is a bit hacky, but I could not think of any better way.

test/exe/Main.hs
```diff
@@ -110,8 +110,6 @@ initializeResponseTests = withResource acquire release tests where
                               , _willSaveWaitUntil = Nothing
                               , _save = Just (SaveOptions {_includeText = Nothing})}))
 
-      nothingWorkspace = Just (WorkspaceOptions {_workspaceFolders = Nothing})
```
cocreature: Can we get a test for this?

src/Development/IDE/LSP/Notifications.hs
```diff
@@ -69,4 +71,9 @@ setHandlersNotifications = PartialHandlers $ \WithMessage{..} x -> return x
             logInfo (ideLogger ide) $ "Files created or deleted: " <> msg
             modifyFileExists ide events
             setSomethingModified ide
+    ,LSP.didChangeWorkspaceFoldersNotificationHandler = withNotification (LSP.didChangeWorkspaceFoldersNotificationHandler x) $
+        \_ ide (DidChangeWorkspaceFoldersParams events) -> do
+            modifyWorkspaceFolders ide
+              $ HS.union      (foldMap (HS.singleton . parseWorkspaceFolder) (_added   events))
```
cocreature:
```suggestion
              $ HS.union (HS.fromList (map parseWorkspaceFolder $ _added events)
```
I find that a bit easier to read and it avoids the (admittedly irrelevant) space leak of `foldMap` using a `foldr` instead of a `foldl'`.

src/Development/IDE/LSP/Notifications.hs
```diff
@@ -69,4 +71,9 @@ setHandlersNotifications = PartialHandlers $ \WithMessage{..} x -> return x
...
+              . HS.difference (foldMap (HS.singleton . parseWorkspaceFolder) (_removed events))
```
cocreature:
```suggestion
              . HS.difference (HS.fromList (map parseWorkspaceFolder $ _removed events))
```

cocreature: Is the `difference` here the right way around? `difference a b` is `a \\ b` whereas I think you want `b \\ a`.

src/Development/IDE/Core/IdeConfiguration.hs
```diff
@@ -0,0 +1,66 @@
...
+getIdeConfiguration :: Action IdeConfiguration
+getIdeConfiguration =
+  getIdeGlobalAction >>= liftIO . readVar . unIdeConfigurationRef
+
+parseConfiguration :: InitializeRequest -> IdeConfiguration
+parseConfiguration RequestMessage { _params = InitializeParams {..} } =
+  IdeConfiguration { .. }
+ where
+  workspaceFolders =
+    foldMap (singleton . toNormalizedUri) _rootUri
```
cocreature:
```suggestion
    HS.fromList [toNormalizedUri folder | folder <- toList _rootUri <> map parseWorkspaceFolder _workspaceFolder]
```

src/Development/IDE/Core/IdeConfiguration.hs
```diff
@@ -0,0 +1,66 @@
...
+      <> (foldMap . foldMap)
+           (singleton . parseWorkspaceFolder)
+           _workspaceFolders
+
+parseWorkspaceFolder :: WorkspaceFolder -> NormalizedUri
```
cocreature:
```suggestion
parseWorkspaceFolder :: WorkspaceFolder -> Uri
```

src/Development/IDE/Core/IdeConfiguration.hs
```diff
@@ -0,0 +1,66 @@
...
+parseWorkspaceFolder =
+  toNormalizedUri . Uri . (_uri :: WorkspaceFolder -> Text)
```
cocreature:
```suggestion
   Uri . (_uri :: WorkspaceFolder -> Text)
```

src/Development/IDE/LSP/Notifications.hs
pepeiborra: Absolutely. Good catch, thanks!

src/Development/IDE/Core/IdeConfiguration.hs
pepeiborra: This one really reads much worse, so I ended up reverting it.

src/Development/IDE/LSP/Notifications.hs
pepeiborra: After adding code to unpack the `List` the suggested versions were much noisier, so I ended up reverting back to the original versions

src/Development/IDE/Core/IdeConfiguration.hs
pepeiborra: This code is now comparing `Uri`s instead of `NormalizedUri`s following the suggested changes. Hopefully that's not making things any worse?

test/exe/Main.hs
pepeiborra: I can add a test that no `didChangeWatchedFile` subscriptions are made for files outside the workspace. Is that what you had in mind?

pepeiborra:
I'll look at adding some tests tomorrow

src/Development/IDE/LSP/Notifications.hs
pepeiborra: Wouldn't it be fun if ghcide would let us peek into the Core generated for an expression?

test/exe/Main.hs
cocreature: I was thinking of something similar to the existing test for `DidFileExists` that you added but with the file being outside of the workspace. Probably needs a direct-style cradle in a `hie.yaml` that has a `-i` flag pointing outside of the workspace.

pepeiborra:
> I'll look at adding some tests tomorrow

Tests added!

test/exe/Main.hs
```diff
@@ -415,6 +414,26 @@ codeLensesTests = testGroup "code lenses"
...
 
+watchedFilesTests :: TestTree
+watchedFilesTests = testGroup "watched files"
+  [ testSession "workspace file" $ do
+      _ <- openDoc' "A.hs" "haskell" "module A where"
+      RequestMessage{_params = RegistrationParams (List regs)} <- skipManyTill anyMessage (message @RegisterCapabilityRequest)
+      let watchedFileRegs =
+            [ args | Registration _id WorkspaceDidChangeWatchedFiles args <- regs ]
+      liftIO $ assertBool "watches workspace files" $ not $ null watchedFileRegs
+  , testSession "non workspace file" $ do
```
cocreature: I don’t understand this test. `openDoc'` always sends the `DidChangeWatchedFiles` notification so how does this work?
Even assuming this works, I am not sure this is the right thing to test. What we want to test is that the logic for files outside of the workspace works correctly and still picks up new files being created. So how about we change `openDoc'` to not send the notification for files outside the workspace and we add a variant of https://github.com/digital-asset/ghcide/blob/2d71599faf5e93d4ee848043ddd68f10dd295260/test/exe/Main.hs#L236 that creates the file outside of the workspace?

src/Development/IDE/Core/IdeConfiguration.hs
```diff
@@ -0,0 +1,65 @@
...
+modifyWorkspaceFolders ide f = do
+  IdeConfigurationVar var <- getIdeGlobalState ide
+  IdeConfiguration    ws  <- readVar var
+  writeVar var (IdeConfiguration (f ws))
+
+isWorkspaceFile :: NormalizedFilePath -> Action Bool
+isWorkspaceFile file = do
+  IdeConfiguration {..} <- getIdeConfiguration
+  return $ any
+    (\root -> getUri root `isPrefixOf` getUri (fromNormalizedUri $ filePathToUri' file))
```
cocreature: Ah good point, we definitely need to normalize `root` here or probably better: store it normalized in `IdeConfiguration`.

test/exe/Main.hs
pepeiborra: The test is not checking what `openDoc'` does, it is checking what `ghcide` does: it should never subscribe to changes in non workspace files. 
I'm happy to add an additional test to check that files outside the workspace get picked up.

src/Development/IDE/Core/IdeConfiguration.hs
pepeiborra: Indeed, that's what the first version of the PR did

alanz:
What is holding this up? I have a PR (#416) that needs to be updated once this lands, so we can select formatters in hls.

cocreature:
@alanz there were changes less than a day ago give us some time please :slightly_smiling_face: Merged now.

alanz:
@cocreature Thanks.  I saw it got updated yesterday, but figured I would give it a kick for time zone differences, so if it was easy to merge it would be done for me to work on it tonight.

I was not intending to harras anyone.

PR #413, opened 2020-02-09
Change getDocumentation to work with parsed modules

pepeiborra:
Just a simple refactoring to scope down the inputs in `getDocumentation` to what is actually being used. This is a spinout from the effort to switch to interface files (#355)

### DIFF
--- a/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -198,7 +198,7 @@ mkPragmaCompl label insertText =
     Nothing Nothing Nothing Nothing Nothing (Just insertText) (Just Snippet)
     Nothing Nothing Nothing Nothing Nothing
 
-cacheDataProducer :: HscEnv -> DynFlags -> TypecheckedModule -> [TypecheckedModule] -> IO CachedCompletions
+cacheDataProducer :: HscEnv -> DynFlags -> TypecheckedModule -> [ParsedModule] -> IO CachedCompletions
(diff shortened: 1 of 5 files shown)


src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -198,7 +198,7 @@ mkPragmaCompl label insertText =
     Nothing Nothing Nothing Nothing Nothing (Just insertText) (Just Snippet)
     Nothing Nothing Nothing Nothing Nothing
 
-cacheDataProducer :: HscEnv -> DynFlags -> TypecheckedModule -> [TypecheckedModule] -> IO CachedCompletions
+cacheDataProducer :: HscEnv -> DynFlags -> TypecheckedModule -> [ParsedModule] -> IO CachedCompletions
 cacheDataProducer packageState dflags tm tcs = do
```
aherrmann-da:
```suggestion
cacheDataProducer packageState dflags tm pmods = do
```
IIUC `tcs` was short for `typeCheckedModules`. In GHC's sources I see `pmod` as a short-cut for `parsedModule` in a couple places. There are a couple more instances of now misnamed `tcs`, `tms`, or `tcms` in this PR.

pepeiborra: I have taken a shot at fixing the names, but I'm not sure if it's what you had in mind. Rather than shortened type names, I've gone with more "semantic" names. Hope it works!

aherrmann-da: Thanks, that's even better!

PR #414, opened 2020-02-09
Fix ModLocation creation

pepeiborra:
This is a correctness fix to consider the dflags configuration.
It a spinout from the effort to switch to interface files (#355)

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -39,6 +39,7 @@ import qualified Parser
 import           Lexer
 import ErrUtils
 
+import           Finder
 import qualified GHC
 import           GhcMonad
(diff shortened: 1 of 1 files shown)


src/Development/IDE/Core/Compile.hs
```diff
@@ -342,7 +333,7 @@ getModSummaryFromBuffer fp contents dflags parsed = do
     , ms_parsed_mod   = Nothing
     }
     where
-      (sourceType, derivedFile) =
+      (sourceType, _derivedFile) =
```
aherrmann-da: This can be simplified further by dropping the `\newExt` stuff and using `takeExtension`. Any reason not to?

pepeiborra: Good point, thanks

PR #415, opened 2020-02-09
Change the representation of filestamps

pepeiborra:
This is needed to be able to compare them, which we need for loading `.hie` and `.hi` files.
Another spinout from the effort to switch to interface files (#355)

### DIFF
--- a/src/Development/IDE/Core/FileStore.hs
+++ b/src/Development/IDE/Core/FileStore.hs
@@ -95,11 +95,12 @@ getModificationTimeRule :: VFSHandle -> Rules ()
 getModificationTimeRule vfs =
     defineEarlyCutoff $ \GetModificationTime file -> do
         let file' = fromNormalizedFilePath file
-        let wrap time = (Just time, ([], Just $ ModificationTime time))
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Core/FileStore.hs
```diff
@@ -115,19 +116,21 @@ getModificationTimeRule vfs =
...
     -- support them, as done for GetFileExists
-    getModTime :: FilePath -> IO BS.ByteString
+    getModTime :: FilePath -> IO (Int,Int)
     getModTime f =
 #ifdef mingw32_HOST_OS
         do time <- Dir.getModificationTime f
-           pure $! BS.pack $ show (toModifiedJulianDay $ utctDay time, diffTimeToPicoseconds $ utctDayTime time)
+           let !day = toModifiedJulianDay $ utctDay time
+               !dayTime = diffTimeToPicoseconds $ utctDayTime time
+           pure $! (day, dayTime)
```
ndmitchell: The `$!` on something whose RHS is in WHNF is redundant.
```suggestion
           pure (day, dayTime)
```

PR #416, opened 2020-02-09
Parameterize the haskell-lsp client config type

alanz:
So that haskell-language-server can use its own config


ndmitchell:
Makes a lot of sense to me.

alanz:
> I somewhat wonder why we even have the type parameter for the configuration in haskell-lsp? 

It seemed to make sense at the time, and is a convenience for off the shelf users of the library. The config change message is part of the spec, after all.

But perhaps the overall structure could be reworked, now we know a lot more about the protocol.

PR #431, opened 2020-02-13
Track module dependencies

pepeiborra:
Another spinout from the effort to switch to interface files (#355), joint work with @lazamar 
This is needed to reason about dependencies at the module level, in particular to tell whether a dependency is a home module or a package module

### DIFF
--- a/src/Development/IDE/Import/DependencyInformation.hs
+++ b/src/Development/IDE/Import/DependencyInformation.hs
@@ -17,7 +17,7 @@ module Development.IDE.Import.DependencyInformation
   , pathToId
   , idToPath
   , reachableModules
-
+  , modLocationToNormalizedFilePath
   , processDependencyInformation
   , transitiveDeps
   ) where
@@ -46,6 +46,7 @@ import GHC.Generics (Generic)
 
 import Development.IDE.Types.Diagnostics
 import Development.IDE.Types.Location
+import Development.IDE.Import.FindImports (ArtifactsLocation(..))
 
 import GHC
 import Module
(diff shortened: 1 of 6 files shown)


src/Development/IDE/Import/DependencyInformation.hs
```diff
@@ -268,22 +295,49 @@ transitiveDeps DependencyInformation{..} file = do
...
+  nmdModLocation :: !ModLocation
+  }
+  deriving Generic
+
+instance Eq NamedModuleDep where
+  a == b = nmdFilePath a == nmdFilePath b
+
+instance NFData NamedModuleDep where
+  rnf NamedModuleDep{..} =
+    rnf nmdFilePath `seq` rnf nmdModuleName `seq` rwhnf nmdModLocation
```
aherrmann-da: Can you add a comment why it's `rwhnf` of `nmdModLocation`?

src/Development/IDE/Import/FindImports.hs
```diff
@@ -29,10 +30,16 @@ import           Control.Monad.IO.Class
 import           System.FilePath
 
 data Import
-  = FileImport !NormalizedFilePath
+  = FileImport !ArtifactsLocation
   | PackageImport !M.InstalledUnitId
   deriving (Show)
 
+data ArtifactsLocation =  ArtifactsLocation !ModLocation
```
aherrmann-da: Why isn't this a `newtype`?

src/Development/IDE/Import/DependencyInformation.hs
```diff
@@ -67,27 +69,33 @@ newtype FilePathId = FilePathId { getFilePathId :: Int }
...
+  { idToPathMap :: !(IntMap ArtifactsLocation)
   , pathToIdMap :: !(HashMap NormalizedFilePath FilePathId)
   }
   deriving (Show, Generic)
 
 instance NFData PathIdMap
 
+modLocationToNormalizedFilePath :: ArtifactsLocation -> NormalizedFilePath
+modLocationToNormalizedFilePath (ArtifactsLocation loc) =
+    let (Just filePath) = ml_hs_file loc
```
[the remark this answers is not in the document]
pepeiborra: I think we do know because we create the `ModLocation` using `mkHomeModLocation`. I'll add a comment explaining this

src/Development/IDE/Import/FindImports.hs
pepeiborra: Fixed

src/Development/IDE/Import/DependencyInformation.hs
pepeiborra: Added (ModLocation lacks an NFData instance)

PR #433, opened 2020-02-15
Fix corner cases in remove redundant imports code action

### LINKED ISSUE #352: Remove import quick fix is the identity change
There are at least two situations in which this quick fix has no effect:

1. For symbolic operators (eg `<?>`) when they are the head of the import list 
2. For constructors imported inside of their datatype eg `Maybe(Just)`

The outcome of this should be a fix for those issues. It might also be a good idea to check that the fix is not empty before sending it to the lsp client.

pepeiborra:
- Redundant "All" imports, e.g. Maybe(..)
 - Redundant datatype plus constructors, e.g. Maybe(Just)

Fixes #352

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -138,6 +138,7 @@ library
         Development.IDE.Spans.Calculate
         Development.IDE.Spans.Documentation
         Development.IDE.Spans.Type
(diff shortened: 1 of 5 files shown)


pepeiborra:
Done

PR #435, opened 2020-02-16
Add a .editorconfig

ndmitchell:
Only thing perhaps controversial here is 4 spaces for an indent, as much of this code uses 2 spaces (👋  at @DavidM-D). But as a project it's good to pick one level of indent to aim at.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,5 @@
 ; This file is for unifying the coding style for different editors and IDEs.
-; More information at http://EditorConfig.org
+; More information at https://EditorConfig.org
 
 root = true
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at http://EditorConfig.org
+
+root = true
+
+[*]
+end_of_line = LF
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true


jneira:
I would add a warning for windows users due to the LF option: they should disable `auto.clrf` in git,  take care of not copy external files with CR/LF and change its editor configuration to take `.editorconfig` if it is supported (or change it manually if not)
I think i had to install a vscode extension to make it honour `.editorconfig` and for other editors a plugin is defintely needed (following editorconfig.org)

ndmitchell:
Where would we put the warning so Windows users would see it? In my view, the auto.clrf thing should never be turned on by any Windows users - it's like the auto crlf thing in tar and WinZip - a constant source of pain for the benefit of those users who want to edit text files in a version of Notepad on Windows ME. `</rant>`

jneira:
Well, i think that, by default, almost all windows editors use `cr/lf` for new files, although they don't change them if you uses an existing one, right?
So you have to ensure your editor uses by default `lf` (convince it to use `.editorconfig` or change its configuration globally). In my experience users does not change the default, by default :wink: so a warning could be useful.
Otoh if you move a file from other location with `cr/lf` without editing it you'll leak it in the repo too.

I agree with you on git `auto.crlf` but the fact is `git for windows` suggest you enable it and quite users will have it (myself!) so add a note in the `contributing` section could help too.

ndmitchell:
Added in https://github.com/digital-asset/ghcide/pull/463.

PR #437, opened 2020-02-16
Code action to suggest adding missing imports from pkg db

pepeiborra:
The implementation looks in modules loaded from the package database. It should
only look in packages declared as dependencies of the project. The package
modules are loaded lazily and are global to the HscEnv, so the success rate will
depend on what has been loaded so far in the env.

I intended to make this as a standalone plugin but we cannot have two plugins providing code actions, see https://github.com/haskell/haskell-language-server/issues/25#issuecomment-586704493

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -358,6 +358,9 @@ suggestNewImport eps ParsedModule {pm_parsed_source = L _ HsModule {..}} Diagnos
           RealSrcLoc s -> Just $ srcLocLine s
           _ -> Nothing
   , insertPos <- Position insertLine 0
+  , extendImportSuggestions <- -- Just [binding, mod, srcspan] <-
+    matchRegex _message
+    "Perhaps you want to add ‘[^’]*’ to the import list in the import of ‘([^’]*)’"
   =
     nubOrdBy
       (compare `on` fst)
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -333,6 +347,37 @@ suggestSignature isQuickFix Diagnostic{_range=_range@Range{..},..}
               | otherwise                  = nm <> ty
 suggestSignature _ _ = []
 
+suggestNewImport :: ExternalPackageState -> ParsedModule -> Diagnostic -> [(T.Text, [TextEdit])]
+suggestNewImport eps ParsedModule {pm_parsed_source = L _ HsModule {..}} Diagnostic{_message}
+  | Just [name, _typ] <- matchRegex (unifySpaces _message) "Variable not in scope: ([^ ]+) :: ([^*•]+)"
```
aherrmann-da: GHC's error message does not always contain the type. E.g. in
```
foo = fromJust
```
The error reads
```
Variable not in scope: fromJust
```
This suggestion doesn't use the type anyway. Could you change the regex to make the type annotation optional?

aherrmann-da: Just to point it out, this doesn't suggest qualified imports. E.g.
```
foo = Maybe.fromJust
```
Produces the error message
```
Not in scope: ‘Maybe.fromJust’
No module named ‘Maybe’ is imported.
```

Supporting qualified imports could be a nice extension of this feature in the future.

pepeiborra: Agreed! I suspect that there are many extensions to this feature, and I wouldn't be able to tackle them all in this PR.

PR #440, opened 2020-02-17
Add .hie files support for home modules

pepeiborra:
`.hie` files are used instead of typecheck artifacts for `getDefinition` functionality when using interface files, as they lack located typechecked ASTs.

We choose to never store .hie files in the Shake graph, as they are
- expensive in space
- quick to load
- only used for go to definition

While there, we remove package module .hie files from the Shake graph too.

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -29,7 +29,7 @@ import Fingerprint
 
 import Data.Binary
 import Data.Bifunctor (second)
-import Control.Monad
+import Control.Monad.Extra
 import Control.Monad.Trans.Class
 import Control.Monad.Trans.Maybe
 import Development.IDE.Core.Compile
@@ -50,6 +50,7 @@ import           Data.Foldable
 import qualified Data.IntMap.Strict as IntMap
 import qualified Data.IntSet as IntSet
 import Data.List
+import Data.Ord
 import qualified Data.Set                                 as Set
 import qualified Data.Text                                as T
 import           Development.IDE.GHC.Error
@@ -58,8 +59,6 @@ import Development.IDE.Core.RuleTypes
 import Development.IDE.Spans.Type
 
 import qualified GHC.LanguageExtensions as LangExt
-import           UniqSupply
-import NameCache
 import HscTypes
 import DynFlags (xopt)
 import GHC.Generics(Generic)
@@ -68,6 +67,7 @@ import qualified Development.IDE.Spans.AtPoint as AtPoint
 import Development.IDE.Core.Service
 import Development.IDE.Core.Shake
 import Development.Shake.Classes
+import qualified System.Directory as IO
 
 -- | This is useful for rules to convert rules that can only produce errors or
 -- a result into the more general IdeResult type that supports producing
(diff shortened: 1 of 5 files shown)


mpickering:
Is this deliberately missing my patch to support the hie file code paths on older GHCs?

pepeiborra:
> Is this deliberately missing my patch to support the hie file code paths on older GHCs?

The reason your patch is missing from this PR is that it did not rebase cleanly. Rest assured, I will include it in a later PR with the rest of interface file changes.

src/Development/IDE/Core/Rules.hs
```diff
@@ -112,9 +112,59 @@ getDefinition :: NormalizedFilePath -> Position -> Action (Maybe Location)
...
+  pm <- use_ GetParsedModule f
+  let normal_hie_f = toNormalizedFilePath hie_f
+      hie_f = ml_hie_file $ ms_location $ pm_mod_summary pm
+  mbHieTimestamp <- use GetModificationTime normal_hie_f
+  srcTimestamp   <- use_ GetModificationTime f
+  case (mbHieTimestamp, srcTimestamp) of
+    (Just hie, src)
+      | comparing modificationTime hie src == GT -> pure ()
+    _ -> do
+      -- typecheck generates a .hie file as a side effect
```
cocreature: This comment is not accurate atm is it?

src/Development/IDE/Core/Rules.hs
```diff
@@ -112,9 +112,59 @@ getDefinition :: NormalizedFilePath -> Position -> Action (Maybe Location)
...
+
+
+getHomeHieFile :: NormalizedFilePath -> Action ([a], Maybe HieFile)
+getHomeHieFile f = do
+  pm <- use_ GetParsedModule f
+  let normal_hie_f = toNormalizedFilePath hie_f
+      hie_f = ml_hie_file $ ms_location $ pm_mod_summary pm
+  mbHieTimestamp <- use GetModificationTime normal_hie_f
+  srcTimestamp   <- use_ GetModificationTime f
+  case (mbHieTimestamp, srcTimestamp) of
```
cocreature: Might be easier to read with something like 

```
when (maybe True (\hieTimeStamp -> hieTimeStamp <= srcTimeStamp) mbHieTimestamp $ 
  use_ TypeCheck f`

File: src/Development/IDE/Core/Rules.hs [hunk 3]
```diff
@@ -112,9 +112,59 @@ getDefinition :: NormalizedFilePath -> Position -> Action (Maybe Location)
...
+    let unitId = moduleUnitId mod
+    case lookupPackageConfig unitId pkgState of
+        Just pkgConfig -> do
+            hieFile <- liftIO $ optLocateHieFile optPkgLocationOpts pkgConfig mod
+            path    <- liftIO $ optLocateSrcFile optPkgLocationOpts pkgConfig mod
+            case (hieFile, path) of
+                (Just hiePath, Just modPath) -> liftIO $ do
+                    -- deliberately loaded outside the Shake graph
+                    -- to avoid dependencies on non-workspace files
+                    foundIt <- IO.doesFileExist hiePath
```
cocreature: The intention of `optLocateHieFile` was to return `Nothing` if the file does not exist, at least that’s how we use it in DAML.

src/Development/IDE/Core/Rules.hs
```diff
@@ -112,9 +112,59 @@ getDefinition :: NormalizedFilePath -> Position -> Action (Maybe Location)
...
-    lift $ AtPoint.gotoDefinition getHieFile opts pkgState (spansExprs spans) pos
+    lift $ AtPoint.gotoDefinition (getHieFile file) opts (spansExprs spans) pos
+
+getHieFile
+  :: NormalizedFilePath -- ^ file we're editing
+  -> Module -- ^ module dep we want info for
+  -> Action (Maybe (HieFile, FilePath)) -- ^ hie stuff for the module
+getHieFile file mod = do
+  TransitiveDependencies {transitiveNamedModuleDeps} <- use_ GetDependencies file
+  case find (\x -> nmdModuleName x == moduleName mod) transitiveNamedModuleDeps of
```
cocreature: The linear search here seems a bit worrying. Should we turn `transitiveNamedModuleDeps` into somtehing like `Map NormalizedFilePath …`?

src/Development/IDE/Core/Rules.hs
pepeiborra: oops. Too much rebasing I guess!

src/Development/IDE/Core/Rules.hs
pepeiborra: Right, I'll fix and add a comment to make it explicit

src/Development/IDE/Core/Rules.hs
pepeiborra: It is O(n) over a small number of n, and it's not in a critical path either, so I don't think it's that worrying.

src/Development/IDE/Core/Rules.hs
pepeiborra: improved

src/Development/IDE/Core/Rules.hs
cocreature: For a decently sized project small can easily mean > 1000 which doesn’t seem that small. But yeah it’s probably fine for now.

pepeiborra: 1000 is a tiny number!

pepeiborra:
I'll look at adding a test tonight

pepeiborra:
Actually, I think that adding a test to this PR is pretty much impossible, since `gotoDefinition` will never request a `.hie` file for a local package. So I suggest leaving the test for the interfaces PR

PR #441, opened 2020-02-17
Local hidir and hiedir folders to avoid conflicts with Cabal

pepeiborra:
`hi` files created by ghcide would have different optimization settings,
triggering Cabal recompilation if they were stored in Cabal folders

`hie` files would end up in the src folder as `-hiedir` is not set by Cabal

Telling GHC to write interface/hie files is pointless since we (will) do it ourselves,
and it also leads to the recompilation checker getting confused

Using hie-bios `initSession` does things like setting up the cache directory for
interface files and other things which ghcide would rather manage itself.


ghcide.cabal
```diff
@@ -170,7 +170,10 @@ executable ghcide
         hslogger,
         base == 4.*,
         binary,
+        base16-bytestring >=0.1.1 && <0.2,
+        bytestring,
         containers,
+        cryptohash-sha1 >=0.11.100 && <0.12,
```
cocreature: not that it really matters here but `sha1` seems like an odd choice. Why not `sha256`?

exe/Main.hs
```diff
@@ -224,14 +240,39 @@ getComponentOptions cradle = do
...
+             setDefaultHieDir cacheDir $
+             setIgnoreInterfacePragmas $
+             disableOptimisation dflags'
         getSession
     initDynLinker env
     newHscEnvEq env
 
+setIgnoreInterfacePragmas :: DynFlags -> DynFlags
+setIgnoreInterfacePragmas df =
+    gopt_set (gopt_set df Opt_IgnoreInterfacePragmas) Opt_IgnoreOptimChanges
```
cocreature: Why do we need `IgnoreOptimChanges` if we set the opt level to 0?

test/exe/Main.hs
```diff
@@ -229,7 +229,7 @@ diagnosticTests = testGroup "diagnostics"
       let change = TextDocumentContentChangeEvent
             { _range = Just (Range (Position 0 0) (Position 0 20))
             , _rangeLength = Nothing
-            , _text = ""
+            , _text = "module ModuleC where"
```
cocreature: Why the change here?

test/exe/Main.hs
```diff
@@ -927,7 +927,9 @@ insertNewDefinitionTests :: TestTree
 insertNewDefinitionTests = testGroup "insert new definition actions"
   [ testSession "insert new function definition" $ do
       let txtB =
-            ["foo True = select [True]"
+            ["module M where"
```
cocreature: Why the change here?

test/exe/Main.hs
```diff
@@ -1557,7 +1561,7 @@ completionTests
 outlineTests :: TestTree
 outlineTests = testGroup
   "outline"
-  [ testSessionWait "type class" $ do
+  [ testSession "type class" $ do
```
cocreature: Reminder to myself and other reviewers: This is a reasonable change since we don’t call `getDiagnostics`.

ghcide.cabal
pepeiborra: No real reason, I just reused what `hie-bios` was doing to hash the flags

exe/Main.hs
pepeiborra: We don't _need_ to. I can remove it if needed

test/exe/Main.hs
pepeiborra: these tests started failing after the changes, with the error `The IO action \8216main\8217 is not defined in module \8216Main\8217"`. Not sure why

cocreature: The one change I see compared to `hie-bios` is the linker flags in https://hackage.haskell.org/package/hie-bios-0.4.0/docs/src/HIE.Bios.Environment.html#setLinkerOptions which looks like it could be the culprit here. Did you explicitly leave them out?

pepeiborra: Doh, I must have missed them. Well spotted Moritz, thank you for your careful review. Breaking down the interfaces branch into smaller PRs is really paying off.

cocreature: Thanks a lot for breaking things off into smaller PRs! I know it takes a lot of effort but it makes reviews way easier.

PR #443, opened 2020-02-17
Group Imports in Outline into a single node

### LINKED ISSUE #438: Outline view contains too many imports
The outline view lists every import my module makes (which are almost entirely uninteresting), followed by the functions if I scroll enough (which has lots of great information). Would be better if the imports were either collapsed by prefix (e.g. Data.* under one tree view), or probably just replaced with the word "imports", or even omitted entirely.

I'm not actually sure if it's Ghcide which is populating this or the Haskell syntax language thing. I hope @shayne-fletcher both knows and knows how to fix it :). If it is the syntax highlighting I suggest the Haskell IDE crew reach out and engage with them about it, and treat that part as part of the whole IDE ecosystem.

image

fendor:
Makes it easier to see symbols in scope when you have lots of
import statements.

Bikeshedding:
Should a single imports still be wrapped into a DocumentSymbol?

In manual tests, behaviour is as I would have expected it.
@ocharles Can you test if your work-flow still works?

closes #438

### DIFF
--- a/src/Development/IDE/LSP/Outline.hs
+++ b/src/Development/IDE/LSP/Outline.hs
@@ -50,7 +50,9 @@ moduleOutline _lsp ideState DocumentSymbolParams { _textDocument = TextDocumentI
                    , _kind  = SkFile
                    , _range = Range (Position 0 0) (Position maxBound 0) -- _ltop is 0 0 0 0
                    }
-               importSymbols = documentSymbolForImportSummary (mapMaybe documentSymbolForImport hsmodImports)
+               importSymbols = maybe [] pure $
+                  documentSymbolForImportSummary
+                    (mapMaybe documentSymbolForImport hsmodImports)
                allSymbols    = case moduleSymbol of
                  Nothing -> importSymbols <> declSymbols
                  Just x ->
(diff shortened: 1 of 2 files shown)


src/Development/IDE/LSP/Outline.hs
```diff
@@ -167,6 +167,28 @@ documentSymbolForDecl (L l (ForD x)) = Just
...
+documentSymbolForImportSummary :: [DocumentSymbol] -> [DocumentSymbol]
+documentSymbolForImportSummary [] = []
+documentSymbolForImportSummary importSymbols = 
+    let 
+      importRange = foldr (\DocumentSymbol {..} (Range accStart accEnd) -> 
+        let Range start end = _range 
+        in  Range (min start accStart) (max end accEnd)
+        ) 
+        (Range (Position maxBound 0) (Position 0 0))
+        importSymbols
```
fendor: I dislike this code. Is there something nicer for it?

ndmitchell:
```
-- safe because if we have no ranges then we don't take this branch
mergeRanges xs = Range (minimum $ map _start xs) (maximum $ map _end xs)
importRange = mergeRanges $ map _range importSymbols
```

src/Development/IDE/LSP/Outline.hs
```diff
@@ -167,6 +167,25 @@ documentSymbolForDecl (L l (ForD x)) = Just
...
+-- | Wrap the Document imports into a hierarchical outline for
+-- a better overview of symbols in scope.
+-- If there are no imports, then no hierarchy will be created.
+documentSymbolForImportSummary :: [DocumentSymbol] -> [DocumentSymbol]
+documentSymbolForImportSummary [] = []
+documentSymbolForImportSummary importSymbols =
+    let
+      -- safe because if we have no ranges then we don't take this branch
+      mergeRanges xs = Range (minimum $ map _start xs) (maximum $ map _end xs)
+      importRange = mergeRanges $ map (_range :: DocumentSymbol -> Range) importSymbols
```
fendor: Type Annotation is needed

src/Development/IDE/LSP/Outline.hs
```diff
@@ -167,6 +167,25 @@ documentSymbolForDecl (L l (ForD x)) = Just
 
 documentSymbolForDecl _ = Nothing
 
+-- | Wrap the Document imports into a hierarchical outline for
+-- a better overview of symbols in scope.
+-- If there are no imports, then no hierarchy will be created.
+documentSymbolForImportSummary :: [DocumentSymbol] -> [DocumentSymbol]
```
cocreature: I think it might be clearer if you make this return a `Maybe DocumentSymbol` and then turn it into a list at the call site. The current type signature doesn’t tell you very much.

fendor: That is true. However, this requires some minor work at call-site and I thought it would be easier to extend, if we decided to change the behaviour to create import groups per prefix. Will change it, though.

Issue #297, opened 2019-12-30
Detect GHC version mismatches


simonmichael:
I built ghcide 0.5.0.0 with GHC 8.6. When I run it so, it fails thus:
```
~/src/hledger$ ghcide hledger-lib
ghcide version: 0.0.5 (GHC: 8.6.5)
Ghcide setup tester in /Users/simon/src/PLAINTEXTACCOUNTING/hledger.
Report bugs at https://github.com/digital-asset/ghcide/issues

Step 1/6: Finding files to test in /Users/simon/src/PLAINTEXTACCOUNTING/hledger
Found 57 files

Step 2/6: Looking for hie.yaml files that control setup
Found 1 cradle

Step 3/6, Cradle 1/1: Implicit cradle for /Users/simon/src/PLAINTEXTACCOUNTING/hledger
Cradle {cradleRootDir = "/Users/simon/src/PLAINTEXTACCOUNTING/hledger", cradleOptsProg = CradleAction: stack}

Step 4/6, Cradle 1/1: Loading GHC Session
ghcide:
lookupSymbol failed in relocateSection (relocate external)
/Users/simon/.stack/programs/x86_64-osx/ghc-8.8.1/lib/ghc-8.8.1/ghc-prim-0.5.3/HSghc-prim-0.5.3.o: unknown symbol `_stg_atomicModifyMutVar2zh'
ghcide: ghcide: unable to load package `ghc-prim-0.5.3'
```
I forget and I'm not finding it in the readme just now - is ghcide supposed to work across GHC versions ? Or do you have to use a ghcide compiled with the same GHC as your project ?

https://github.com/digital-asset/ghcide/issues/53 looks similar but the fix there doesn't seem to apply here.

simonmichael:
Now I see it: "It's important that ghcide is compiled with the same compiler you use to build your projects." Alas!

simonmichael:
I installed another ghcide binary, this time building with GHC 8.8. This solved the problem.

It would be great if the tool could detect and warn about this common mistake.

In this project there's another stack yaml file for GHC 8.6. I hoped this would make ghcide happy, but it didn't:
```
stack --stack-yaml stack-ghc8.6.yaml exec -- ghcide hledger-lib
```

aherrmann-da:
> In this project there's another stack yaml file for GHC 8.6. I hoped this would make ghcide happy, but it didn't [...]

[...]

PR #447, opened 2020-02-21
Build against lts-15.0 and GHC 8.8.2

afcowie:
Just for testing, I needed **ghcide** built with GHC 8.8.2, so I tried `lts-15.0`. Everything it needs to build is in the stable snapshot now, so the _stack.yaml_ simplified somewhat. Everything appears to be working.

Concurrently I decided to build **stack** from `'master'` in order to make the annoying warning about untested versions of compiler and Cabal go away, but that was cosmetic.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,17 +1,6 @@
-resolver: nightly-2019-09-21
+resolver: lts-15.0
 packages:
 - .
-extra-deps:
-- haskell-lsp-0.20.0.0
-- haskell-lsp-types-0.20.0.0
-- lsp-test-0.10.1.0
-- hie-bios-0.4.0
-- fuzzy-0.1.0.0
-- regex-pcre-builtin-0.95.1.1.8.43
-- regex-base-0.94.0.0
-- regex-tdfa-1.3.1.0
-- shake-0.18.5
-- parser-combinators-1.2.1
-- haddock-library-1.8.0
+extra-deps: []
 nix:
   packages: [zlib]


ndmitchell:
In my view LTS is actively harmful - using the most recent nightly would have the same result and make it easier to smoothly upgrade in future. http://neilmitchell.blogspot.com/2015/12/whats-point-of-stackage-lts.html?m=1

We currently have a stack88.yaml with no overrides. I guess the question is what should our default stack.yaml be?

cocreature:
Ì agree with Neil, I don’t think LTS is helping us in any way. And I definitely want to keep testing on GHC 8.6.5. Until GHC 8.8.2 starts being usable on Windows, I’d rather have `stack.yaml` point to something that works there, so I think for now GHC 8.6.5 is better.

afcowie:
{shrug} fair enough. Although you might be missing something - `nightly` becomes the next stable (ie sometime next week, each week). It's on GHC 8.8.2 right now. Today's `nightly` will be tomorrow's `lts-15.1`.

If you'd like, I can resubmit this for `lts-14.27`, the current stable (very solid imho) with GHC 8.6.5.

The other way to do this is having something like _stack.yaml-8.8_ and _stack.yaml-8.6_ or _stack.yaml-14_ and _stack.yaml-15_ in your project root which is moderately important because **ghcide** has to be built locally for the compiler you're using in your project, right? Having upgraded to stable `lts-15.0` this week I had to rebuild the _ghcide_ binary (c.f. https://github.com/digital-asset/ghcide/issues/297#issuecomment-569830763) and that's why I submitted this suggestion.

People will [end up having to] do this themselves, so it's just a question of whether you want to help show working combination(s) they might consider.

fwiw The moniker "lts" on Stackage's part was unfortunate. "stable" would have been better of them; "lts" of Ubuntu fame implies years old and years out of date; Stackage does a new "lts" each week or so, whenever `nightly` converges safely when building **acme-everything** and/or they cull out the misbehaving packages.

cocreature:
Having multiple stack.yaml files is exactly the approach we are using atm? Not quite sure what you are proposing.

ndmitchell:
That's not how LTS works - packages don't get major 0.1 bumps in an LTS so tomorrow's nightly doesn't become the next LTS. It's just excessively old code that upgrades in a Big Bang with each new .0 version, as opposed to the continually evolving nightly. I don't think anyone should ever use LTS.

cocreature:
Stackage LTS also plays very badly with a package ecosystem where the vast majority of package maintainers does not backport bugfixes or even worse security fixes (and I don’t blame maintainers for not doing that).

PR #452, opened 2020-02-23
Adjust PR #377 for GHC 8.10 (DynFlags/Settings optP)

sheaf:
Unfortunately PR #377, which was meant to help with GHC 8.10 compatibility, also requires a fix for GHC 8.10.

Recent versions of GHC's `DynFlags` module define an [`addOptP` function](https://gitlab.haskell.org/ghc/ghc/blob/ghc-8.10/compiler/main/DynFlags.hs#L2716), but it is unfortunately not exported, so the definition is copied here.

With this additional fix, the master branch of `ghcide` builds on GHC 8.10 rc1 on my end.

### DIFF
--- a/src/Development/IDE/Core/Preprocessor.hs
+++ b/src/Development/IDE/Core/Preprocessor.hs
@@ -166,18 +166,11 @@ runLhs dflags filename contents = withTempDir $ \dir -> do
     escape (c:cs)    = c : escape cs
     escape []        = []
 
-
-modifyOptP :: ([String] -> [String]) -> DynFlags -> DynFlags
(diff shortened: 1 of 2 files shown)


ndmitchell:
Could we quickly send a patch to GHC HQ to get addOptP exposed?

sheaf:
To be backported to the GHC 8.10 branch, you mean? Anything else that might need to be exposed?

ndmitchell:
Ideally backported if there is time. If not, added to HEAD, so that for 8.12 we can get rid of the code. Nothing else on our immediate list (we've been sending things upstream as they arise).

PR #457, opened 2020-02-24
Support for interface files

### LINKED ISSUE #355: Use interface files (*.hi) when available

pepeiborra:
This PR is joint work with @mpickering. There have been also contributions from @lazamar  and @adamse but those are already merged. 

Ghcide will now rely on interface files for type checking all modules, except for files of interest. Since interface files do not provide typed source spans, ghcide now relies on extended interface files (aka `.hie` files) for hover and getDefinition functionality. Even though only GHC >8.8 supports extended interface files, @mpickering has hand rolled support for GHC 8.6 (included in this PR).

Instead of reusing Cabal/Stack/User (extended) interface files, ghcide generates them on its own. This is to avoid recompilations due to different flag settings amongst other problems. Interface files are generated as a by-product of type checking, and Info diagnostics are show while interface files are unavailable. On an effort to prevent bloating the Shake graph, typechecking artifacts are only recorded in the graph for files of interest and not for dependencies.

Template Haskell does negate some or most of the benefits, @mpickering  has opened #422 to track this. 

Even though dependencies are not type checked, they are still parsed. A possible improvement would be to skip parsing altogether, @lazamar has done some initial work around this - https://github.com/pepeiborra/ghcide/pull/1

We are reasonably confident that this PR works correctly for large codebases, having tried it on the GHC repository as well as on the Strats codebase at StanChart. The lack of new tests specific to interface files is partially justified by the new diagnostics for unavailable interface files.  

Closes #355


README.md
```diff
@@ -64,6 +64,13 @@ The `ghcide` executable mostly relies on [`hie-bios`](https://github.com/mpicker
 
 If you can't get `ghcide` working outside the editor, see [this setup troubleshooting guide](docs/Setup.md). Once you have got `ghcide` working outside the editor, the next step is to pick which editor to integrate with.
 
+### Optimal project setup
+
+`ghcide` has been designed to handle projects with hundreds or thousands of modules. If `ghci` can handle it, then `ghcide` should be able to handle it too provided the following conditions are met:
+
+1. Use GHC >= 8.6
+2. Give `ghcide` a chance to load all the project modules and generate interface files. This happens automatically for all the dependencies of the modules being edited, but may take a while and require a restart to free up resources after complete.
```
ocharles: This could be clearer - how do I "give `ghcide` a chance" to load everything? Will it automatically find everything itself, or do I have to tell it do something? Do I do this from my editor, or from the command line? Also, you mention it will take a while - how will I know when it's done?

pepeiborra: You are right, this is not very helpful. Since everything happens automatically and (Info) diagnostics are generated with the message `additional resource use while generating interface files in the background` while this is being done, I think we can safely remove the second item in the list. The only caveat is that trying to benchmark ghcide memory usage on a single run is not going to be accurate, since it will be slower / use more memory due to the generation of `.hi`/`.hie` files for project modules.

pepeiborra: How about:
```

`ghcide` has been designed to handle projects with hundreds or thousands of modules. If `ghci` can handle it, then `ghcide` should be able to handle it. The only caveat is that this currently requires GHC >= 8.6, and that the first time a module is loaded in the editor will trigger generation of support files in the background if those do not already exist.
```

ocharles: The latter is much better imo. It tells me what is happening, but doesn't ask me to do anything in particular

ocharles: The latter is much better imo. It tells me what is happening, but doesn't ask me to do anything in particular

jneira: Not sure if all `8.6.*` will support hie.files, if it is not the case we should note the specific 8.6 version that does it.

src/Development/IDE/Core/Rules.hs
```diff
@@ -398,6 +444,88 @@ loadGhcSession = do
...
+                  return (Just (fingerprintToBS (mi_mod_hash iface)), ([], Just result))
+                Left err -> do
+                  let diag = ideErrorWithSource (Just "interface file loading") (Just DsError) f . T.pack $ err
+                  return (Nothing, (pure diag, Nothing))
+
+mkInterfaceFilesGenerationDiag :: a -> String -> [(a, ShowDiagnostic, Diagnostic)]
+mkInterfaceFilesGenerationDiag f intro = mkDiag $ intro <> msg
+  where
+      msg = ": additional resource use while generating interface files in the background."
+      mkDiag = pure . ideErrorWithSource (Just "interface file loading") (Just DsInfo) f . T.pack
```
mpickering: Is this still going to produce the warnings that we discussed?

src/Development/IDE/Core/Compile.hs
```diff
@@ -275,26 +347,40 @@ setupEnv tmsIn = do
...
-
 
 -- | Load a module, quickly. Input doesn't need to be desugared.
 -- A module must be loaded before dependent modules can be typechecked.
 -- This variant of loadModuleHome will *never* cause recompilation, it just
 -- modifies the session.
+--
+-- The order modules are loaded is important when there are hs-boot files.
+-- In particular you should make sure to load the .hs version of a file after the
+-- .hs-boot version.
```
mpickering: This comment is a bit confusing. The order of modules is always important!

src/Development/IDE/Core/Rules.hs
pepeiborra: Updated as suggested here - https://github.com/pepeiborra/ghcide/commit/9f1ce515140da365ba3023d9b68c8413928dda44#r37339599

src/Development/IDE/Core/Compile.hs
pepeiborra: but you authored this comment, according to git blame

src/Development/IDE/Core/Rules.hs
pepeiborra: I've moved the interface loading diagnostics behind the recently added `--test` flag

mpickering:
I have based my reflex reimplementation on this branch and was doing some profiling. The calls to `mkHieFile` accounted for 25% of the profile so it might be something to look out for on this branch as well.

pepeiborra:
> I have based my reflex reimplementation on this branch and was doing some profiling. The calls to `mkHieFile` accounted for 25% of the profile so it might be something to look out for on this branch as well.

That is a bit surprising, I would expect `mkHieFile` to be called only for `getDefinition`. One thing to keep in mind is that we do not cache the hie files in the graph, but they are cached in the file system.

mpickering:
`generateAndWriteHieFile` is called from inside `typecheckRuleDefinition`, at least on the version of the branch I am using. Did you change that?

pepeiborra:
> `generateAndWriteHieFile` is called from inside `typecheckRuleDefinition`, at least on the version of the branch I am using. Did you change that?

Oh that's right sorry, we generate them as a side-effect of type checking, which is needed to avoid long delays on getDefinition. 

This is a very good point. I think a better approach would be to skip generating the `hie` file  when type checking a file of interest, since in that case it's not needed. Same applies to the interface file generation probably.

mpickering:
Most files are not files of interest though? Another easy way to fix this is use the GHC 8.6.5 implementation for all versions as that doesn't traverse the whole AST.

pepeiborra:
> Most files are not files of interest though? Another easy way to fix this is use the GHC 8.6.5 implementation for all versions as that doesn't traverse the whole AST.

Files not-of-interest are only type checked zero-or-once. The costs of generating hie files for them are initialisation costs, which is not a cause of concern, at least currently. 

Files-of-interest are type checked once per keystroke, accounting for 100% of the calls to `typecheckRuleDefinition` after initialisation

mpickering:
Ah right, yes, this was with a cold cache I think.

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -139,11 +138,11 @@ locationsAtPoint getHieFile IdeOptions{..} pos =
                 -- so we instead read the .hie files to get useful source spans.
```
cocreature: The comment that this mostly arises when the definition is in an external package probably needs updating. This is the case for DAML but I believe for Haskell, we will actually never find a `.hie` file for an external package.

src/Development/IDE/Core/Rules.hs
```diff
@@ -398,6 +454,97 @@ loadGhcSession = do
         opts <- getIdeOptions
         return ("" <$ optShakeFiles opts, ([], Just val))
 
+getHiFileRule :: Rules ()
+getHiFileRule = defineEarlyCutoff $ \GetHiFile f -> do
+  session <- hscEnv <$> use_ GhcSession f
+  -- get all dependencies interface files, to check for freshness
+  (deps,_)<- use_ GetLocatedImports f
```
cocreature:
```suggestion
  (deps,_) <- use_ GetLocatedImports f
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -398,6 +454,97 @@ loadGhcSession = do
...
+    Just deps -> do
+      gotHiFile <- getFileExists hiFile
+      if not gotHiFile
+        then do
+          let d = mkInterfaceFilesGenerationDiag f "Missing interface file"
+          pure (Nothing, (d, Nothing))
+        else do
+          hiVersion  <- use_ GetModificationTime hiFile
+          modVersion <- use_ GetModificationTime f
+          let sourceModified = LT == comparing modificationTime hiVersion modVersion
```
cocreature:
```suggestion
          let sourceModified = modificationTime hiVersion < modificationTime modVersion
```
Maybe that is just me but I find that easier to read.

src/Development/IDE/Core/Rules.hs
```diff
@@ -398,6 +454,97 @@ loadGhcSession = do
...
+    case mbHiFile of
+        Just x ->
+            return ([], Just x)
+        Nothing
+          | fileOfInterest -> do
+            -- For files of interest only, create a Shake dependency on typecheck
+            tmr <- use TypeCheck f
+            return ([], extract tmr)
+          | otherwise -> do
+            -- Otherwise invoke typechecking directly
```
cocreature:
```suggestion
            -- Otherwise the interface file does not exist or is out of date. Invoke typechecking directly to update it without incurring a dependency on the typecheck rule.
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -318,46 +343,77 @@ getSpanInfoRule =
...
+  res <- liftIO $ typecheckModule defer hsc (zipWith unpack mirs bytecodes) pm
+
+  case res of
+    (diags, Just (hsc,tcm)) | DoGenerateInterfaceFiles <- generateArtifacts -> do
+      (_, contents) <- getFileContents file
+      diagsHie <- liftIO $
+        generateAndWriteHieFile hsc (TE.encodeUtf8 <$> contents) (tmrModule tcm)
+      diagsHi  <- liftIO $
+        generateAndWriteHiFile hsc tcm
+      return (diags <> diagsHi <> diagsHie, Just tcm)
```
cocreature:
```suggestion
      liftIO $ do
          diagsHie <- generateAndWriteHieFile hsc (TE.encodeUtf8 <$> contents) (tmrModule tcm)
          diagsHi  <- generateAndWriteHiFile hsc tcm
          return (diags <> diagsHi <> diagsHie, Just tcm)
```

mpickering:
I still think there is something broken with go-to definition with this branch to do with absolute vs relative paths. Go-to definition consistently doesn't work for me on GHC's code base.

Could you perhaps verify it works for you @cocreature ?

pepeiborra:
> I still think there is something broken with go-to definition with this branch to do with absolute vs relative paths. Go-to definition consistently doesn't work for me on GHC's code base.
> 
> Could you perhaps verify it works for you @cocreature ?

@mpickering can you confirm it doesn’t happen with master?

test/exe/Main.hs
```diff
@@ -1318,87 +1322,93 @@ findDefinitionAndHoverTests = let
...
-  dnbL29 = Position 29 18  ;  dnb    = [ExpectHoverText [":: ()"],   mkR  29 12   29 21]
-  dnbL30 = Position 30 23
-  lcbL33 = Position 33 26  ;  lcb    = [ExpectHoverText [":: Char"], mkR  33 26   33 27]
-  lclL33 = Position 33 22
-  mclL36 = Position 36  1  ;  mcl    = [mkR  36  0   36 14]
-  mclL37 = Position 37  1
-  spaceL37 = Position 37  24 ; space = [ExpectNoDefinitions, ExpectHoverText [":: Char"]]
-  docL41 = Position 41  1  ;  doc    = [ExpectHoverText ["Recognizable docs: kpqz"]]
+  fffL4  = _start fffR     ;  fffR = mkRange 8  4    8  7 ; fff  = [ExpectRange fffR]
+  fffL8  = Position 12  4  ;
```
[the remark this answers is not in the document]
pepeiborra: Yes. The extra vertical gap is padding in case more imports are needed in the future

test/exe/Main.hs
```diff
@@ -2006,6 +2020,12 @@ runInDir dir s = do
   -- since the package import test creates "Data/List.hs", which otherwise has no physical home
   createDirectoryIfMissing True $ dir ++ "/Data"
 
+  -- Copy all the test data files to the temporary workspace
+  testDataFiles <- getDirectoryFilesIO "test/data" ["//*"]
```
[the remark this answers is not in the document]
pepeiborra: Iirc before there was only one test module, so OpenDoc suffices.

src/Development/IDE/Core/Compile.hs
```diff
@@ -244,18 +261,73 @@ mkTcModuleResult tcm = do
...
+    withBootSuffix = case ms_hsc_src modSummary of
+                HsBootFile -> addBootSuffix
+                _ -> id
+    dflags = hsc_dflags hscEnv
+
+handleGenerationErrors :: DynFlags -> T.Text -> IO () -> IO [FileDiagnostic]
+handleGenerationErrors dflags source action =
+  action >> return [] `catches`
+    [ Handler $ return . diagFromGhcException source dflags
+    , Handler $ return . diagFromString source DsError (noSpan "<internal>")
```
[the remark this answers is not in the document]
pepeiborra: I can't remember exactly, all I remember is being surprised at some that wasn't a `GhcException` being thrown by `writeIfaceFile`. You are completely right that catching `SomeException` is not going to interact well, I'll investigate more

src/Development/IDE/Core/Compile.hs
```diff
@@ -244,18 +261,73 @@ mkTcModuleResult tcm = do
...
+  (write tempFilePath >> renameFile tempFilePath targetPath) `onException` cleanUp
+
+generateAndWriteHieFile :: HscEnv -> Maybe ByteString -> TypecheckedModule -> IO [FileDiagnostic]
+generateAndWriteHieFile hscEnv mb_src tcm =
+  handleGenerationErrors dflags "extended interface generation" $ do
+    src <- maybe (BS.readFile `traverse` srcPath) (return . Just) mb_src
+    case (src, tm_renamed_source tcm) of
+      (Just src, Just rnsrc) -> do
+        hf <- runHsc hscEnv $
+          GHC.mkHieFile mod_summary (fst $ tm_internals_ tcm) rnsrc src
```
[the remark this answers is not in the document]
pepeiborra: The `src` argument is used by the real GHC API to embed the source in the `.hie` file. The 8.6 compatibility layer doesn't use it. 

But your point - that we don't care about `.hie` files embedding the source - is good, it should be fine to drop it.

src/Development/IDE/GHC/Compat.hs
```diff
@@ -178,3 +187,66 @@ dontWriteHieFiles d =
...
 #endif
+
+nameListFromAvails :: [AvailInfo] -> [(SrcSpan, Name)]
+nameListFromAvails as =
+  map (\n -> (nameSrcSpan n, n)) (concatMap availNames as)
+
+#if !MIN_GHC_API_VERSION(8,8,0)
+-- Reimplementations of functions for HIE files for GHC 8.6
+
+mkHieFile :: ModSummary -> TcGblEnv -> RenamedSource -> ByteString -> Hsc HieFile
```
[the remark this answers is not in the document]
pepeiborra: Question for @mpickering

src/Development/IDE/GHC/HieAst.hs
```diff
@@ -0,0 +1,1789 @@
+
+{-
+Forked from GHC v8.8.1 to work around the readFile side effect in mkHiefile
```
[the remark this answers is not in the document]
pepeiborra: Precisely. Matthew has also contributed it upstream so we will be able to drop it in GHC 8.10.1

pepeiborra:
Thank you for the review @cocreature ! 
I have addressed most of the feedback, the remaining bits may have to wait until the weekend.

mpickering:
@pepeiborra It also seems broken on master anyway.

src/Development/IDE/GHC/Compat.hs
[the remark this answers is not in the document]
mpickering: I think the two arguments for using `.hie` files are.

1. Other tools might want to use `.hie` files as well, for example, if `weeder` was integrated into `ghcide`.
2. Contributes will not be surprised to find our custom implementation.

src/Development/IDE/Import/DependencyInformation.hs
```diff
@@ -303,11 +326,21 @@ transitiveDeps DependencyInformation{..} file = do
...
   pure TransitiveDependencies {..}
   where
-    (g, fromVertex, toVertex) = graphFromEdges (map (\(f, fs) -> (f, f, IntSet.toList fs)) $ IntMap.toList depModuleDeps)
+    (g, fromVertex, toVertex) = graphFromEdges edges
+    edges = map (\(f, fs) -> (f, f, IntSet.toList fs ++ boot_edge f)) $ IntMap.toList depModuleDeps
+
+    -- Need to add an edge between the .hs and .hs-boot file if it exists
+    -- so the .hs file gets loaded after the .hs-boot file and the right
+    -- stuff ends up in the HPT. If you don't have this check then GHC will
+    -- fail to work with ghcide.
```
mpickering:
```suggestion
    -- fail to work with ghc.
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -318,46 +342,74 @@ getSpanInfoRule =
...
-        addByteCode :: Linkable -> TcModuleResult -> TcModuleResult
-        addByteCode lm tmr = tmr { tmrModInfo = (tmrModInfo tmr) { hm_linkable = Just lm } }
+typeCheckRule = define $ \TypeCheck file ->
+    -- do not generate interface files as this rule is called
+    -- for files of interest on every keystroke
+    typeCheckRuleDefinition file SkipGenerationOfInterfaceFiles
+
+data GenerateInterfaceFiles
+    = DoGenerateInterfaceFiles
+    | SkipGenerationOfInterfaceFiles
```
mpickering:
```suggestion
    | SkipGenerationOfInterfaceFiles
    deriving (Show)
```

src/Development/IDE/GHC/Compat.hs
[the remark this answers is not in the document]
cocreature: 1 seems compelling. I don’t really buy 2 atm as we have to copy much more code into `ghcide` because the upstream API doesn’t work for us but hopefully that will be fixed in future GHC versions.

Alright, then let’s go with this :+1:

src/Development/IDE/Core/Compile.hs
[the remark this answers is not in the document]
pepeiborra: Alas, I remember seeing GHC  throw an exception other than `GhcException` here but I cannot make it happen. But I am reluctant to remove the `Handler` at this point, so going for `Control.Exception.Safe`. Good spot @cocreature !

src/Development/IDE/Core/Compile.hs
```diff
@@ -244,18 +261,73 @@ mkTcModuleResult tcm = do
   where
     (tcGblEnv, details) = tm_internals_ tcm
 
+atomicFileUpdate :: FilePath -> (FilePath -> IO a) -> IO ()
```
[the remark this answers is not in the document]
pepeiborra: `atomic-write` is great, but not a good fit here as the GHC api is not `Handle` based. I have renamed to `atomicFileWrite` as suggested

pepeiborra:
All the feedback has been addressed and the conflicts resolved.

src/Development/IDE/GHC/Util.hs
```diff
@@ -2,61 +2,71 @@
 -- SPDX-License-Identifier: Apache-2.0
 
 -- | General utility functions, mostly focused around GHC operations.
-module Development.IDE.GHC.Util(
+module Development.IDE.GHC.Util
```
cocreature: Why did you reformat this file now? That makes diff noisy and `git blame` harder to use.

pepeiborra: I guess my editor did - will undo

cocreature: Thanks! CI is failing now due to #479 being merged (CI tests the merge commit, so you need to rebase to reproduce).

PR #459, opened 2020-02-27
Guess all imports

pepeiborra:
At the expense of some space and initialization time, suggest imports now is able to find suggestions in all the packages available to the project.

The only thing missing, but not for this PR, is to guess import in the home modules, i.e. the modules of  the project loaded in the ide.


src/Development/IDE/Plugin/CodeAction/Rules.hs
```diff
@@ -0,0 +1,59 @@
...
+import LoadIface
+import Maybes
+import Module (Module (..), ModuleName, moduleNameString)
+import Packages (explicitPackages, exposedModules, packageConfigId)
+import TcRnMonad (WhereFrom (ImportByUser), initIfaceLoad)
+import GHC (DynFlags(pkgState))
+import Control.Monad.Trans (MonadTrans(lift))
+
+getVisiblePackageExports :: NormalizedFilePath -> Action [PackageExportsMap]
+getVisiblePackageExports f = fmap (fromMaybe []) $ runMaybeT $ do
```
cocreature: This looks unnecessarily expensive. Afaict you are calling this for individual files but it only depends on the GHC session which is the same at least within a component. Can we only do this once per GHC session?

src/Development/IDE/Plugin/CodeAction/Rules.hs
```diff
@@ -0,0 +1,59 @@
...
+        session <- MaybeT $ use GhcSession f
+        let exposedPkgs = explicitPackages $ pkgState (hsc_dflags $ hscEnv session)
+        traverse (\p -> lift $ use_ (fromUnitId p) f) exposedPkgs
+
+rulePackageExports :: Rules ()
+rulePackageExports = define $ \p file -> do
```
cocreature: A type annotation on `p` would be quite useful to see which rule is being defined here.

test/exe/Main.hs
```diff
@@ -900,34 +900,41 @@ suggestImportTests :: TestTree
...
+    , test True []          "f = pack"                    []                "import Data.Text (pack)"
+    , test True []          "f :: Text"                   ["f = undefined"] "import Data.Text (Text)"
     ]
   ]
   where
-    test wanted imps def other newImp = testSession (T.unpack def) $ do
+    test wanted imps def other newImp = testSession' (T.unpack def) $ \dir -> do
       let before = T.unlines $ "module A where" : ["import " <> x | x <- imps] ++ def : other
           after  = T.unlines $ "module A where" : ["import " <> x | x <- imps] ++ [newImp] ++ def : other
+          cradle = "cradle: {direct: {arguments: [-hide-all-packages, -package, base, -package, text, -package-env, -]}}"
```
cocreature: Why do we need the `-package-env -` flag here?

test/exe/Main.hs
```diff
@@ -900,34 +900,41 @@ suggestImportTests :: TestTree
...
+      liftIO$ writeFileUTF8 (dir </> "hie.yaml") cradle
```
cocreature:
```suggestion
      liftIO $ writeFileUTF8 (dir </> "hie.yaml") cradle
```

src/Development/IDE/Plugin/CodeAction/Rules.hs
```diff
@@ -0,0 +1,59 @@
...
+  pkgState <- hscEnv <$> use_ GhcSession file
+  let pkg = lookupPackageConfig (packageExportsUnitId p) pkgState
+  case pkg of
+    Nothing -> return ([], Just mempty)
+    Just pkg -> do
+      results <- forM (exposedModules pkg) $ \(mn, _) -> do
+            modIface <-
+              liftIO $ initIfaceLoad pkgState $
```
cocreature: How does GHC normally load interfaces of dependencies? I thought it was somewhat lazy and didn’t need to suck in all interfaces of all dependencies whereas this will force all of them to be loaded.

src/Development/IDE/Plugin/CodeAction/Rules.hs
pepeiborra: Why do you think this is expensive? It's just retrieving things from the Shake graph, there is no actual computation involved.

If you are concerned it can be cached in the Shake graph, at the cost of more memory.

src/Development/IDE/Plugin/CodeAction/Rules.hs
pepeiborra: That's exactly right. GHC loads dependencies lazily, which is an obstacle to guessing imports of dependencies that have not been loaded in the session so far. To work around that, we load the dependencies upfront. Note that we only load the *direct* dependencies, and not the whole package db.

Since the package modules are loaded via interface files this is relatively fast and lightweight. If it ends up bloating the GHC session too much, we could look into loading them into a separate session, then throwing it away after harversting all the info.

test/exe/Main.hs
pepeiborra: I won't claim to understand how package environments work, but see https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/packages.html#package-environments

src/Development/IDE/Plugin/CodeAction/Rules.hs
cocreature: Sorry should have been more clear, I mean the `rulePackageExports` rule which seems to be called once per `dependencies * files` in your project.

pepeiborra: You are right, the `PackageExports` rule is file-dependent. The suggestion is to make it session dependent instead. But that seems tricky, since a session is not something that can be used as a key in the Shake graph. Will have to think about it

cocreature: `HscEnvEq` can probably be used as a key so you might be able to define it with `defineNoFile` and accept the `HscEnvEq` as an argument. Alternatively something like the `memoIO` approach used to initialize the session itself might work but that might be a bit tricky given that the internals of `ghcide` don’t know about this.

pepeiborra: Thanks for the suggestion to go with `HscEnvEq`.

pepeiborra:
Anything missing here?

PR #460, opened 2020-02-27
Tests for cradle loading

pepeiborra:
Tests to ensure that the cradle loading logic does not regress into calling `runCradle` multiple times per cradle

### DIFF
--- a/exe/Rules.hs
+++ b/exe/Rules.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE OverloadedStrings #-}
 module Rules
   ( loadGhcSession
   , cradleToSession
--- a/exe/Rules.hs
+++ b/exe/Rules.hs
@@ -8,7 +8,7 @@ module Rules
 where
 
 import           Control.Exception
-import           Control.Monad                  (filterM)
+import           Control.Monad                  (filterM, when)
 import qualified Crypto.Hash.SHA1               as H
 import           Data.ByteString.Base16         (encode)
 import qualified Data.ByteString.Char8          as B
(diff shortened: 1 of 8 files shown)


test/exe/Main.hs
```diff
@@ -1864,6 +1865,44 @@ haddockTests
...
+            msgs <- manyTill (skipManyTill anyMessage cradleLoadedMessage) (skipManyTill anyMessage (message @PublishDiagnosticsNotification))
+            liftIO $ length msgs @?= 0
+            _ <- openDoc' "A.hs" "haskell" "module A where\nimport Bar"
+            msgs <- manyTill (skipManyTill anyMessage cradleLoadedMessage) (skipManyTill anyMessage (message @PublishDiagnosticsNotification))
+            liftIO $ length msgs @?= 0
+
+
+cradleLoadedMessage :: Session ()
+cradleLoadedMessage = do
+    m <- anyMessage
```
cocreature: this looks like it might be a bit nicer if you use `satisfy` instead of first using `anyMessage` which is `satisfy $ const True` and then doing the filtering afterwards.

exe/Rules.hs
```diff
@@ -0,0 +1,150 @@
...
+
+cradleToSession :: Rules ()
+cradleToSession = define $ \LoadCradle nfp -> do
+    let f = fromNormalizedFilePath nfp
+
+    -- If the path points to a directory, load the implicit cradle
+    mbYaml <- doesDirectoryExist f <&> \isDir -> if isDir then Nothing else Just f
+    cradle <- liftIO $  maybe (loadImplicitCradle $ addTrailingPathSeparator f) loadCradle mbYaml
+
+    sendEvent $ notifyCradleLoaded f
```
cocreature: I would prefer to hide this behind a flag that we pass only in the tests. Especially if the message does not start with `$/`, clients might fall over if they see this message and don’t understand it and even if we prefix it with `$/` it’s really something that is only intended for tests.

pepeiborra: Done!

test/exe/Main.hs
pepeiborra: Done, thanks!

PR #462, opened 2020-02-29
Detect ghc mismatch

pepeiborra:
Note that this PR builds on top of #460 

This PR makes ghcide fail early if the compile-time and run-time versions of ghc do not match, with a clear error message that should decrease the maintenance burden. It builds on top of a tiny package called `ghc-check` that I have put together to encapsulate the potentially reusable bits.

The PR may require adjustment for ghc-lib, but I'm not entirely sure under which scenarios ghc-lib is used, and what it means to use ghc-lib instead of ghc.

### DIFF
--- a/exe/Rules.hs
+++ b/exe/Rules.hs
@@ -8,13 +8,14 @@ module Rules
 where
 
 import           Control.Exception
-import           Control.Monad                  (filterM, when)
+import           Control.Monad                  (filterM, when, unless)
 import qualified Crypto.Hash.SHA1               as H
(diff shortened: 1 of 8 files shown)


pepeiborra:
1. What's there to do exactly? The ghc version check lives in the executable, which is not buildable when the ghc-lib cabal flag is enabled
2. Added explicit libDir support
3. Rebased
4. Improved the UI by delivering the error as an lsp diagnostic instead of an initialization crash

cocreature:
Needs rebasing.

PR #463, opened 2020-03-01
Add a note about hacking on ghcide from Windows

ndmitchell:
See https://github.com/digital-asset/ghcide/pull/435#issuecomment-592504569 for context

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ using `stack install ghcide` within a project.
 
 To build and work on `ghcide` itself, you can use Stack or cabal, e.g.,
 running `stack test` will execute the test suite.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -252,6 +252,7 @@ using `stack install ghcide` within a project.
 
 To build and work on `ghcide` itself, you can use Stack or cabal, e.g.,
 running `stack test` will execute the test suite.
+If you are using Windows, you should disable the `auto.crlf` setting and configure your editor to use LF line endings.
```
jneira:
```suggestion
If you are using Windows, you should disable the `auto.crlf` setting and configure your editor to use LF line endings, directly or making it use the existing `.editor-config`.
```

PR #465, opened 2020-03-03
Try removing allow-newer

ndmitchell:
Seems to cause problems for users, e.g. #464


cocreature:
However, I don’t think this has anything to do with #464. This only changes the `stack.yaml` for GHC 8.4 whereas #464 is about GHC 8.8.

ndmitchell:
Agreed, it's not the immediate cause of that issue (I suspect they generated their own stack file), but it shows having this in the code is a bad idea.

PR #471, opened 2020-03-08
Fix two regressions since 0.1.0

pepeiborra:
I noticed these running VSCode + ghcide master on the GHC codebase

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -430,13 +430,13 @@ codeLensesTests = testGroup "code lenses"
 watchedFilesTests :: TestTree
 watchedFilesTests = testGroup "watched files"
   [ testSession "workspace file" $ do
-      _ <- openDoc' "A.hs" "haskell" "module A where"
+      _ <- openDoc' "C/A.hs" "haskell" "module A where"
       RequestMessage{_params = RegistrationParams (List regs)} <- skipManyTill anyMessage (message @RegisterCapabilityRequest)
       let watchedFileRegs =
             [ args | Registration _id WorkspaceDidChangeWatchedFiles args <- regs ]
       liftIO $ assertBool "watches workspace files" $ not $ null watchedFileRegs
   , testSession "non workspace file" $ do
-      _ <- openDoc' "/tmp/A.hs" "haskell" "module A where"
+      _ <- openDoc' "/tmp/A.hs" "haskell" "{-# LANGUAGE NoImplicitPrelude#-}\nmodule A where"
       msgs <- manyTill (Just <$> message @RegisterCapabilityRequest <|> Nothing <$ anyMessage) (message @WorkDoneProgressEndNotification)
       let watchedFileRegs =
             [ args
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -12,6 +12,7 @@ import Control.Applicative.Combinators
 import Control.Exception (catch)
 import Control.Monad
 import Control.Monad.IO.Class (liftIO)
+import Data.Aeson (Value)
 import Data.Char (toLower)
 import Data.Foldable
 import Data.List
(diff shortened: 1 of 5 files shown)


mpickering:
How did you observe these?

pepeiborra:
> How did you observe these?

Shake profiling showed an unexpected large number of `GetFileExists` calls for hovers. I ran into the second one anecdotally during workspace reloads

ndmitchell:
FWIW, I noticed the issue with hie-bios earlier today, and raised the upstream issue at https://github.com/snoyberg/yaml/issues/185 - since that's the most common way I had the hie-bios initialisation failing. Agreed this is a good fix though! I do wonder if uninteruptibleMask is required though, since everywhere I've used mask alone, sooner or later I've regretted it.

mpickering:
I have also been seeing that issue, thanks for tracking it down Neil and Pepe.

pepeiborra:
> FWIW, I noticed the issue with hie-bios earlier today, and raised the upstream issue at [snoyberg/yaml#185](https://github.com/snoyberg/yaml/issues/185) - since that's the most common way I had the hie-bios initialisation failing. Agreed this is a good fix though! I do wonder if uninteruptibleMask is required though, since everywhere I've used mask alone, sooner or later I've regretted it.

I don't think this is the same issue though. The problem I'm trying to fix is the `LoadCradle` retry happening before the OS has had a chance to reclaim the resources used by the external processes started by hie-bios. In the case of Hadrian, the Shake file lock. Before the fix, this was failing as below.

About `mask` vs `uninterruptibleMask`, I'm happy to follow your advice here.

```
Started LSP server in 0.30s
[DEBUG] Set files of interest to: [NormalizedFilePath "/home/pepe/scratch/ghc/compiler/main/GHC.hs"]
[DEBUG] Starting shakeRun (aborting the previous one took 0.00s)
[INFO] Opened text document: file:///home/pepe/scratch/ghc/compiler/main/GHC.hs
[DEBUG] Running cradle /home/pepe/scratch/ghc/hie.yaml
[DEBUG] Starting shakeRun (aborting the previous one took 0.00s)
[DEBUG] Finishing shakeRun (took 0.04s, exception: AsyncCancelled)
[DEBUG] Running cradle /home/pepe/scratch/ghc/hie.yaml
> hadrian: Shake failed to acquire a file lock on .hie-bios/hadrian/.shake.lock
> Shake process ID 31573 is using this lock.
> setLock: resource exhausted (Resource temporarily unavailable)
> CallStack (from HasCallStack):
>   errorIO, called at src/General/FileLock.hs:67:9 in shake-0.18.3-af1d490ec7e20705e035a339d81952289a7a6aad11ff39a6e3d7c78f9a6beb32:General.FileLock
```

test/exe/Main.hs
```diff
@@ -430,13 +430,13 @@ codeLensesTests = testGroup "code lenses"
...
   [ testSession "workspace file" $ do
-      _ <- openDoc' "A.hs" "haskell" "module A where"
+      _ <- openDoc' "C/A.hs" "haskell" "module A where"
       RequestMessage{_params = RegistrationParams (List regs)} <- skipManyTill anyMessage (message @RegisterCapabilityRequest)
       let watchedFileRegs =
             [ args | Registration _id WorkspaceDidChangeWatchedFiles args <- regs ]
       liftIO $ assertBool "watches workspace files" $ not $ null watchedFileRegs
   , testSession "non workspace file" $ do
-      _ <- openDoc' "/tmp/A.hs" "haskell" "module A where"
+      _ <- openDoc' "/tmp/A.hs" "haskell" "{-# LANGUAGE NoImplicitPrelude#-}\nmodule A where"
```
cocreature: What changed to require this?

test/exe/Main.hs
```diff
@@ -430,13 +430,13 @@ codeLensesTests = testGroup "code lenses"
 watchedFilesTests :: TestTree
 watchedFilesTests = testGroup "watched files"
   [ testSession "workspace file" $ do
-      _ <- openDoc' "A.hs" "haskell" "module A where"
+      _ <- openDoc' "C/A.hs" "haskell" "module A where"
```
cocreature: Why did this change?

test/exe/Main.hs
pepeiborra: Now there is a `DidChangeWatchedFile` subscription message for `Prelude` on various paths, which were not emitted before as they would use relative paths.

cocreature: I’m not following, why do we get a `DidChangeWatchedFile` notification for `Prelude`? Surely we are not editing the contents of that file in this test?

test/exe/Main.hs
pepeiborra: This is an incomplete attempt at making a test more demanding. Will try to push an improved version tonight.

test/exe/Main.hs
pepeiborra: `DidChangeWatchedFile` subscriptions are used to implement the `GetFileExists` Shake rule, which is used by `GetLocatedImports` and other Shake rules. So it's not just a matter of whether we are editing the contents of that file in the test.

test/exe/Main.hs
pepeiborra: pushed

test/exe/Main.hs
cocreature: Oh sorry, I mixed up `DidChangeTextDoucment` and `DidChangeWatchedFiles`. Makes sense now :+1:

cocreature:
looks like this needs rebasing on master

pepeiborra:
Rebased and force pushed. By the way, I ended up not following the suggestion to use `uninterruptibleMask` as `mask` seems enough and the Haddocks for `uninterruptibleMask` are very scary - I don't want to make ghcide unkillable if e.g. the wrong craddle is selected and stack is building the world.

test/exe/Main.hs
alanz: I am hitting this problem in haskell-language-server tests too.  Is the correct solution not to filter out the watch request on non-existent Prelude files?

cocreature: @alanz no, we need to detect if a `Prelude` file starts existing.

alanz: @cocreature The file monitored as the location of the implicit prelude file is assumed to be in the same directory as the original file. Or in some unlikely location, like deep in the stack workspace, with the project subdirectory appended.

So does it make sense to monitor for that file? Would it not make more sense to monitor wherever the implicit prelude actually exists/comes from?  If that file does suddenly spring into existence, will it replace the implicit prelude?

cocreature: Prelude isn’t all that special in GHC. E.g., you can use `base-noprelude` and provide your own `Prelude` module and GHC will pick that. So I don’t think it makes sense to special case prelude in any way.

Issue #474, opened 2020-03-09
Failing tests in windows


jneira:
I've run the test suite locally and in [azure ci](https://github.com/digital-asset/ghcide/issues/455) and i've got those failing tests:
```
 cpp
  cpp-error:                                        FAIL (4.16s)
diagnostics
  add missing module (non workspace):               FAIL
watched files
  non workspace file:                               FAIL
```
Full test log of failing tests here: https://gist.github.com/jneira/afe77151a540c5a27943d7661eefe3fa

jneira:
Locally (windows 7) i've got an additional one:
```
  plugins:                                            FAIL (0.45s)
    test/src\Development\IDE\Test.hs:35:
    Could not find (DsError,(8,14),"Variable not in scope: c") in List [Diagnostic {_range =
Range {_start = Position {_line = 0, _character = 0}, _end = Position {_line = 100000, _character = 0}}, _severity = Just DsError, _code = Nothing, _source = Just "typecheck", _message = "Unexpected usage error\nCould not find module \8216GHC.TypeLits.KnownNat.Solver\8217\nUse -v to see a list of the files searched for.\n", _relatedInformation = Nothing}]
```

ndmitchell:
One of the CPP ones is actually a GHC bug which is being fixed as part of https://gitlab.haskell.org/ghc/ghc/issues/17786

jneira:
I am thinking in creating a helper function `knowBrokenInWindows`, use it in the actual failing tests for windows. 
At least the test suite will be executed.

jneira:
Local failing tests:
```
cpp
    cpp-error:                                                                                            FAIL (3.30s)
      test/src\Development\IDE\Test.hs:41:
      Could not find (DsError,(2,1),"error: unterminated") in List [Diagnostic {_range = Range {_start = Position {_line = 0, _character = 0}, _end = Position {_line = 1, _character = 0}}, _severity = Just DsError, _code = Nothing, _source = Just "compiler", _message = "`gcc.exe' failed in phase `C pre-processor'. (Exit code: 1)", _tags = Nothing, _relatedInformation = Nothing}]

add missing module (non workspace):                                                                   FAIL
      Exception: Maybe.fromJust: Nothing
      CallStack (from HasCallStack):
        error, called at libraries\base\Data\Maybe.hs:148:21 in base:Data.Maybe
        fromJust, called at test/src\Development\IDE\Test.hs:108:59 in main:Development.IDE.Test

[...]

PR #475, opened 2020-03-09
Add azure job for windows and restructure azure config files

### LINKED ISSUE #455: Add windows to azure ci
* The project is not tested in ci with windows and it would catch some issues earlier
* Not sure it it had been tried and some issue made it not feasible
* The setup for haskell-ide-engine could be used as reference: https://github.com/haskell/haskell-ide-engine/blob/master/.azure/windows-stack.yml

jneira:
* Tests are disabled for now cause there are [three failing ones.](https://github.com/digital-asset/ghcide/issues/474) They have to be:
  * Fixed :smile: 
  * Marked as broken in test code
  * Avoided using `--pattern`
  * But it could be done in a follow-up
* I've used some azure elements not present in existing jobs (using Cache@2, using matrix startegy). They can be update in (another) follow-up if you consider it could a improvement over them
* I've added branches starting with `azure` in the trigger, cause i've found them useful to test azure configurations without worrying of trigger manual executions

Closes #455


azure-pipelines.yml
```diff
@@ -197,3 +198,68 @@ jobs:
                --data "{\"text\":\"<!here> *FAILED* $(Agent.JobName): <https://dev.azure.com/digitalasset/ghcide/_build/results?buildId=$(Build.BuildId)|$MESSAGE>\n\"}" \
                $(Slack.URL)
         condition: and(failed(), eq(variables['Build.SourceBranchName'], 'master'))
+  - job: ghcide_stack_windows
```
cocreature: Can we use a single matrix for all platforms? It seems a bit weird that the Windows test setup looks very different from the Linux test setup and also makes it hard to figure out what is different. The matrix setup definily looks nicer than the mess we have currently.

jneira: well the bash script and cache handling for linux and windows are different and i am not sure if they can be integrated without too much if's
what about a matrix per os like in hie?

jneira:
@cocreature i've used the hie layout for azure configuration, using templates and a job for each operating system.
I think that it will be more maintainable this way cause it is probable that macos will be added (with its own way to installing deps) and the scripts of each step would be either one-liners or up to three if blocks. Even steps itself could be different. 
If we want to reuse bash script, like the used to send a notificaction we can extract it and call from the azure config file.

cocreature:
Needs rebasing on master.

jneira:
I rebased it but the build has failed for two windows jobs, they dont seem to be related with the changes though. I am rerunning it in my azure pipeline: https://dev.azure.com/jneira/haskell-language-server/_build/results?buildId=726

jneira:
I wonder if 
/azp run
would work for me

jneira:
@cocreature my pipeline ran succesfully https://dev.azure.com/jneira/haskell-language-server/_build/results?buildId=726

PR #477, opened 2020-03-10
Working on Plugin support for haskell-language-server

alanz:
This tweaks the existing Plugin infrastructure to

~~[ ] add `PluginCommands` to the `Plugin` data definition~~
[ ] return `ResponseError` when a plugin fails
[ ] move the command registration list into `exe/Main.hs`

This allows us to build a layer in haskell-language-server that specifies plugins based on a particular tool, be it a formatter, diagnostics, provider, whatever, rather than focusing on the handlers.

The layer also manages to combine the output of the assorted plugins, and is based directly on the code in haskell-ide-engine.  The branch making use of this is at https://github.com/alanz/haskell-language-server/tree/plugin-api


alanz:
And I just realised the whole `PluginCommand` thing is not used at all in ghcide. So should probably migrate to hls instead.

alanz:
The CI fails because of an hlint warning, about `OverloadedStrings` appearing in `exe/Main.hs`.

But OverloadedStrings is set as a default extension in the cabal file on the master branch.

So something is not right here, either remove the hlint failure condition, or disable the extension in the cabal file too.  Or do I misunderstand?

cc @cocreature @ndmitchell

cocreature:
It’s the other way around. Because it is in `default-extensions` you should not enable it in the file.

alanz:
@cocreature ok, but the log is not very clear

```
2020-03-11T20:31:31.9393812Z ######################################################################## 100.0%
2020-03-11T20:31:39.0698941Z ./exe/Main.hs:5:1: Warning: Avoid restricted extensions
2020-03-11T20:31:39.0700347Z Found:
2020-03-11T20:31:39.0701633Z   {-# LANGUAGE OverloadedStrings #-}
2020-03-11T20:31:39.0702234Z Note: may break the code
2020-03-11T20:31:39.0702712Z 
2020-03-11T20:31:39.0703142Z 1 hint
2020-03-11T20:31:39.0993288Z 
2020-03-11T20:31:39.1077451Z ##[error]Bash exited with code '1'.
2020-03-11T20:31:39.1093451Z ##[section]Finishing: HLint via ./fmt.sh
2020-03-11T20:31:39.1229032Z ##[section]Starting: Cache stack artifacts
```

alanz:
@cocreature how do I request a review? Bar pinging you now, that is.

exe/Main.hs
```diff
@@ -84,12 +85,15 @@ main = do
     let plugins = Completions.plugin <> CodeAction.plugin
         onInitialConfiguration = const $ Right ()
         onConfigurationChange  = const $ Right ()
+        options = def { LSP.executeCommandCommands = Just ["typesignature.add"]
+                      , LSP.completionTriggerCharacters = Just "."
+                      }
```
ndmitchell: What's the reason for these two changes to options? I don't see what's necessary. Is it fixing a bug? Enabling something else - and if so what?

src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -147,14 +147,16 @@ runLanguageServer options userHandlers onInitialConfig onConfigChange getIdeStat
...
-                            sendFunc $ wrap $ ResponseMessage "2.0" (responseId _id) (Just res) Nothing
-                            case newReq of
-                                Nothing -> return ()
-                                Just (rm, newReqParams) -> do
-                                    reqId <- getNextReqId
-                                    sendFunc $ wrapNewReq $ RequestMessage "2.0" reqId rm newReqParams
+                                case res of
+                                    Left e  -> sendFunc $ wrap $ ResponseMessage "2.0" (responseId _id) Nothing (Just e)
+                                    Right r -> sendFunc $ wrap $ ResponseMessage "2.0" (responseId _id) (Just r) Nothing
+                                case newReq of
```
ndmitchell: I'd have written: `whenJust newReq $ \(rm, newReqParams) -> ...` - but that's how I'd have written the previous one, so personal preference.

src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -147,14 +147,16 @@ runLanguageServer options userHandlers onInitialConfig onConfigChange getIdeStat
                               Left e  -> sendFunc $ wrap $ ResponseMessage "2.0" (responseId _id) Nothing (Just e)
                               Right r -> sendFunc $ wrap $ ResponseMessage "2.0" (responseId _id) (Just r) Nothing
                     ResponseAndRequest x@RequestMessage{_id, _params} wrap wrapNewReq act ->
-                        checkCancelled ide clearReqId waitForCancel lspFuncs wrap act x _id _params $
+                        checkCancelled ide clearReqId waitForCancel lspFuncs wrap act x _id _params $ do
```
ndmitchell: What's this `do` doing? And is it in the reader monad?

exe/Main.hs
alanz: In haskell-language-server, the plugin mechanism allows each plugin to register a list of commands. So we only know when the plugins are assembled which ones we need to register.  Hence it needs to be an option we pass in, rather than hardcoded inside ghcide.

Similarly, we allow multiple completion providers, which may trigger on different characters depending on need.

ndmitchell: Yep, I can totally guess that this needs to become a list passed in and modified. But here you are just hardcoding it, without any corresponding other change anywhere, which puzzles me.

src/Development/IDE/LSP/LanguageServer.hs
alanz: we now have two case statements one after each other, so need to sequence them.  I have no idea what Monad it is in, I just made GHC happy :)

If someone wants to propose an alternative implementation for this I am happy to use it. Main thing is to populate either the error or result portion of the reply message.

exe/Main.hs
alanz: This is the `ghcide` version.

See 
https://github.com/alanz/haskell-language-server/blob/plugin-api/exe/Main.hs#L149-L150
where I just realised I have not worked through the trigger characters yet. But they will be needed, may as well bring them in as a coherent change.

src/Development/IDE/LSP/LanguageServer.hs
ndmitchell: Both case statements are nested under a lambda which already has a `do` on it. This do only seems to wrap the function.

src/Development/IDE/LSP/LanguageServer.hs
cocreature: Definitely in favor of using `whenJust`.

src/Development/IDE/LSP/LanguageServer.hs
cocreature: Yeah that `do` doesn’t look right to me. Are you sure we cannot simply remove it?

src/Development/IDE/Plugin.hs
```diff
@@ -1,5 +1,9 @@
-
-module Development.IDE.Plugin(Plugin(..), codeActionPlugin, codeActionPluginWithRules) where
+module Development.IDE.Plugin
```
cocreature: Please don’t include unnecessary layout changes in a PR. That just creates noise during review.

PR #479, opened 2020-03-11
Use NormalizedFilePath and adapt types of haskell-lsp-0.21 

jneira:
* This should wait to a `haskell-lsp-types` release with https://github.com/alanz/haskell-lsp/pull/224 included but i wanted to implement the ghcide adaption in case we have to change the actual `haskell-lsp` pr
* To keep ghcide special case for empty paths i had to duplicate `fromNormalizedFilePath` and `toNormalizedFilePath` and make a newtype for `NormalizedPath` to override the `IsString` instance. It needs hiding those definitions from `Haskell.LSP.Types` in several modules. Let me know if you would prefer another solution (like remove the IsString instance? use functions versions with `'` to avoid hidings?)
* I am little bit worried with performance and i am not sure how to measure it for the pr. Assuming that test suite duration can be an aproximation, [i've ran it with master and the pr in azure](https://dev.azure.com/jneira/haskell-language-server/_build?definitionId=4&_a=summary) several times:

#20200311.20 master         c74e9b5 All 233 tests passed (322.79s)
#20200311.24 master         c74e9b5 All 233 tests passed (340.53s)
#20200311.27 master         c74e9b5 All 233 tests passed (321.96s)
#20200311.32 master         c74e9b5 All 233 tests passed (280.83s)

#20200311.23 norm-filepath  d718658 All 233 tests passed (276.39s)
#20200311.25 norm-filepath  d718658 All 233 tests passed (312.28s)
#20200311.28 norm-filepath  d718658 All 233 tests passed (327.65s)
#20200311.30 norm-filepath  d718658 All 233 tests passed (328.59s)

#20200311.22 norm-fp-nowrap 04d4679 All 233 tests passed (295.12s)
#20200311.26 norm-fp-nowrap 04d4679 All 233 tests passed (298.30s)
#20200311.29 norm-fp-nowrap 04d4679 All 233 tests passed (319.41s)
#20200311.31 norm-fp-nowrap 04d4679 All 233 tests passed (330.24s)

At least the pr times are similar to master.


jneira:
I've tested manually in vscode over ghcide and i dont note any significative perfomance degradation.
So i think this could be reviewed :smile: // cc @cocreature

jneira:
I am getting an error statring ghcide with this branch:
```
[Error - 11:43:14] Request textDocument/documentSymbol failed.
  Message: Error when running Shake build system:
  at action, called at exe\\Main.hs:101:80 in main:Main
  at apply, called at src\Development\IDE\Core\Shake.hs:532:63 in ghcide-0.1.0-KJEoEjwtfuKt9utY5Hxte:Development.IDE.Core.Shake
* Depends on: TypeCheck; D:\ws\haskell\ghcide\src\Development\IDE\Core\IdeConfiguration.hs
  at apply, called at src\Development\IDE\Core\Shake.hs:532:63 in ghcide-0.1.0-KJEoEjwtfuKt9utY5Hxte:Development.IDE.Core.Shake
* Depends on: GetParsedModule; D:\ws\haskell\ghcide\src\Development\IDE\Core\IdeConfiguration.hs
  at apply, called at src\Development\IDE\Core\Shake.hs:532:63 in ghcide-0.1.0-KJEoEjwtfuKt9utY5Hxte:Development.IDE.Core.Shake
* Depends on: GhcSession; D:\ws\haskell\ghcide\src\Development\IDE\Core\IdeConfiguration.hs
  at apply, called at src\Development\IDE\Core\Shake.hs:532:63 in ghcide-0.1.0-KJEoEjwtfuKt9utY5Hxte:Development.IDE.Core.Shake
* Depends on: LoadCradle; D:\ws\haskell\ghcide\hie.yaml
* Raised the exception:
thread killed

  Code: -32603 
```
After the error vscode seems to work fine, though :thinking:

src/Development/IDE/Types/Location.hs
```diff
@@ -10,151 +10,63 @@ module Development.IDE.Types.Location
...
-toNormalizedFilePath "" = NormalizedFilePath (NormalizedUriWrapper emptyPathUri) (hash ("" :: String)) ""
-toNormalizedFilePath fp =
-  let nfp = normalise fp
-  in NormalizedFilePath (NormalizedUriWrapper $ filePathToUriInternal' nfp) (hash nfp) nfp
+toNormalizedFilePath "" = NormalizedFilePath $ LSP.NormalizedFilePath emptyPathUri ""
+toNormalizedFilePath fp = NormalizedFilePath $ LSP.toNormalizedFilePath fp
 
 fromNormalizedFilePath :: NormalizedFilePath -> FilePath
-fromNormalizedFilePath (NormalizedFilePath _ _ fp) = fp
+fromNormalizedFilePath (NormalizedFilePath nfp) =  LSP.fromNormalizedFilePath nfp
```
cocreature: I would prefer a version with `'`. That is consistent with `uriToFilePath'` and it has the advantage that it’s easy to grep for all uses of the unsuffixed version to audit whether they are fine. Currently I need to take imports into account which makes grep pretty useless.

jneira:
Thanks for reviewing this

> As for the error you are seeing, are you sure this was introduced by this branch? It looks like somehow the exception we send to kill a thread if something changed is interpreted as an actual error. I’m not sure I’ve seen this before but it also doesn’t look related to your PR. Maybe this is something windows specific?

I've tested again in my windows 10 and the error is not reproduced. :+1:

> As for performance tests, I expect that there might be too much noise in the test suite for it to be a reliable test and all the tests have a small number of files whereas the performance here matters more for a large number of files. I would recommend to run a shake profile on hover, like in [#254 (comment)](https://github.com/digital-asset/ghcide/pull/254#issuecomment-569924242) and compare that with and without this PR.

Again the shake profiling files are generated in windows 10 (it did not work in my other windows env, to investigate). I am just generatint them for both versions.

jneira:
I've tested to open all the files under `src\Development\Ide\Core`,making hovers, change an identifier and applying a quick fix with results quite similar between the pr and master:
* This pr: 
imagen
* Master (8b328bb)
imagen

jneira:
@cocreature i've renamed the ghcide specific newtype and functions adding `'`.

src/Development/IDE/Types/Location.hs
```diff
@@ -10,151 +10,64 @@ module Development.IDE.Types.Location
...
-  , Uri(..)
-  , toNormalizedUri
-  , fromNormalizedUri
-  )
 import SrcLoc as GHC
 import Text.ParserCombinators.ReadP as ReadP
-import GHC.Generics
 
+-- We need a newtype to override the IsString instance
+newtype NormalizedFilePath' = NormalizedFilePath' LSP.NormalizedFilePath
```
cocreature: I think it’s better to just kill the `IsString` instance. The few cases where we actually have `""` literals should be easy to detect and having a separate type all over the place seems confusing.

jneira: Mmm ok, but are you ok with keeping `toNormalizedFilePath'`/`fromNormalizedFilePath'` (with `'`) to take in account the empty path case?

cocreature: Yes, those are harder to get rid off since the input is a variable rather than a literal so I think it makes sense to keep them.

jneira:
@cocreature It looks cleaner without the `IsString` instance and the newtype.

jneira:
Mmm i cant reproduce the error shown in ci with `stack test --ghc-options=-Werror --stack-yaml ghc88.yaml`:

```
[4 of 5] Compiling Rules

/home/vsts/work/1/s/exe/Rules.hs:22:1: error: [-Wunused-imports, -Werror=unused-imports]
    The import of ‘Development.IDE.Types.Location’ is redundant
      except perhaps to import instances from ‘Development.IDE.Types.Location’
    To import instances alone, use: import Development.IDE.Types.Location()
   |
22 | import           Development.IDE.Types.Location (fromNormalizedFilePath)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
If i remove the import the build fails in local. 🤔

cocreature:
Note that azure tests merge commits rather than your branch. So if you get an error that you cannot reproduce locally it might just be that you need to rebase on master.

jneira:
Just i had rebased master in the commit force pushed before the first ci error, i'll try again

jneira:
@cocreature i've used the recent released version of `haskell-lsp` but i have to make some changes to adapt ghcide code to other type changes. I hope they will be sensible

jneira:
I've corrected the stack.yaml used in the key of azure cache ci. I think the build errors were related with that.

jneira:
I've added `stack test --ghc-options="-Wall" --no-run-tests` to try to reproduce the error 
```The import of ‘Development.IDE.Types.Location’ is redundant...```

It is reproduced in windows and linux but only for `ghc88.yaml`. I cant reproduce it in my windows using the exactly same `stack`invocation ans `stack88.yaml` files.

jneira:
ed0e4b8 has fixed the build for ghc-8.8.3, but it worked for other ghc versions in ci and in other envs for that ghc-version before so 🤷‍♂

cocreature:
@jneira Ah yeah, GHC 8.8 has a different/improved unused imports checker. I noticed that when I was trying to upgrade the DAML codebase as well.

PR #483, opened 2020-03-15
Remove unused code

ocharles:
The following code was identified with https://github.com/ocharles/weeder/.

I'm sure there is code that should be retained, but I thought it might be helpful for maintainers to see this, while I work on weeder.

### DIFF
--- a/src/Development/IDE/Spans/Common.hs
+++ b/src/Development/IDE/Spans/Common.hs
@@ -12,7 +12,6 @@ module Development.IDE.Spans.Common (
 , SpanDoc(..)
 , emptySpanDoc
 , spanDocToMarkdown
-, spanDocToMarkdownForTest
 ) where
 
 import Data.Data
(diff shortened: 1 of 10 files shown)


src/Development/IDE/Spans/Common.hs
```diff
@@ -80,10 +79,6 @@ spanDocToMarkdown (SpanDocString _)
 #endif
 spanDocToMarkdown (SpanDocText txt) = txt
 
-spanDocToMarkdownForTest :: String -> String
```
cocreature: this is even used in `ghcide`s own test suite, shouldn’t weeder notice this?

ocharles: Ah, I didn't build with tests. I tend to take the view that test-only code should live in test libraries, so when I run weeder I don't enable tests.

ocharles:
Thanks for the speedy review!

PR #484, opened 2020-03-15
Use tasty-rerun to allow rerunning failed tests only

alanz:
Use by

  stack --stack-yaml=stack84.yaml test --test-arguments "--rerun"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist-newstyle/
 cabal.project.local
 *~
 *.lock
+/.tasty-rerun-log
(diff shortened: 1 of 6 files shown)


ocharles:
Ha, cool to see some of my early creations have users! :smile:

alanz:
I am used to this basic functionality in hspec.

I would *love* to have the equivalent of the hspec `--match` capability to just run arbitrary tests.

And this is the fundamental reason haskell-language-server decided to stick with hspec for tests.

But now I am diving into ghcide myself, need to improve the experience a bit.

cocreature:
> I would _love_ to have the equivalent of the hspec `--match` capability to just run arbitrary tests.

Are you looking for Tasty’s `-p` flag?

alanz:
> Are you looking for Tasty’s -p flag?

Sounds like it. I did a search but did not find anything, obviously not looking that well.

cocreature:
Tasty’s readme has pretty decent documentation for it https://github.com/feuerbach/tasty/#patterns

jneira:
i personally like hspec `--match` and `--skip` matching rules over tasty `--pattern` (but i guess it is matter of habit)

PR #495, opened 2020-03-20
Improve completion contexts

cocreature:
The completion context determines whether we show completions for
types or completions for values. This is done by looking at the parsed
module.

This PR fixes two things:

1. While we only use the parsed module for getting the context
   previously we got the parsed module out of the typechecked
   module. This means that if you have a module that parses but
   doesn’t typecheck, we will use the parsed module at the point where
   it last typechecked which is out of date and produces incorrect (or
   just no) contexts.
2. When we could not find a context, we defaulted to assuming we are
   in a value context. Especially in combination with 1 but also just
   in general, this is rather annoying. If we aren’t sure we should
   show the user everything we have and not filter out some
   completions. Filtering out completions interacts particularly badly
   with VSCode’s default behavior of accepting the first completion
   when you press return.

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -1630,6 +1630,27 @@ completionTests
         docId <- openDoc' "A.hs" "haskell" source
         compls <- getCompletions docId (Position 1 9)
         liftIO $ compls @?= [keywordItem "newtype"]
+    , testSessionWait "type context" $ do
+        let source = T.unlines
+                [ "{-# OPTIONS_GHC -Wunused-binds #-}"
+                , "module A () where"
+                , "f = f"
+                ]
+        docId <- openDoc' "A.hs" "haskell" source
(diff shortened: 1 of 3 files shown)


test/exe/Main.hs
```diff
@@ -1630,6 +1630,27 @@ completionTests
...
+                   , "module A () where"
+                   , "f = f"
+                   , "g :: Intege"
+                   ]
+             ]
+        -- At this point the module parses but does not typecheck.
+        -- This should be sufficient to detect that we are in a
+        -- type context and only show the completion to the type.
+        compls <- getCompletions docId (Position 3 11)
+        liftIO $ map dropDocs compls @?= [complItem "Integer"(Just CiStruct) (Just "*")]
```
cocreature: Before this PR we _only_ produces `fromInteger` and `toInteger` here but not `Integer`.

PR #497, opened 2020-03-20
Treat alex the same way as happy to avoid CI issues

cocreature:


jneira:
> 
> 
> No idea why that would fix it but it does not seem harmful, so 👍.

I dont know it for sure (i added the workaround) but i guess it could be related with parallel builds (it is an intermittent error). However it is better than set `-j1`

PR #499, opened 2020-03-21
Expose codeAction and codeLens providers for haskell-language-server

alanz:
Also tweak the code action reply type to generate well-formed JSON

### DIFF
--- a/src/Development/IDE/Plugin.hs
+++ b/src/Development/IDE/Plugin.hs
@@ -26,12 +26,12 @@ instance Monoid (Plugin c) where
     mempty = def
 
 
-codeActionPlugin :: (LSP.LspFuncs c -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO (Either ResponseError [CAResult])) -> Plugin c
+codeActionPlugin :: (LSP.LspFuncs c -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO (Either ResponseError (List CAResult))) -> Plugin c
 codeActionPlugin = codeActionPluginWithRules mempty
 
-codeActionPluginWithRules :: Rules () -> (LSP.LspFuncs c -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO (Either ResponseError [CAResult])) -> Plugin c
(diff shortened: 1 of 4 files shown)


src/Development/IDE/Plugin.hs
```diff
@@ -26,12 +26,12 @@ instance Monoid (Plugin c) where
...
 codeActionPlugin = codeActionPluginWithRules mempty
 
-codeActionPluginWithRules :: Rules () -> (LSP.LspFuncs c -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO (Either ResponseError [CAResult])) -> Plugin c
+codeActionPluginWithRules :: Rules () -> (LSP.LspFuncs c -> IdeState -> TextDocumentIdentifier -> Range -> CodeActionContext -> IO (Either ResponseError (List CAResult))) -> Plugin c
 codeActionPluginWithRules rr f = Plugin rr $ PartialHandlers $ \WithMessage{..} x -> return x{
     LSP.codeActionHandler = withResponse RspCodeAction g
     }
     where
-      g lsp state (CodeActionParams a b c _) = fmap List <$> f lsp state a b c
+      g lsp state (CodeActionParams a b c _) = f lsp state a b c
```
cocreature: I don’t understand why we need `List`. The `ToJSON` instance of `List` is exactly the same as the one of `[]`. What is different is the `FromJSON` instance but that doesn’t matter here since this is a response. What am I missing?

alanz: See the haddocks for [List](http://hackage.haskell.org/package/haskell-lsp-types-0.21.0.0/docs/Language-Haskell-LSP-Types.html#t:List).

> This data type is used to host a FromJSON instance for the encoding used by elisp, where an empty list shows up as "null"

It is about encoding an empty list in a way that emacs is happy.

cocreature: Even the docs say that this is about `FromJSON` not about `ToJSON` which is what matters here?

alanz: Possibly, but at the end of the day I prefer to comply with the type signatures as per the underlying `haskell-lsp` library.  Which uses those everywhere to avoid JSON problems.

cocreature: Also we used `List` before as well but now you are forcing every plugin definition to take care of the wrapping which doesn’t seem helpful.

alanz: Well, if you feel strongly I can rework it.

cocreature: I would prefer the previous state, thanks!

PR #500, opened 2020-03-21
Add a custom prefix to command IDs

alanz:
A client can run more than one instance of ghcide.  Some clients have a global
command registry, and all commands must be unique in that registry.

So to make the command ids unique, prefix them with the ghcide server process
id, as is done in haskell-ide-engine.

### DIFF
--- a/src/Development/IDE/Compat.hs
+++ b/src/Development/IDE/Compat.hs
@@ -1,0 +1,19 @@
+{-# LANGUAGE CPP #-}
+module Development.IDE.Compat
+    (
+        getProcessID
+    ) where
+
+#ifdef mingw32_HOST_OS
+
+import qualified System.Win32.Process as P (getCurrentProcessId)
+getProcessID :: IO Int
+getProcessID = fromIntegral <$> P.getCurrentProcessId
+
+#else
+
+import qualified System.Posix.Process as P (getProcessID)
+getProcessID :: IO Int
+getProcessID = fromIntegral <$> P.getProcessID
+
+#endif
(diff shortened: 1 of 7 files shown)


alanz:
And the code action needs to supply the prefix too. Update needed. Tomorrow.

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -84,13 +84,14 @@ codeLens
     -> CodeLensParams
     -> IO (Either ResponseError (List CodeLens))
 codeLens _lsp ideState CodeLensParams{_textDocument=TextDocumentIdentifier uri} = do
+    pid <- getPid
```
cocreature: It seems fragile that we are adding the prefix in multiple places independently (e.g. here and in `main`). How about, we add a field to `IdeOptions` with the prefix and then fill that in in `main` and then have something like `prefixCommand :: IdeOptions -> Text -> Text` that we use here and in `main`?

alanz: I did not put it in IdeOptions, as we do not have those whole creating the prefixed command in Main.hs.  But the pid is stable, so wrapped it in a function used in the two places.

alanz:
And hlint fails it because the pid detection module has CPP in it.  To me it does not make sense to put it in the GHC/Compat.hs module.

what is the best way to resolve this? platform-specific source directories?

ndmitchell:
@alanz If there's a good reason to put CPP in the code, put it in and add an exception in `.hlint.yaml`. It's meant to reduce the CPP, and make it hard to sneak in, not eliminate it. Platform specific directories would be way worse.

src/Development/IDE/Compat.hs
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE CPP #-}
+module Development.IDE.Compat
+    (
+        getProcessID
+    ) where
+
+#ifdef mingw32_HOST_OS
+
+import qualified System.Win32.Process as P (getCurrentProcessId)
```
cocreature: This import is failing since you did not add a dependency on `Win32`.

jneira: happy to see windows azure ci in action! :smile_cat:

PR #502, opened 2020-03-23
Fix emptyPathUri

cocreature:


src/Development/IDE/Types/Location.hs
```diff
@@ -53,7 +55,13 @@ uriToFilePath' uri
     | otherwise = LSP.uriToFilePath uri
 
 emptyPathUri :: LSP.NormalizedUri
-emptyPathUri = LSP.NormalizedUri (hash ("" :: String)) ""
+emptyPathUri =
+    -- The difference here matches the behavior of platformAwareFilePathToUri.
+    -- haskell-lsp does not expose that directy (or only with a warning) so we
+    -- just inline the definition here.
+    let s | isWindows = "file:///"
```
cocreature: I believe we can make this non-platform dependent, verifying atm.

jneira:
Jumm, sorry for not catching that

PR #504, opened 2020-03-24
Expose an option to control the number of Shake threads

pepeiborra:
/cc @mpickering

### DIFF
--- a/exe/Arguments.hs
+++ b/exe/Arguments.hs
@@ -13,6 +13,7 @@ data Arguments = Arguments
     ,argsVersion :: Bool
     ,argsShakeProfiling :: Maybe FilePath
     ,argsTesting :: Bool
+    ,argsThreads :: Int
     }
 
 getArguments :: IO Arguments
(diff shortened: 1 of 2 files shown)


hie.yaml
```diff
@@ -1 +1,3 @@
-cradle: {stack: {component: "ghcide:lib"}}
+# cradle: {cabal: {component: "ghcide:exe:ghcide"}}
+# cradle: {cabal: {component: "ghcide:ghcide-tests"}}
+cradle: {cabal: {component: "ghcide"}}
```
mpickering: Accident.

test/exe/Main.hs
```diff
@@ -2094,7 +2094,7 @@ runInDir dir s = do
       -- which can be quite useful for debugging.
       -- { logStdErr = True, logColor = False }
       -- If you really want to, you can also see all messages
-      -- { logMessages = True, logColor = False }
+      { logMessages = True, logColor = True}
```
mpickering: Accident?

hie.yaml
pepeiborra: oops, fixed

test/exe/Main.hs
pepeiborra: fixed too

PR #516, opened 2020-03-29
Track dependencies when using qAddDependentFile

mpickering:
Do I also need to `need` the files or is calling `doesFileExist` enough to track the dependency?

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -72,6 +72,7 @@ main = defaultMainWithRerun $ testGroup "HIE"
   , positionMappingTests
   , watchedFilesTests
   , cradleTests
+  , dependentFileTest
   ]
 
 initializeResponseTests :: TestTree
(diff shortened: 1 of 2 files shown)


ndmitchell:
`doesFileExist` tracks that the file exists or not. `need` tracks the modification time too and asserts that it must exist. My guess is you want `need` here?

pepeiborra:
`need` uses the plain Shake file rules, wouldn't `GetModificationTime` be a better fit here?

mpickering:
@pepeiborra What is your exact suggestion?

pepeiborra:
`use_ GetModificationTime` instead of `doesFileExist`

ndmitchell:
Yes, you're right. A `GetModificationTime` would be better.

mpickering:
Is this what you have in mind?

pepeiborra:
> Is this what you have in mind?

Yes, this is exactly what I had in mind

mpickering:
How would a test work?

pepeiborra:
Check that a program using `qAddDependentFile` to e.g. inject code, gets recompiled after changing the indirect dep?

cocreature:
a very simple way to check that it gets recompiled might be to just delete the file that you are adding via `qAddDependentFile` and see that you get a diagnostic afterwards.

mpickering:
I added a test now.

test/exe/Main.hs
```diff
@@ -2009,6 +2010,42 @@ loadCradleOnlyonce = testGroup "load cradle only once"
...
+              , "module Foo where"
+              , "import Language.Haskell.TH.Syntax"
+              , "foo :: Int"
+              , "foo = 1 + $(do"
+              , "               qAddDependentFile \"dep-file.txt\""
+              , "               f <- qRunIO (readFile \"dep-file.txt\")"
+              , "               if f == \"B\" then [| 1 |] else lift f)"
+              ]
+        let bazContent = T.unlines ["module Baz where", "import Foo"]
+        _ <- openDoc' "Foo.hs" "haskell" fooContent
```
cocreature, line 2033:
```suggestion
        _ <-createDoc "Foo.hs" "haskell" fooContent
```

test/exe/Main.hs
```diff
@@ -2009,6 +2010,42 @@ loadCradleOnlyonce = testGroup "load cradle only once"
...
+        doc <- openDoc' "Baz.hs" "haskell" bazContent
```
cocreature, line 2034:
```suggestion
        doc <- createDoc "Baz.hs" "haskell" bazContent
```

Issue #494, opened 2020-03-20
Call to mask cradle initialisation is probably ineffective


mpickering:
Pepe inserted a call to mask the cradle initialisation recently but I don't think it will work as expected because the cradle initialisation will call functions from `System.Process` which I think are interruptible. 

https://github.com/digital-asset/ghcide/blob/master/exe/Rules.hs#L80

On the multi-cradle branch I have certainly observed async exceptions bringing down this code path, which were "fixed" by using `uninterruptibleMask` rather than `mask`.

Can you comment on what the right fix for this is @ndmitchell ?

cc @pepeiborra

ndmitchell:
My experience is using mask is nearly always wrong, and using mask or any related functions correctly is nearly impossible. I'm somewhat disheartened with async exceptions entirely!

pepeiborra:
Have you run into the error described [here](https://github.com/digital-asset/ghcide/pull/471#issuecomment-596231636) after my change? 

The argument against `uninterruptibleMask` is that it will probably make it impossible to kill ghcide. So what do you do if the wrong cradle is picked up and `stack` starts compiling the universe?

mpickering:
Yes it is exactly that issue that led me to investigate this.

mpickering:
Using `mask` to try to fix the hadrian error that Pepe mentions above is not really the correct fix. The error arises because hadrian does not like being invoked in quite succession, it is already not called concurrently because there is a lock guarding this part of the code. The `uninterruptibleMask` call stops it being called in close succession as the first thread is never killed. 

So I can think of two possible ways of fixing this:

1. Create a specific thread where we do cradle initialisation, which is never killed. 
2. Make sure there is at least 0.1s between calls to initialise a cradle, so the cradle is not called and killed in quick succession.

ndmitchell:
The former of those seems most appropriate. We'd also want things to wait on a previously started but not yet finished cradle init rather than starting a new one.

pepeiborra:
Worth considered a third option, fix this in hie-bios by intercepting asyncronous exceptions and waiting for clean up before rethrowing

ndmitchell:
Even if hie-bios could intercept the exception and roll back everything, I don't think we'd want to - from ghcide it makes sense to let hie-bios actions that have already started run to completion since we know the chances of them being required is super high and the cost of aborting midway through is also high.

mpickering:
What clean up could be observed? 

[...]

Issue #488, opened 2020-03-18
ghcide does not deal with relative source directories well


mpickering:
For a long time ghcide has been broken when working on GHC because hadrian was providing relative source directories (`-i` flags to ghcide). This leads to a bunch of confusing errors about paths not existing and so on. Today I tried to fix it by providing absolute paths and it made things work much better.

`ghcide` should probably canonicalise all paths and make sure to only use absolute paths.

cocreature:
We use `ghcide` also in the DAML CLI compiler where we only have relative paths to ensure that build outputs are deterministic and independent of the working directory. So just making everything absolute isn’t going to work.

For paths that we get from LSP, paths should be absolute anyway so I don’t think we have to do anything about those. The only case where we do create paths ourselves as far as I know is when locating a module so maybe we can just add a flag to `IdeOptions` whether those need to be made absolute or not.

aherrmann-da:
Pulling the relevant comments out of https://github.com/digital-asset/ghcide/pull/522.

`rules_haskell` (Bazel) is hitting the same issue with relative import dirs (`-i`). E.g. go-to-definition tries to open relative paths as absolute paths, i.e. `/tests/library-deps/sublib/TestSubLib.hs` instead of `tests/library-deps/sublib/TestSubLib.hs`.

@mpickering suggested
> You can probably also fix the issue with go-to definition by using absolute paths for the include directories (`-i` flags), I had the same issue with hadrian and that fixed it. (See #488)

Sadly, for `rules_haskell` that's not a great option. Bazel doesn't give access to absolute paths (barring hacky workarounds), and for good reason: Introducing absolute paths would be inhermetic and break distributed caching, e.g. it would leak `$HOME` into the build. One could try to work around this by patching `-i` paths in the bios script to be converted to absolute paths, but that introduces issues around escaping and is a can of worms I'd prefer to leave closed. Fixing this in `ghcide` seems like the better solution.

jneira:
is this still reproduced with last ghcide?

PR #522, opened 2020-04-06
Multi Component

mpickering:
In this commit we add support for loading multiple components into one
ghcide session.

The current behaviour is that each component is loaded lazily into the
session. When a file from an unrecognised component is loaded, the
cradle is consulted again to get a new set of options for the new
component. This will cause all the currently loaded files to be
reloaded into a new HscEnv which is shared by all the currently known
components. The result of this is that functions such as go-to
definition work between components if they have been loaded into the
same session but you have to open at least one file from each component
before it will work.

Only minimal changes are needed to the internals to ghcide to make the
file searching logic look in include directories for all currently
loaded components. The main changes are in exe/Main.hs which has been
heavily rewritten to avoid shake indirections. A global map is created
which maps a filepath to the HscEnv which should be used to compile it.
When a new component is created this map is completely refreshed so each
path maps to a new

Which paths belong to a componenent is determined by the targets listed
by the cradle. Therefore it is important that each cradle also lists all
the targets for the cradle. There are some other choices here as well
which are less accurate such as mapping via include directories  which
is the aproach that I implemented in haskell-ide-engine.

The commit has been tested so far with cabal and hadrian.

Also deleted the .ghci file which was causing errors during testing and
seemed broken anyway.


runeksvendsen:
I seem to experience an intermittent error using this branch of `ghcide` in a project of mine, in which a test and an executable depend on the library. The error is that the library in question ( `rule-lang`) cannot be found:

```
ghcide: <command line>: cannot satisfy -package rule-lang-0.1.0.0
```

After outputting the above `ghcide` hangs (i.e. the executable never exits, but simply hangs after printing the above line).

I made a commit of the code that exhibits the above symptoms here: https://github.com/runeksvendsen/rule-lang/commit/6bdea22bce58e1efa1c8ae4f60a3e28ebac35d6b.

I'm wondering if there's some non-determinism in the way multiple components are loaded, such that if an executable or test (which depend on the library) is loaded before the library then ghcide will complain that the library can't be found.

Full log output of running `ghcide` executable :

```
ghcide version: 0.1.0 (GHC: 8.6.5) (PATH: /Users/runesvendsen/.local/bin/ghcide) (GIT hash: c43a078c829be65d002c6ebcd3b1c9f2f534ca43)
Ghcide setup tester in /Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs.
Report bugs at https://github.com/digital-asset/ghcide/issues

Step 1/6: Finding files to test in /Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs
Found 17 files

Step 2/6: Looking for hie.yaml files that control setup
Found 1 cradle

Step 3/6: Initializing the IDE

Step 4/6: Type checking the files
Shelling out to cabal "/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/src/Absyn.hs"
> rule-lang> initial-build-steps (lib + exe)
> Configuring GHCi with the following packages: rule-lang
> /Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/install/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb:/Users/runesvendsen/.stack/snapshots/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb:/Users/runesvendsen/.stack/programs/x86_64-osx/ghc-8.6.5/lib/ghc-8.6.5/package.conf.d
ComponentOptions {componentOptions = ["-i","-odir=/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/odir","-hidir=/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/odir","-hide-all-packages","-XOverloadedStrings","-XStandaloneDeriving","-XDeriveGeneric","-XNoImplicitPrelude","-XFlexibleContexts","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/src","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/autogen","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/global-autogen","-stubdir=/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build","-package-id=aeson-1.4.4.0-8dIyFiOKltcCLD7t2J0OHI","-package-id=base-4.12.0.0","-package-id=containers-0.6.0.1","-package-id=errors-2.3.0-GKWdqDKmaPrGy1XArjac2y","-package-id=hashable-1.2.7.0-2SI038axTEd7AEZJ275kpi","-package-id=megaparsec-7.0.5-9VrJ76Lt7nJ5Iji4PA7A8S","-package-id=parsers-0.12.10-GWrQJtO53B83vKycLKDMVC","-package-id=protolude-0.2.3-6EKDEuuh3ZRFm0XjK4luPw","-package-id=text-1.2.3.1","-package-id=transformers-0.5.6.2","-package-id=unordered-containers-0.2.10.0-LgoTL3wbBEY5bZIDJiyxW4","-Wall","-optP-include","-optP/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/ghci/168fd77d/cabal_macros.h","-ghci-script=/private/var/folders/dm/d2y4x8mx50n__pwvsmpj_j_r0000gn/T/haskell-stack-ghci/fda1367d/ghci-script","-package-db","/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/install/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb","-package-db","/Users/runesvendsen/.stack/snapshots/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb","-package-db","/Users/runesvendsen/.stack/programs/x86_64-osx/ghc-8.6.5/lib/ghc-8.6.5/package.conf.d"], componentDependencies = ["rule-lang.cabal","package.yaml","stack.yaml"]}
"Making new HscEnv[main]"
(HscEnvEq 1,fromList [("package.yaml",Just 2020-03-31 08:05:45.386518558 UTC),("rule-lang.cabal",Just 2020-04-07 08:50:32.00053941 UTC),("stack.yaml",Just 2020-03-17 16:08:13.409014676 UTC)])
Shelling out to cabal "/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/test/Spec/Parse.hs"
> Using main module: 1. Package `rule-lang' component rule-lang:test:rule-lang-test with main-is file: /Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/test/Spec.hs
> rule-lang> initial-build-steps (lib + exe)
> The following GHC options are incompatible with GHCi and have not been passed to it: -O2
> Configuring GHCi with the following packages: rule-lang
> /Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/install/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb:/Users/runesvendsen/.stack/snapshots/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb:/Users/runesvendsen/.stack/programs/x86_64-osx/ghc-8.6.5/lib/ghc-8.6.5/package.conf.d
ComponentOptions {componentOptions = ["-i","-odir=/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/odir","-hidir=/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/odir","-hide-all-packages","-XOverloadedStrings","-XStandaloneDeriving","-XDeriveGeneric","-XNoImplicitPrelude","-XFlexibleContexts","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/rule-lang-test","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/test","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/rule-lang-test/autogen","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/global-autogen","-i/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build/rule-lang-test/rule-lang-test-tmp","-stubdir=/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/dist/x86_64-osx/Cabal-2.4.0.1/build","-package-id=base-4.12.0.0","-package-id=hspec-2.7.1-BxmpM3H5M4w4Y2ViaUKRtb","-package-id=hspec-expectations-pretty-diff-0.7.2.4-DNpNgJ1MOyu63xg9pNKuPu","-package-id=hspec-smallcheck-0.5.2-7Iq7U1gcFRNBmJABwPLv5E","-package-id=megaparsec-7.0.5-9VrJ76Lt7nJ5Iji4PA7A8S","-package=rule-lang-0.1.0.0","-package-id=smallcheck-1.1.5-8zXm09BFgY5BYxznod0SEL","-package-id=tasty-1.2.3-hTb5MwkGozC98ZFlavggW","-package-id=tasty-smallcheck-0.8.1-1OLucf7iboX6WjE9zcAsBp","-package-id=text-1.2.3.1","-Wall","-optP-include","-optP/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/ghci/33d2892b/cabal_macros.h","-ghci-script=/private/var/folders/dm/d2y4x8mx50n__pwvsmpj_j_r0000gn/T/haskell-stack-ghci/29833fcc/ghci-script","-package-db","/Users/runesvendsen/Documents/ITU/06BSc/code/RuleLangHs/.stack-work/install/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb","-package-db","/Users/runesvendsen/.stack/snapshots/x86_64-osx/94377dc6990ad6b243b760bbfb93aab30063cf768abe840214a74b980aac577e/8.6.5/pkgdb","-package-db","/Users/runesvendsen/.stack/programs/x86_64-osx/ghc-8.6.5/lib/ghc-8.6.5/package.conf.d"], componentDependencies = ["rule-lang.cabal","package.yaml","stack.yaml"]}
ghcide: <command line>: cannot satisfy -package rule-lang-0.1.0.0
    (use -v for more information)
^C
```

mpickering:
@runeksvendsen I tested your branch using cabal and it worked for me (after I fixed the build errors in the library component).

mpickering:
I noticed the logic to ensure that multiple initialisations do not happen concurrently is not currently correct.

runeksvendsen:
Thanks @mpickering. I figure the problem appeared when I edit the library component, so that it doesn't compile, and then go to a file in the executable target, which would result in VS Code hanging with the status "Processing x/y...", since ghcide didn't exit. However, I can't reproduce this now with the aforementioned commit.

mpickering:
> What is the right thing to do with autogen Cabal modules?
> I got an error on the ghcide workspace (using this branch):
> 
> ```
> Shelling out to cabal "/home/pepe/scratch/ghcide/dist-newstyle/build/x86_64-linux/ghc-8.8.3/ghcide-0.1.0/build/autogen/Paths_ghcide.hs"
> ghcide: CradleError ExitSuccess ["Multi Cradle: No prefixes matched","pwd: /home/pepe/scratch/ghcide","filepath: /home/pepe/scratch/ghcide/dist-newstyle/build/x86_64-linux/ghc-8.8.3/ghcide-0.1.0/build/autogen/Paths_ghcide.hs","prefixes:","(\"./src\",Cabal {component = Just \"ghcide:lib:ghcide\"})","(\"./exe\",Cabal {component = Just \"ghcide:exe:ghcide\"})","(\"./test\",Cabal {component = Just \"ghcide:test:ghcide-tests\"})","(\"./test/preprocessor\",Cabal {component = Just \"ghcide:exe:ghcide-test-preprocessor\"})"]
> ```

How do you trigger that error? Did you open that file in an editor? What I would expect to happen is that if you previously loaded any other file from the component then it should work. Otherwise it is a limitation of the manual multi-cradle that you are manually specifying the directories.

pepeiborra:
> > What is the right thing to do with autogen Cabal modules?
> > I got an error on the ghcide workspace (using this branch):
> > ```
> > Shelling out to cabal "/home/pepe/scratch/ghcide/dist-newstyle/build/x86_64-linux/ghc-8.8.3/ghcide-0.1.0/build/autogen/Paths_ghcide.hs"
> > ghcide: CradleError ExitSuccess ["Multi Cradle: No prefixes matched","pwd: /home/pepe/scratch/ghcide","filepath: /home/pepe/scratch/ghcide/dist-newstyle/build/x86_64-linux/ghc-8.8.3/ghcide-0.1.0/build/autogen/Paths_ghcide.hs","prefixes:","(\"./src\",Cabal {component = Just \"ghcide:lib:ghcide\"})","(\"./exe\",Cabal {component = Just \"ghcide:exe:ghcide\"})","(\"./test\",Cabal {component = Just \"ghcide:test:ghcide-tests\"})","(\"./test/preprocessor\",Cabal {component = Just \"ghcide:exe:ghcide-test-preprocessor\"})"]
> > ```
> 
> How do you trigger that error? Did you open that file in an editor? What I would expect to happen is that if you previously loaded any other file from the component then it should work. Otherwise it is a limitation of the manual multi-cradle that you are manually specifying the directories.

I did not open that file directly. Was trying to get a hover in `exe/Main.hs` after having loaded some modules from the library component.

exe/Main.hs
```diff
@@ -188,22 +190,235 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+                -- fork a new thread here which won't be killed by shake
+                -- throwing an async exception
+                void $ forkIO $ withLock cradle_lock $ do
+                  putStrLn $ "Shelling out to cabal " <> show file
+                  cradle <- maybe (loadImplicitCradle $ addTrailingPathSeparator dir) loadCradle hieYaml
+                  opts <- cradleToSessionOpts cradle cfp
+                  print opts
+                  res <- fst <$> session (hieYaml, toNormalizedFilePath' cfp, opts)
+                  signalBarrier finished_barrier res
+                waitBarrier finished_barrier
```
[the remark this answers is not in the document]
mpickering: After further testing I don't think this code works as I intended at all. Need to rework it.

aherrmann-da:
I tested this in combination with `rules_haskell` and Bazel. (I'm working on ghcide/hie-bios support for `rules_haskell`, see https://github.com/tweag/rules_haskell/pull/1262)

### Reproduce

* Checkout https://github.com/tweag/rules_haskell/tree/ghcide-multi (at 3f9395be3d8c119962023236043cb0ae0b39aef4 at the time of writing).
* Configure it for Nix
  - Linux:
    ```
    $ echo "build --host_platform=@rules_haskell//haskell/platforms:linux_x86_64_nixpkgs" >> .bazelrc.local
    ```
  - MacOS:
    ```
    $ echo "build --host_platform=@rules_haskell//haskell/platforms:darwin_x86_64_nixpkgs" >> .bazelrc.local
    ```
* Enter a Nix shell
  - Linux:
    ```
    $ nix-shell
    ```
  - MacOS:
    ```
    $ nix-shell --arg docTools false
    ```
* Then invoke `ghcide` on different files (first invocation is going to take a while to build ghcide)
  ```
  $ ./.ghcide tests/RunTests.hs
  $ ./.ghcide tests/library-deps/TestLib.hs
  $ ./.ghcide tests/binary-indirect-cbits/Main.hs
  ```
* For context, `hie.yaml` looks like this:
  ```
  cradle:
    multi:
      - path: "tests"
        config: {cradle: {bios: {program: ".hie-bios"}}}
      - path: "tests/binary-indirect-cbits"
        config: {cradle: {bios: {program: "tests/binary-indirect-cbits/.hie-bios"}}}
      - path: "tests/library-deps"
        config: {cradle: {bios: {program: "tests/library-deps/.hie-bios"}}}
  ```

### Observations

The first two seem to be working fine. Also opening these files in an editor gives expected IDE features, like hover and goto definition. (There's an issue with goto-definition across files where it tries to open files under an absolute path, e.g. `/tests/library-deps/sublib/TestSubLib.hs` instead of `tests/library-deps/sublib/TestSubLib.hs`, but AFAIK that's not related to multi-cradle support).

However, the third one freezes after printing the following to the terminal:
```
Shelling out to cabal "/home/aj/tweag.io/da/bazel-projects/rules_haskell_ghcide_multi/tests/binary-indirect-cbits/Main.hs"
ComponentOptions {componentOptions = ["-package-id","base-4.12.0.0","-ltests_Sdata_Slibourclibrary","-Lbazel-out/k8-fastbuild/bin/_solib_k8","-pgma","bazel-out/host/bin/haskell/cc_wrapper-python","-pgmc","bazel-out/host/bin/haskell/cc_wrapper-python","-pgml","bazel-out/host/bin/haskell/cc_wrapper-python","-pgmP","bazel-out/host/bin/haskell/cc_wrapper-python","-optc-fno-stack-protector","-optP-E","-optP-undef","-optP-traditional","-itests/binary-indirect-cbits/","-ibazel-out/k8-fastbuild/bin/tests/binary-indirect-cbits/","-itests/library-with-cbits/","-ibazel-out/k8-fastbuild/bin/tests/library-with-cbits/","-ibazel-out/k8-fastbuild/bin/tests/library-with-cbits/_hsc/library-with-cbits","-XStandaloneDeriving","-threaded","-DTESTS_TOOLCHAIN_COMPILER_FLAGS","-XNoOverloadedStrings"], componentDependencies = []}
"Making new HscEnv[main]"
(HscEnvEq 1,fromList [])
Shelling out to cabal "bazel-out/k8-fastbuild/bin/tests/library-with-cbits/_hsc/library-with-cbits/AddOne.hs"
ghcide: CradleError ExitSuccess ["Multi Cradle: No prefixes matched","pwd: /home/aj/tweag.io/da/bazel-projects/rules_haskell_ghcide_multi","filepath: /home/aj/.cache/bazel/_bazel_aj/0a879060aafa76d7881430a3c48b06d5/execroot/rules_haskell/bazel-out/k8-fastbuild/bin/tests/library-with-cbits/_hsc/library-with-cbits/AddOne.hs","prefixes:","(\"tests\",Bios {prog = \".hie-bios\", depsProg = Nothing})","(\"tests/binary-indirect-cbits\",Bios {prog = \"tests/binary-indirect-cbits/.hie-bios\", depsProg = Nothing})","(\"tests/library-deps\",Bios {prog = \"tests/library-deps/.hie-bios\", depsProg = Nothing})"]
```
The trouble seems to be that the third one depends on `AddOne.hs`, which is generated by `rules_haskell` from `AddOne.hsc`. Generated files are placed under a different prefix, here `~/.cache/.../k8-fastbuild/.../tests/library-with-cbits/_hsc`, so it seems they are not   associated with the same cradle. I can work around this by adding another cradle that points to the same `.hie-bios` for the generated files.
```
cradle:
  multi:
    - path: "bazel-bin/tests/library-with-cbits"
      config: {cradle: {bios: {program: "tests/binary-indirect-cbits/.hie-bios"}}}
```
Fortunately, this works when using the convenience symlink `bazel-bin` that Bazel generates. However, it is unfortunate that such a workaround is necessary.

### Issues

* I found that ghcide prints `Shelling out to cabal` which is not accurate in a bios cradle like these here.
* I don't think ghcide should hang as it does with the generated `AddOne.hs` described above.
* The workaround for the generated files is not unbearable, but it is annoying and adding boilerplate. Do you have an idea of how this may be fixed? I assume e.g. Cabal would have similar issues with generated files that end up under `dist` or similar. (This seems similar to what @pepeiborra is encountering above)

mpickering:
@aherrmann Is there a reason you are using the multi-cradle AND bios cradle together? You should ideally modify your bios script so that given a filepath it returns the set of options for the component which contains that file. 

Just so we are clear here, the multi-cradle is intended to patch over the fact that build tools do not in general support querying for a component based on filepaths. 

It is also possible your cradle is not listing all the targets for the component it is trying to load. If it doesn't then it will not work with this branch.

Also the specific behaviour that I have implemented about what should/shouldn't happen in the following situations is up for discussion. What I have implemented in this patch are the foundations which mean making these other decision is easier. 

* Which components do we load on initialisation? (Currently none)
* When do we load new components? (When the file is opened for the first time)
* What happens when a file does not match any component (There is an error, improving on this is related to implementing support for the `none` cradle I believe)
* What happens when a file matches no existing components? (The cradle is consulted again and all existing HscEnvs are replaced with a new "super" HscEnv which combines together all the currently known components)
* How do we decide which files belong to an already loaded component? (We create all the possible paths by combining the targets and include directories as returned by the cradle)

Ultimately it is not just up to me to decide on these things and to also implement them. The branch is working for my purposes working on GHC and with cabal projects so it will take a wider community effort to bring it to a state where it can be merged.

mpickering:
@aherrmann-da I looked at your PR a little and it seems the `.hie-bios` script does not take the filepath into account when working out the correct options for the component. If you can add a mapping from path to component and then make the script return the right options, things should work a lot better. 

You can probably also fix the issue with go-to definition by using absolute paths for the include directories (`-i` flags), I had the same issue with hadrian and that fixed it. (See #488)

aherrmann-da:
> @aherrmann Is there a reason you are using the multi-cradle AND bios cradle together? You should ideally modify your bios script so that given a filepath it returns the set of options for the component which contains that file.

That's a very useful feature (that the filepath is passed to the bios script). I was not aware that this PR also introduced it. Note, this is not available as of `ghcide` `master` (39605333c34039241768a1809024c739df3fb2bd). I've verified this by echoing `$@` into a log file in `.hie-bios`. I didn't see this feature mentioned in this PR's description, did I miss it somewhere?

> It is also possible your cradle is not listing all the targets for the component it is trying to load. If it doesn't then it will not work with this branch.

Apart from the issue around the path prefix of generated files the targets are complete. Note, that the additional entry I added for the generated files points to the same `.hie-bios` script.

> @aherrmann-da I looked at your PR a little and it seems the `.hie-bios` script does not take the filepath into account when working out the correct options for the component. If you can add a mapping from path to component and then make the script return the right options, things should work a lot better.

That PR is targeting `ghcide` as of `master`. So, the `.hie-bios` script does not receive any arguments. The multi component experiments are tracked in a separate branch, see the "reproduce" section in my comment above.

I'll experiment with using the argument to `.hie-bios`. So far it's looking good, thanks!

> If you can add a mapping from path to component and then make the script return the right options, things should work a lot better. I believe this is how the first iteration of support for bazel that I wrote worked.

Yes, it was a builtin cradle (not bios) that, given a source file, queried for the corresponding `haskell_*` target, then took the corresponding `@repl` script and plucked the ghci flags out of it. The approach taken in https://github.com/tweag/rules_haskell/pull/1262 is aiming to be more flexible. `rules_haskell` has since gained the [`haskell_repl`](https://api.haskell.build/haskell/defs.html#haskell_repl) rule, which allows to define a REPL for multiple `haskell_*` targets. Meaning if you have interdependent targets `:lib-a <-- :lib-b <-- :exe` then you can load them all into the same ghci session by source, instead of having to compile say `:lib-a` and `:lib-b` to load `:exe`. We can use this for `ghcide` to load multiple targets into the same session.

However, this means there is no longer a 1:1 mapping from source to REPL target, as a user might have multiple REPL targets (indirectly) including the same source file. That made multi cradle look like an obvious candidate to let the user define this mapping. Defining that mapping in `.hie-bios` is an alternative option.

mpickering:
> That's a very useful feature (that the filepath is passed to the bios script). I was not aware that this PR also introduced it. Note, this is not available as of `ghcide` `master` ([3960533](https://github.com/digital-asset/ghcide/commit/39605333c34039241768a1809024c739df3fb2bd)). I've verified this by echoing `$@` into a log file in `.hie-bios`. I didn't see this feature mentioned in this PR's description, did I miss it somewhere?

All cradles have had this interface since the earliest iteration of `hie-bios`. `ghcide` has never supported it and always passed an empty string when querying a cradle to get the options. You are right that I can't see that is is documented in the hie-bios README. 

You are right there is an issue about deciding how to map a FilePath to a component, but that is for the author of the build tool to determine, not ghcide.

exe/Main.hs
```diff
@@ -188,22 +190,235 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+    mapM (fmap toNormalizedFilePath' . canonicalizePath) fps
+targetToFile _ (TargetFile f _) = do
+  f' <- canonicalizePath f
+  return [(toNormalizedFilePath' f')]
+
+setNameCache :: IORef NameCache -> HscEnv -> HscEnv
+setNameCache nc hsc = hsc { hsc_NC = nc }
+
+-- This is the key function which implements multi-component support. All
+-- components mapping to the same hie,yaml file are mapped to the same
```
jneira, line 231: typo
```suggestion
-- components mapping to the same hie.yaml file are mapped to the same
```

fendor:
I am using this branch as my daily driver for about two weeks now. Except for the occasional GHC bug, #529, my expierence has been vastly positive.
Memory usage is good so far, performance is good and even automatic reload if dependencies change works.
My experience is only based on cabal so far.

pepeiborra:
This needs rebasing.

mpickering:
I have some more fixes on the staging branch and waiting for one PR to be ready.

exe/Main.hs
```diff
@@ -44,14 +50,38 @@ import qualified System.Directory.Extra as IO
...
-import Development.Shake (Action, Rules, action)
+import Development.Shake (Action,  action)
 import qualified Data.HashSet as HashSet
+import qualified Data.HashMap.Strict as HM
 import qualified Data.Map.Strict as Map
-import HIE.Bios
-import Rules
-import RuleTypes
+import Data.Either
+--import Outputable (pprTraceM, ppr, text)
```
cocreature, line 61:
```suggestion
```

exe/Main.hs
```diff
@@ -44,14 +50,38 @@ import qualified System.Directory.Extra as IO
...
+import qualified Crypto.Hash.SHA1               as H
+import qualified Data.ByteString.Char8          as B
+import           Data.ByteString.Base16         (encode)
```
cocreature, line 64: Please don’t align imports. It just makes diffs more noisy.

exe/Main.hs
```diff
@@ -44,14 +50,38 @@ import qualified System.Directory.Extra as IO
...
+
+import           HIE.Bios.Cradle
+import           HIE.Bios.Types
+import System.Directory
+
+import Utils
+
+--import Rules
+--import RuleTypes
+--
```
cocreature, line 84:
```suggestion
```

exe/Main.hs
```diff
@@ -124,44 +155,17 @@ main = do
...
-
-cradleRules :: Rules ()
-cradleRules = do
-    loadGhcSession
-    cradleToSession
+        putStrLn "\nStep 4/6: Type checking the files"
+        setFilesOfInterest ide $ HashSet.fromList $ map toNormalizedFilePath' files
+        _ <- runActionSync ide $ uses TypeCheck (map toNormalizedFilePath' files)
+--        results <- runActionSync ide $ use TypeCheck $ toNormalizedFilePath' "src/Development/IDE/Core/Rules.hs"
+--        results <- runActionSync ide $ use TypeCheck $ toNormalizedFilePath' "exe/Main.hs"
```
cocreature, line 167:
```suggestion
```

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
 
+    -- Create a new HscEnv from a hieYaml root and a set of options
+    -- If the hieYaml file already has an HscEnv, the new component is
+    -- combined with the components in the old HscEnv into a new HscEnv
+    -- which contains both.
+    packageSetup <- return $ \(hieYaml, cfp, opts) -> do
+        -- Parse DynFlags for the newly discovered component
+        hscEnv <- emptyHscEnv
+        (df, targets) <- evalGhcEnv hscEnv $ do
+                          setOptions opts (hsc_dflags hscEnv)
```
cocreature, line 259: The `do` seems to be redundant and the indentation is a bit weird.

exe/Main.hs
```diff
@@ -188,22 +190,235 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+        -- (unitId, DynFlag, Targets)
+        modifyVar hscEnvs $ \m -> do
+            -- Just deps if there's already an HscEnv
+            -- Nothing is it's the first time we are making an HscEnv
+            let oldDeps = Map.lookup hieYaml m
+            let -- Add the raw information about this component to the list
+                -- We will modify the unitId and DynFlags used for
+                -- compilation but these are the true source of
+                -- information.
+                new_deps = (thisInstalledUnitId df, df, targets, cfp, dep_info) : maybe [] snd oldDeps
```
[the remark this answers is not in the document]
cocreature: Sounds like a good idea :+1:

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+setNameCache :: IORef NameCache -> HscEnv -> HscEnv
+setNameCache nc hsc = hsc { hsc_NC = nc }
+
+-- This is the key function which implements multi-component support. All
+-- components mapping to the same hie,yaml file are mapped to the same
+-- HscEnv which is updated as new components are discovered.
 loadSession :: FilePath -> Action (FilePath -> Action HscEnvEq)
 loadSession dir = liftIO $ do
+    -- Mapping from hie.yaml file to HscEnv, one per hie.yaml file
+    hscEnvs <- newVar Map.empty
```
cocreature, line 238: Type annotations on `hscEnvs` and `fileToFlags` would probably make this a bit easier to follow.

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+                rearrange (uid, (df, uids), ts, cfp, di) = (uid, (df, uids, ts, cfp, di))
+                do_one (uid,df, ts, cfp, di) = rearrange (uid, removeInplacePackages inplace df, ts, cfp, di)
+                -- All deps, but without any packages which are also loaded
+                -- into memory
+                new_deps' = map do_one new_deps
+            -- Make a new HscEnv, we have to recompile everything from
+            -- scratch again (for now)
+            -- It's important to keep the same NameCache though for reasons
+            -- that I do not fully understand
+            print ("Making new HscEnv" ++ (show inplace))
```
cocreature, line 287: Looks like leftover debugging output?

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+            newHscEnv <-
+              -- Add the options for the current component to the HscEnv
+              evalGhcEnv hscEnv $ do
+                _ <- setSessionDynFlags df
+                getSession
+            -- Modify the map so the hieYaml now maps to the newly created
+            -- HscEnv
+            -- Returns
+            -- * the new HscEnv so it can be used to modify the
+            --   FilePath -> HscEnv map
```
cocreature, line 300:
```suggestion
            --   FilePath -> HscEnv map (fileToFlags)
```

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+            -- * The information for the new component which caused this cache miss
+            -- * The modified information (without -inplace flags) for
+            --   existing packages
+            pure (Map.insert hieYaml (newHscEnv, new_deps) m, (newHscEnv, head new_deps', tail new_deps'))
+
+
+    session <- return $ \(hieYaml, cfp, opts) -> do
+        (hscEnv, new, old_deps) <- packageSetup (hieYaml, cfp, opts)
+        -- TODO Handle the case where there is no hie.yaml
```
cocreature, line 309: What happens here atm?

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+        let new_cache (_iuid, (df, _uis, targets, cfp, di)) =  do
+              let hscEnv' = hscEnv { hsc_dflags = df
+                                   , hsc_IC = (hsc_IC hscEnv) { ic_dflags = df } }
+
+              versionMismatch <- evalGhcEnv hscEnv' checkGhcVersion
+              henv <- case versionMismatch of
+                        Just mismatch -> return mismatch
+                        Nothing -> newHscEnvEq hscEnv' uids
+              let res = (henv, di)
+              print res
```
cocreature, line 328: leftover debugging output?

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+
+              let is = importPaths df
+              ctargets <- concatMapM (targetToFile is  . targetId) targets
+              -- A special target for the file which caused this wonderful
```
cocreature, line 332: Why do we need this? Is this to guard against build tools that do not list all targets (e.g. iirc stack doesn’t list the paths modules or something like that?)? Would be great to document this in a comment.

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+              -- component to be created.
+              let special_target = (cfp, res)
+              --pprTraceM "TARGETS" (ppr (map (text . show) ctargets))
```
cocreature, line 335:
```suggestion
```

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+
+
+checkDependencyInfo :: Map.Map FilePath (Maybe UTCTime) -> IO Bool
+checkDependencyInfo old_di = do
+  di <- getDependencyInfo (Map.keys old_di)
+  return (di == old_di)
+
+
+
+getDependencyInfo :: [FilePath] -> IO (Map.Map FilePath (Maybe UTCTime))
```
cocreature, line 417: This looks like we are reimplementing file modification checking instead of piggy backing on the one we have for source files. I think the cradle setup is sufficiently special that I don’t mind this if reusing it is difficult but I’d like to have a comment that explains that this is only for cradle dependencies and ideally also an explanation why it is hard to reuse the one we already have.

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+getDependencyInfo fs = Map.fromList <$> mapM do_one fs
+
+  where
+    do_one fp = do
+      exists <- IO.doesFileExist fp
```
cocreature, line 422: Any reason not to call `getModificationTime` and catch the exception? While it probably doesn’t matter much here, I always feel a bit uneasy about having separate file exists check which are inherently racy. Performance wise, I would expect that the separate check is probably also slower but it shouldn’t be on the hot path so I doubt it matters.

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+        cfp <- liftIO $ canonicalizePath file
+        case HM.lookup (toNormalizedFilePath' cfp) v of
+          Just (_, old_di) -> do
+            deps_ok <- checkDependencyInfo old_di
+            unless deps_ok $ do
+              modifyVar_ fileToFlags (const (return Map.empty))
+              -- Keep the same name cache
+              modifyVar_ hscEnvs (return . Map.adjust (\(h, _) -> (h, [])) hieYaml )
+          Nothing -> return ()
+        -- We sort so exact matches come first.
```
cocreature, line 367: This comment looks like it might be leftover from my first attempt at multicomponent support. At least I don’t understand how it applies here.

exe/Main.hs
```diff
@@ -188,22 +190,235 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+        case HM.lookup (toNormalizedFilePath' cfp) v of
```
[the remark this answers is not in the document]
cocreature: Yeah this doesn’t look right to me.

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+            Just opts -> do
+                --putStrLn $ "Cached component of " <> show file
+                pure (fst opts)
+            Nothing-> do
+                finished_barrier <- newBarrier
```
cocreature, line 373: I don’t understand this code. It sounds like you are trying to keep this running in the background so the next time you get here it will finish quickly if you get interrupted the first time. However, you will create a new barrier the next time you get here so it seems like the work of the thread is just wasted. Am I missing something?

exe/Main.hs
```diff
@@ -188,22 +192,256 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+    let getOptions file = do
+            hieYaml <- cradleLoc file
+            sessionOpts (hieYaml, file)
+    -- The lock is on the `runningCradle` resource
+    return $ \file -> liftIO $ withLock lock $ do
+        as <- readIORef runningCradle
+        finished <- poll as
+        case finished of
+            Just {} -> do
+                as <- async $ getOptions file
```
cocreature, line 398: It looks like you can get interrupted between creating the async and updating the IORef. Should this be masked?

src/Development/IDE/Core/Compile.hs
```diff
@@ -447,8 +449,51 @@ getModSummaryFromBuffer fp contents dflags parsed = do
     where
       sourceType = if "-boot" `isSuffixOf` takeExtension fp then HsBootFile else HsSrcFile
 
-
--- | Given a buffer, flags, file path and module summary, produce a
+-- | Given a buffer, env and filepath, produce a module summary by parsing only the imports.
+--   Runs preprocessors as needed.
+getModSummaryFromImports
```
cocreature, line 454: This looks like it has probably become obsolete by the changes from @pepeiborra that we landed in the meantime.

test/exe/Main.hs
```diff
@@ -2323,9 +2345,9 @@ nthLine i r
     | i >= Rope.rows r = error $ "Row number out of bounds: " <> show i <> "/" <> show (Rope.rows r)
     | otherwise = Rope.takeWhile (/= '\n') $ fst $ Rope.splitAtLine 1 $ snd $ Rope.splitAtLine (i - 1) r
 
-getWatchedFilesSubscriptionsUntilProgressEnd :: Session [Maybe Value]
-getWatchedFilesSubscriptionsUntilProgressEnd = do
-      msgs <- manyTill (Just <$> message @RegisterCapabilityRequest <|> Nothing <$ anyMessage) (message @WorkDoneProgressEndNotification)
+getWatchedFilesSubscriptionsUntil :: forall end . (FromJSON end, Typeable end) => Session [Maybe Value]
```
cocreature, line 2348: How hard is it to setup a test that verifies that multi cradles at least work to some degree and we don’t break them completely? I guess direct-style cradles are a bit annoying to setup correctly but maybe we can make some simple example with two packages A, B and a cabal file and verify that we can first load a file from B check that go to definition to A does not work. Then load a file from A and check that it works?

mpickering:
Thanks for your comments Moritz. There is a rebased version on my fork which also has a few other fixes which are not obvious so I will get that pushed sometime next week. There is one known issue which I am not sure what to do about.

If you have two packages in your project A and C and if A depends on B which depends on C then if you load A and C into a session but not B then you get confusing errors about the same type coming from different packages.

My idea about how to fix this is to try using OneShot mode instead of --make mode, which seems to make sense to me anyway. I am predicting this will fix it as this will bypass the HPT which is why the fake_uid thing is necessary in the first place, to my understanding. 

Would you still merge the patch with this issue? It is still an improvement anyway over no multi-component support at all but I am not sure how to detect this situation and fail gracefully.

cocreature:
> Would you still merge the patch with this issue? It is still an improvement anyway over no multi-component support at all but I am not sure how to detect this situation and fail gracefully.

Yes, that doesn’t seem too bad. Just add it to the troubleshooting section.

mpickering:
ok then the plan is I will update the PR with the other changes from the fork and address your review comments. Then ping you again for a second review and after it is merged I can try the OneShot mode.

exe/Main.hs
mpickering: Ok, but there are other places where imports are already aligned..

ndmitchell: Yep, the formatting on this code reveals the history of multiple different authors. One day it would be great to run a real formatter over it, since that's the only way to recover from the mess we have now.

exe/Main.hs
mpickering: There is one session which contains everything which doesn't have a hie.yaml file. This is probably wrong..

exe/Main.hs
mpickering: Yes but things are very broken anyway if you do not list all the targets. There is potential to improve this in the future by somehow calling downsweep to establish what should implicitly be in the component but it has not been clear to me how to do this yet as the rules currently depend on `GetSession`.

exe/Main.hs
mpickering: Good point.

exe/Main.hs
mpickering: It seems you might be right that this `finished_barrier` is unnecessary now. I will try removing it.

exe/Main.hs
mpickering: Yes good point.

mpickering:
@cocreature I have pushed a new version which contains some bugs fixes.

* Supporting PackageImports properly
* Creating different interface file directories for different components
* Many more comments and code clean ups

I think it would benefit from another complete read.

src/Development/IDE/GHC/Util.hs
```diff
@@ -166,16 +167,20 @@ moduleImportPath (takeDirectory . fromNormalizedFilePath -> pathDir) mn
 --   if they are created with the same call to 'newHscEnvEq'.
 data HscEnvEq
     = HscEnvEq !Unique !HscEnv
+               [(InstalledUnitId, DynFlags)] -- In memory components for this HscEnv
+               -- This is only used at the moment for the import dirs in
+               -- the DynFlags
     | GhcVersionMismatch { compileTime :: !Version
```
mpickering, line 173: We can make this GhcVersionMismatch into a proper diagnostic now I think rather than this special constructor on `HscEnvEq`. Not sure whether to do that in this patch.

src/Development/IDE/Core/Rules.hs
```diff
@@ -472,7 +480,7 @@ instance Hashable GhcSessionIO
 instance NFData   GhcSessionIO
 instance Binary   GhcSessionIO
 
-newtype GhcSessionFun = GhcSessionFun (FilePath -> Action HscEnvEq)
+newtype GhcSessionFun = GhcSessionFun (FilePath -> Action (IdeResult HscEnvEq))
```
mpickering, line 483: Notice I made `GhcSessionFun` return a proper `IdeResult` so that you can get proper diagnostics for failing cradle loads now.

src/Development/IDE/GHC/Util.hs
pepeiborra: I have another PR making changes to `HscEnvEq` which will need to be reviewed once this is merged, and I am happy to convert `GhcVersionMismatch` to a proper diagnostic if that is possible at all

mpickering: Would you prefer me to remove it in this patch? Or leave it until later?

hie.yaml
```diff
@@ -1 +1,10 @@
...
+cradle:
+  cabal:
+    - path: "./src"
+      component: "ghcide:lib:ghcide"
+    - path: "./exe"
+      component: "ghcide:exe:ghcide"
+    - path: "./test"
+      component: "ghcide:test:ghcide-tests"
+    - path: "./test/preprocessor"
+      component: "ghcide:exe:ghcide-test-preprocessor"
```
fendor, line 10: Needs a stack pendant

exe/Main.hs
```diff
@@ -190,22 +184,404 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+    -- The main function which gets options for a file. We only want one of these running
+    -- at a time. Therefore the IORef contains the currently running cradle, if we try
+    -- to get some more options then we wait for the currently running action to finish
+    -- before attempting to do so.
+    let getOptions :: FilePath -> IO (IdeResult HscEnvEq)
+        getOptions file = do
+            hieYaml <- cradleLoc file
+            sessionOpts (hieYaml, file)
+    -- The lock is on the `runningCradle` resource
+    lock <- newLock
```
cocreature, line 388: Why do we use a lock + an IORef instead of an `MVar/Var`?

exe/Main.hs
```diff
@@ -190,22 +184,404 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+    return $ \file -> do
+      liftIO $ withLock lock $ do
+        as <- readIORef runningCradle
+        -- Is a cradle running?
+        finished <- poll as
```
cocreature, line 393: why do we first poll instead of just calling `wait` directly?

exe/Main.hs
```diff
@@ -190,22 +184,404 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+
+-- See Note [Multi Cradle Dependency Info]
+type DependencyInfo = Map.Map FilePath (Maybe UTCTime)
+type HieMap = Map.Map (Maybe FilePath) (HscEnv, [RawComponentInfo])
+type FlagsMap = Map.Map (Maybe FilePath) (HM.HashMap NormalizedFilePath (IdeResult HscEnvEq, DependencyInfo))
+
+-- This is pristine information about a component
+data RawComponentInfo = RawComponentInfo
+  { rawComponentUnitId :: InstalledUnitId
+  -- | Unprocessed DynFlags. Contains inplace packages such as libraries.
```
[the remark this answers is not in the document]
fendor: The first record does not have a comment. So, this is "intentional", although I see how it could be confusing.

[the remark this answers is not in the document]
cocreature: I know but afaik you still need to move it after the comma for haddock to associate it with the correct field.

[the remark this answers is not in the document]
fendor: Haddock

looks right to me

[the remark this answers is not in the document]
cocreature: interesting, seems weird to me that haddock allows putting it before the comment but I guess if it works it’s fine. Thanks for checking!

src/Development/IDE/Import/FindImports.hs
```diff
@@ -55,55 +55,72 @@ instance NFData Import where
...
+  findM doesExist (concatMap candidates import_dirss)
   where
     maybeBoot ext
       | isSource = ext ++ "-boot"
       | otherwise = ext
 
+mkImportDirs :: (M.InstalledUnitId, DynFlags) -> (PackageName, [FilePath])
+mkImportDirs (i, df@DynFlags{importPaths}) = (pn , importPaths)
+  where
+    pn = case getPackageName df i of
```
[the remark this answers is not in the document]
mpickering: No it's not safe and I have a patch which I need to apply to this branch which fixes it.

exe/Main.hs
mpickering: I am not sure what you are suggesting here.

exe/Main.hs
mpickering: Seems like that should work.

src/Development/IDE/Core/Rules.hs
```diff
@@ -482,11 +490,18 @@ loadGhcSession = do
...
+    -- However, it should also cut-off early if it's the same HscEnv as
+    -- last time
     defineEarlyCutoff $ \GhcSession file -> do
         GhcSessionFun fun <- useNoFile_ GhcSessionIO
+        alwaysRerun
         val <- fun $ fromNormalizedFilePath file
-        opts <- getIdeOptions
-        return ("" <$ optShakeFiles opts, ([], Just val))
+
+        -- TODO: What was this doing before?
```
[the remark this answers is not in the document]
mpickering: It still seems pretty indirect to check for `optShakeFiles` being set and then ignore the result. Perhaps what we really need is an `optDaml` flag? I have put back this case for now.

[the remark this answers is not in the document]
cocreature: That’s a fair point. I think for this PR putting the hack back is simpler but very happy to revisit this later.

exe/Main.hs
```diff
@@ -190,22 +184,404 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+      liftIO $ withLock lock $ do
+        as <- readIORef runningCradle
+        -- Is a cradle running?
+        finished <- poll as
+        -- If the cradle is not finished, then wait for it to finish.
+        when (isNothing finished) (void $ wait as)
+        join $ mask_ $ do
+          as <- async $ getOptions file
+          writeIORef runningCradle as
+          return (wait as)
```
cocreature, line 399:
```suggestion
    runningCradle <- newVar dummyAs
    return $ \file -> do
      liftIO $ modifyVar $ \as ->
        -- Is a cradle running?
        finished <- poll as
        -- If the cradle is not finished, then wait for it to finish.
        when (isNothing finished) (void $ wait as)
        join $ mask_ $ do
          as <- async $ getOptions file
          return (as, wait as)
```
This is what I was trying to suggest. Definitely did not check if it typechecks :slightly_smiling_face:

mpickering: I pushed a fix based on this. Can you check it to see whether you think I need a `mask_`?

exe/Main.hs
```diff
@@ -190,22 +190,397 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+    -- before attempting to do so.
+    let getOptions :: FilePath -> IO (IdeResult HscEnvEq)
+        getOptions file = do
+            hieYaml <- cradleLoc file
+            sessionOpts (hieYaml, file)
+    return $ \file -> do
+      liftIO $ join $ modifyVar runningCradle $ \as -> do
+        -- If the cradle is not finished, then wait for it to finish.
+        void $ wait as
+        as <- async $ getOptions file
```
cocreature, line 397: You can still get killed here afaict which will mean that the async will be running but you haven’t updated the var. I think you might want a `mask_` (but not uninterruptible) around the whole `modifyVar`? Then you can get interrupted during the `wait` but not between the `async` and updating the mvar.

mpickering: Yes, that seems to be what happens in `memoIO` as well. I am starting to look into fixing the tests and adding a test now if we are happy with the rest of the patch.

pepeiborra, line 397: IIUC the reason for running the cradle in a background thread is to address #494.

I believe that because of the `alwaysRerun` in the `GhcSession` rule, this async dance is performed once, for every file, for every `shakeRun` that involves an Action that depends on the `GhcSession` rule. So if we want to typecheck a project with 200 files, that's 200 one-shot threads which all they do is probably read off a hashmap. And because of kick, this will happen for every single keystroke...

This worries me, it could be a material performance regressoin for non multi-cradle projects. Have any measurements been done? 
I would prefer if this was doing the simplest thing possible for now, i.e. invoke hie-bios in the current thread, and the fix for #494 left for a future PR.

mpickering: I have used my branch working on GHC regularly, but that doesn't have `kick`. The patch is broken without fixing #494 as you point out. 

We are already spawning thousands of threads of every keystroke because of all the `GetFileExists` rules which go via shake, and that is even worse by default because shake spawns each one onto a new capability (see https://github.com/ndmitchell/shake/pull/751)

So I would prefer not to change this as it already took a bit of effort to get right and doesn't seem to be an issue in practice.

exe/Main.hs
```diff
@@ -190,22 +187,372 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+  f' <- canonicalizePath f
+  return [(toNormalizedFilePath' f')]
+
+setNameCache :: IORef NameCache -> HscEnv -> HscEnv
+setNameCache nc hsc = hsc { hsc_NC = nc }
+
+-- This is the key function which implements multi-component support. All
+-- components mapping to the same hie,yaml file are mapped to the same
+-- HscEnv which is updated as new components are discovered.
+loadSession :: FilePath -> Action (FilePath -> Action (IdeResult HscEnvEq))
```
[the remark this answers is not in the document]
mpickering: Done.

exe/Main.hs
```diff
@@ -190,22 +187,372 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Originally, we set the cache directory for the various components once
+on the first occurrence of the component.
+This works fine if these components have no references to each other,
+but you have components that depend on each other, the interface files are
+updated for each component.
+After restarting the session and only opening the component that depended
+on the other, suddenly the interface files of this component are stale.
+However, from the point of view of `ghcide`, they do not look stale,
+thus, not regenerated and the IDE shows weird errors such as:
```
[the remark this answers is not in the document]
mpickering: I thought about it but don't think so. This seemed the most obviously correct way to implement it.

exe/Main.hs
pepeiborra: Your branch has fundamentally different performance characteristics, so that previous experience doesn't apply. 

Does Shake spawn a new thread for every dependency? I thought it used a threadpool. This change will really spawn one thread for every module, for every keystroke. That includes even simply moving the cursor in some clients like Emacs, which send a hover request per keystroke.

We don't need to change this if the measurements show a non-material performance impact.

exe/Main.hs
```diff
@@ -190,22 +187,372 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+Since this causes a lot of recompilation, we only update the cache-directory,
+if the dependencies of a component have really changed.
+E.g. when you load two executables, they can not depend on each other. They
+should be filtered out, such that we dont have to re-compile everything.
+-}
+
+
+setCacheDir :: MonadIO m => String -> [String] -> ComponentOptions -> DynFlags -> m DynFlags
+setCacheDir prefix hscComponents comps dflags = do
+    cacheDir <- liftIO $ getCacheDir prefix (hscComponents ++ componentOptions comps)
```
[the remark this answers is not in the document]
mpickering: Thanks.

exe/Main.hs
```diff
@@ -190,22 +187,372 @@ showEvent lock (EventFileDiagnostics (toNormalizedFilePath' -> file) diags) =
...
+
+checkDependencyInfo :: DependencyInfo -> IO Bool
+checkDependencyInfo old_di = do
+  di <- getDependencyInfo (Map.keys old_di)
+  return (di == old_di)
+
+-- Why do we implement our own file modification tracking here?
+-- The primary reason is that the custom caching logic is quite complicated and going into shake
+-- adds even more complexity and more indirection. I did try for about 5 hours to work out how to
+-- use shake rules rather than IO but eventually gave up.
```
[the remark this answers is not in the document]
mpickering: Yes exactly. Glad it wasn't just my poor brain who came to this conclusion.

mpickering:
Fixed HLint and responded to Pepe's comments.

exe/Main.hs
mpickering: Yes, shake spawns a new thread for each dependency but bounds the number of threads to the number of capabilities. If you look at an eventlog you will see what happens is that each dependency gets scheduled onto a different capability which is really bad if you have lots of short actions. 

I tried to do some tests just now but GHC master seems broken with ghcide so will have to do it later.

mpickering:
There is something broken either with this patch or with master which means loading GHC is broken. It is working on my branch. Perhaps it is fixed by #558 

Will need to come back to this later now.

pepeiborra:
> There is something broken either with this patch or with master which means loading GHC is broken. It is working on my branch. Perhaps it is fixed by #558
> 
> Will need to come back to this later now.

Worth opening a separate issue for the GHC loading problems?

exe/Main.hs
ndmitchell: Shake could easily avoid spawning fresh threads for each action. I've left a load of comments on https://github.com/ndmitchell/shake/pull/751 - I think the fixes proposed are great, unless the predictions are wrong, and then the whole system becomes single threaded. Happy to get clarification on that ticket.

pepeiborra: And this code is committing the same sin as Shake, using `async` instead of `asyncOn`. Maybe applying the same trick here would be enough to minimize the cost until we can take some measurements?

src/Development/IDE/GHC/Compat.hs
```diff
@@ -302,3 +304,10 @@ getConArgs = GHC.getConArgs
 #else
 getConArgs = GHC.getConDetails
 #endif
+
+getPackageName :: DynFlags -> Module.InstalledUnitId -> Maybe PackageName
+#if MIN_GHC_API_VERSION(8,10,0)
+getPackageName dfs i = getPackageName <$> lookupPackage dfs (DefUnitId i)
+#else
+getPackageName dfs i = packageName <$> lookupPackage dfs (Module.DefiniteUnitId (Module.DefUnitId i))
+#endif
```
pepeiborra, line 313: This builds in 8.10 for me:
```suggestion
getPackageName dfs i = packageName <$> lookupPackage dfs (Module.DefiniteUnitId (Module.DefUnitId i))
```

exe/Main.hs
pepeiborra: After testing in the Strats codebase I don't see a huge performance difference, so I am dropping my reservations. What I tested:
- hover (warm)
- diagnostics after a single edit
- hover after edit

mpickering: Thanks Pepe, it is a good observation and I would like set up some more robust telemetry to be able to analyse these issues easily in future.

The difference with shake is that we use an unbounded number of threads so GHC's scheduler can decide how to assign each thread to a capability.

src/Development/IDE/GHC/Compat.hs
mpickering: :+1:

mpickering:
@cocreature I tried to add a test but you are right it's quite difficult to get something robust working as it would also require settting up package databases and so on. I think the best idea for testing is to rely on the hls tests which afaik  invoke tools like cabal to set up the cradles. This is not very satisfactory but not sure what else to do.

cocreature:
What is stopping us from invoking cabal/stack in our tests?

mpickering:
@cocreature None of the other tests invoke cabal.

CI seems to be failing for 8.4 which I thought was unsupported at this point.

cocreature:
> None of the other tests invoke cabal.

That’s mainly because none of the other tests need cabal. I think having a test in this repo would be quite useful. Happy to help with that.

As for 8.4, so far that was still supported. It doesn’t benefit from some of the performance improvements in particular around `.hie` files but it does work. I’m open to dropping it. @ndmitchell originally added the support because he needed it. Not sure if that is still the case. Is there something specific that makes this patch hard to get working on 8.4? The failure on CI looks fairly harmless but I don’t know how much breaks after fixing that.

ndmitchell:
It would still be useful to me, as some code bases I work on are still stuck at 8.4.

mpickering:
@cocreature Can you please finish this off by fixing CI so that cabal is on the PATH? I made an attempt but it doesn't appear to be enough.

cocreature:
Sure, I’ll take a look. Thanks!

pepeiborra:
@mpickering any chance that you can rebase on top of https://github.com/digital-asset/ghcide/pull/590 and run the benchmark suite, showing the BEFORE and AFTER timings?

mpickering:
@pepeiborra I really do not want to do that given how long this review cycle has already gone on for. I am now turning my attention to upstreaming other patches from my branch so we can get Zubin's hiedb patches merged which will alter any timing measurements anyway.

pepeiborra:
> @pepeiborra I really do not want to do that given how long this review cycle has already gone on for. I am now turning my attention to upstreaming other patches from my branch so we can get Zubin's hiedb patches merged which will alter any timing measurements anyway.

I am very keen in seeing this PR get merged too, as I do have 2 PRs which have been open for weeks waiting for this one to get merged first. But given how cheap it is to run the benchmark suite, and the benefit of understanding what the performance impact is, if any, I don't see any reason not to do it. This is what I get in my home desktop PC (2011 quad-core Mac) : 
```
Benchmark BEFORE:

TOTAL hover = 2.57s (10 repetitions)
TOTAL getDefinition = 2.67s (10 repetitions)
TOTAL documentSymbols = 1.52s (100 repetitions)
TOTAL documentSymbols after edit = 2.92s (100 repetitions)
TOTAL completions after edit = 4.51s (10 repetitions)
TOTAL code actions = 2.47s (10 repetitions)
TOTAL code actions after edit = 3.59s (10 repetitions)

Benchmark AFTER:

TOTAL hover = 3.35s (10 repetitions)
TOTAL getDefinition = 3.37s (10 repetitions)
TOTAL documentSymbols = 1.39s (100 repetitions)
TOTAL documentSymbols after edit = 3.02s (100 repetitions)
TOTAL completions after edit = 5.32s (10 repetitions)
TOTAL code actions = 3.23s (10 repetitions)
TOTAL code actions after edit = 4.33s (10 repetitions)
```

mpickering:
It is not cheap to run the benchmark suite if I am now also expected to fix these regressions.. which is why I didn't want to run it.

ndmitchell:
@mpickering do you have a hypothesis as to why the performance goes down?

My personal view is that with big restructurings such as these, it's OK to merge with performance/cleanliness issues, and clean up after (case in point: HLint 3.0 codebase is/was a mess, with performance regressions, and we disabled space leaks - but we got rid of haskell-src-exts - and now we can start the cleanup).

cocreature:
@mpickering do the tests pass for you locally? They fail on CI as well as locally for me both when running the tests via Stack and via Cabal.
The error is a rather uninformative
```
       test/exe/Main.hs:1319:
        number of definitions
        expected: 1
         but got: 0
```
i’ve enabled debugging logs locally but couldn’t spot anything that looks wrong.

mpickering:
Yes works for me. Perhaps there is some timing issue?

```
    multi
      simple-multi-test:                                         OK (6.44s)
```

mpickering:
I had that error locally before when I got the source position of the thing wrong but given I fixed that. Not sure why it happens.

cocreature:
I don’t think this should be timing sensitive but I tried inserting a 10s timeout before opening `B` and before calling `getDefinitions` but that didn’t change anything. I also tried 8.8.3 (I used 8.6.5) before with no change.

With some debugging output I got the following:
```
# It is setting up the multi-hscenv
[INFO] Making new HscEnv[b-1.0.0-inplace,a-1.0.0-inplace]
# This is the output of GetSpanInfo so it has the right span
SpansInfo {spansExprs = [(SpanInfo 2 0 2 3 "()" (Named qux)) docs(SpanDocText []),(SpanInfo 2 6 2 9 "()" (Named foo)) docs(SpanDocText [])], spansConstraints = []}
# This is the output after filtering via `spansAtPoint` so the location is right
[(SpanInfo 2 6 2 9 "()" (Named foo)) docs(SpanDocText [])]
# It then ends up being filtered out by `getSpan` since the span is an `UnhelpfulSpan`
```
Afaik you get an unhelpful span if it’s from an external package so it looks like somehow the multi-component HscEnv isn’t being used but I don’t understand why.

mpickering:
@ndmitchell I imagine it is slower because of the issues with the shake scheduler with many short actions. Now `GhcSession` is always rerun I imagine that code path gets hit a bit harder?

ndmitchell:
@mpickering The benchmarks seem to show a 25% slowdown, which for an important feature touching all areas of the code, doesn't seem unreasonable. Better focus on correctness (which this code gives) and optimise later.

From what I can tell, you make `GhcSession` `alwaysRerun` so that means you need to run 1 extra rule per file. Given we already run in the region of 8 per file, I don't see that being a meaningful difference. In particular #589 removes 1 rule per file, saving about 7% total execution time, so I suspect of that 25% slowdown 7% is explained by Shake, and 18% isn't.

mpickering:
Anyone have any ideas about how I can reproduce the test failure?

mpickering:
@fendor Made some progress diagnosing that the test passes if you have a dirty dist-newstyle directory already in the `multi` folder. Still investigating exactly what the issue is.

cocreature:
Thanks for looking into it @fendor!

mpickering:
I think what is happening is this:

1. A is loaded, which depends on B so all identifiers from A and B are loaded into the NameCache
2. Because B is not initially in the HPT, the Name which is put into the NameCache has an empty `nameSrcSpan` field.
3. Therefore when we attempt to jump to definition into B, `nameSrcSpan` is empty so we fall back to the HIE file case.
4. However, there is also not an HIE file for module B as it was only loaded by the TypeCheck rule, which doesn't generate .hi files or .hie files.

I think the way forward here is:

1. For now, always generate .hie and .hi files for modules.
2. When hiedb lands, this code path will be rewritten to query the database so the problem will also be avoided. 

Thanks to @fendor and @wz1000 for helping me debug it.

mpickering:
Hmm always generating hie files doesn't help the situation.

cocreature:
I’m not quite following your explanation:
1. It’s not A that depends on B. B imports A and we load A first.
2. If we do it the other way around and load B first, then I thought loading A would recompile B. Does this not refresh the namecache?
3. By default cabal-install does not generate .hie files afaik so I don’t understand why having a dirty dist-newstyle directory helps. Are you manually enabling .hie files?

pepeiborra:
ghcide does not end up customizing the hieDir @cocreature, as it turns out that the flag always has a value, so the `.hie` files do end up inside the `dist-newstyle` folder.
https://github.com/digital-asset/ghcide/blob/0c9a0961abbeef851b4117e6408f15a6d46eb1f1/src/Development/IDE/GHC/Compat.hs#L192-L195

I can reproduce the problem using VSCode. And indeed, a dirty dist-newstyle folder makes the problem go away. Wiping the dist-newstyle folder and reloading brings the problem back. As an experiment I tried to wipe everything except for the `.hie` files, and the problem came back after reloading, so I don't think this is related to `.hie` files.

mpickering:
I have worked out why this is happening now and it's nothing to do with hie files as pointed out by Moritz and Pepe. The bug is in some new code I added to make package imports work.

mpickering:
Pushed some changes which I hope should fix the tests now, and a bonus test as well which loads the components in the other order. 

Who would have thought adding a test would actually show up a bug.

mpickering:
It now appears that `stack` sets an env var which cabal doesn't like. :woman_shrugging:

cocreature:
Pushed a fix (hopefully) for that.

mpickering:
Thanks Moritz, you beat me to a fix.. my push failed! Did you come up with that list of variablesthrough testing locally or from using your mind?

cocreature:
Tested locally, I already had this change from when I was looking at the failing test on the weekend. I just forgot to push it since it failed with cabal as well so I forgot about it.

mpickering:
Thanks, looks like the simple-multi tests pass on CI now.

cocreature:
Crap that breaks the plugin tests, let me try to unset the env more locally.

PR #535, opened 2020-04-24
Improve ghc version check

### LINKED ISSUE #534: GHC check broken when using haskell.nix
With ghcide + haskell.nix + ghc 8.8, I'm getting

>   ghcide compiled against GHC 8.8.3 but currently using an unknown version of GHC . This is
>   unsupported, ghcide must be compiled with the same GHC version as the project.

... because ghc is broken in the package database. With `reinstallableLibGhc`, ghc just isn't in the package database. Either scenario breaks the compatibility check.

I guess in the long term, hopefully the package database stuff can be fixed, but we probably also need an opt-out flag from this check (or that it turns no ghc version information into a warning).

_Originally posted by @fiadliel in https://github.com/digital-asset/ghcide/issues/532#issuecomment-618276756_

pepeiborra:
- retrieve runtime version from ghc executable, not from pkg db (ghc-check 0.3.0.0)
- Do not fail if unable to retrieve runtime version

Fixes #534 and #536

### DIFF
--- a/exe/Rules.hs
+++ b/exe/Rules.hs
@@ -15,7 +15,6 @@ import           Data.ByteString.Base16         (encode)
 import qualified Data.ByteString.Char8          as B
 import           Data.Functor                   ((<&>))
 import           Data.Text                      (Text, pack)
-import           Data.Version                   (Version)
 import           Development.IDE.Core.Rules     (defineNoFile)
 import           Development.IDE.Core.Service   (getIdeOptions)
 import           Development.IDE.Core.Shake     (actionLogger, sendEvent, define, useNoFile_)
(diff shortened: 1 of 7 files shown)


mpickering:
I am not sure this approach is very good. It's confusing at least. If you are going to require `ghc` is on the path, why not use `ghc --print-libdir` (like hie uses) rather than guessing with this hacky combination of env var and GHC.Paths. You definitely need to use the right libdir for your project as companies using forked GHCs may have compilers which report the same version but have incompatible libdirs with the stock GHC which ghcide may have been built with.

pepeiborra:
> I am not sure this approach is very good. It's confusing at least. If you are going to require `ghc` is on the path, why not use `ghc --print-libdir` (like hie uses) rather than guessing with this hacky combination of env var and GHC.Paths. You definitely need to use the right libdir for your project as companies using forked GHCs may have compilers which report the same version but have incompatible libdirs with the stock GHC which ghcide may have been built with.

Which approach do you have in mind? `makeGhcVersionCheck` takes the libdir as an argument, and uses that to drive the `Ghc` monad as well as to guess the path to the `ghc` binary. The libdir comes from ghcide, this hasn't changed.

Retrieving the runtime version from the package database only works if the ghc api can read the package database, which is not always the case across major versions.

domenkozar:
Is there a path forward? This is making ghcide-nix useless on GHC 8.8 :)

mpickering:
I think I misunderstood how the `ghc` executable was being called. It is actually calculated relative to the supplied libdir rather than requiring GHC be on the path. So, seems like an improvement overall.

cocreature:
Can you please try rebasing this branch. Somehow Azure seems to be stuck for reasons I don’t understand. Sorry for the trouble.

pepeiborra:
No trouble at all. Done!

PR #542, opened 2020-05-03
Report progress as a percentage

alanz:


ndmitchell:
I quite like seeing it in terms of modules completed vs modules total, since that has a very natural mapping to what I expect, and the fact it isn't linear is a lot less surprising. What motivates the change to percentage?

alanz:
The main motivation is that emacs lsp-mode does not show the "n of m" updates, so all you get is a spinning wheel without that.

And my haskell skills are stronger than my elisp ones

ndmitchell:
Ah, so we're sending the percentage and the message. I hope VS code displays both, and that would make quite an appealing UI.

alanz:
This is what I get, running hls incorporating that

Screenshot from 2020-05-03 15-22-10

ndmitchell:
Cool - including the percentage seems like a nice win then.

ndmitchell:
@alanz does this work in Emacs even if the percentage falls? My suggestion would be if it does, and doesn't harm Vim, then probably best to go forward with it - but if some client shows up that has issues when we (gently) violate the spec then we revert.

alanz:
I recall revisiting this some time back, and realising that the count made basically no sense, it kept flipping between 0,1,2 regardless of the amount of work to be done.

So just a generic "busy" makes more sense.  So I am happy to abandon this for now.

pepeiborra:
closing due to inactivity

PR #544, opened 2020-05-03
Fix HLint

cocreature:
Looks like the new version of hlint has a couple of new hints.

changelog_begin
changelog_end

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -14,9 +14,8 @@ import Control.Exception (catch)
 import Control.Monad
 import Control.Monad.IO.Class (liftIO)
 import Data.Aeson (FromJSON, Value)
-import Data.Char (toLower)
 import Data.Foldable
-import Data.List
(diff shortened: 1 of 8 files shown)


test/exe/Main.hs
```diff
@@ -1984,6 +1984,8 @@ cradleTests = testGroup "cradle"
     ,testGroup "loading" [loadCradleOnlyonce]
     ]
 
+{-# ANN loadCradleOnlyonce ("HLint: ignore Redundant bracket" :: String) #-}
```
ndmitchell, line 1987: Note the recommended ignore is `{- HLINT ignore Redundant bracket -}` as that won't cause a Template Haskell overhead parsing the Ann. I will fix the bug though!

cocreature: Thanks fixed! I can never remember the syntax. And thanks for opening an issue. I was just about to report it but clearly you beat me to it :slightly_smiling_face:

Issue #158, opened 2019-10-07
Strip path/span output from the actual error message


merijn:
`ghcide` already reports file and line/column offsets in the LSP metadata, so having them also be in the message is clutter. At the moment I'm using a patched version that just strips everything up to `: warning:` and `: error:`, but that's a bit of a hacky solution.

An easy (and reliable) first step to improve this situation might be to inject the `-fhide-source-paths` flag into the GHC configuration, which will at least stop reporting the full path and only report the file name.

I'll play around with flags and stripping some more to see if I can figure out a less hacky solution.

merijn:
Actually manually parsing the file prefix when `-fhide-source-paths` is enabled should be fairly easy, since it only reports the module's filename, because we can safely assume that name doesn't include a `:` (since GHC doesn't allow those in module names), which we can't assume for arbitrary full paths.

andrevdm:
@merijn Would you mind sharing which file you patched to get that working? I've not looked at the code yet but I'd love to add that in the interim as a quick fix.

merijn:
```diff
diff -r 61331d1d506d src/Development/IDE/GHC/Error.hs
--- a/src/Development/IDE/GHC/Error.hs  Mon Sep 30 07:54:57 2019 +0100
+++ b/src/Development/IDE/GHC/Error.hs  Thu Nov 07 16:52:53 2019 +0100
@@ -39,10 +39,14 @@
     { _range    = srcSpanToRange loc
     , _severity = Just sev
     , _source   = Just diagSource -- not shown in the IDE, but useful for ghcide developers
-    , _message  = msg
+    , _message  = T.strip . keepAfter ": warning:" . keepAfter ": error:" $ msg
     , _code     = Nothing
     , _relatedInformation = Nothing
     }
+  where
+    keepAfter s t
+        | T.isInfixOf s t = snd $ T.breakOnEnd s t
+        | otherwise = t
 
 -- | Produce a GHC-style error from a source span and a message.
 diagFromErrMsg :: T.Text -> DynFlags -> ErrMsg -> [FileDiagnostic]
```

tek:
is this going to be integrated?

ndmitchell:
Is there no way to remove this at the GHC level? I'd rather not monkey around in text messages unless we have to.

merijn:
I already looked into that, there isn't (currently) a flag in GHC that does this.

ndmitchell:
[...]

PR #546, opened 2020-05-04
Strip path information from diagnostic messages (#158)

tek:
As discussed in #158. Don't know if the test is up to standard, please advise.


tek:
looks like the test "redundant operator" failed, doesn't look related to this PR

pepeiborra:
> looks like the test "redundant operator" failed, doesn't look related to this PR

This PR changes the parsing of error messages into diagnostics, which are then analysed by the code actions logic to suggest fixes. So it is very much related.

tek:
ok, then I'll have another look!

tek:
The offending location had an `#if` on the ghc version to define the expected message, and with the changes the message for 8.4 became the same as for 8.6. I would speculate that using `ErrUtils.formatErrDoc` just coincides the two paths and that simply removing the `#if` is a proper fix, but I know too little about the project. What's your assessment of this, @pepeiborra ?

ndmitchell:
@tek - if that fixes the test, it's very likely the right solution - sounds totally plausible.

tek:
wonderful!

cocreature (approved):
Awesome, thank you!

PR #552, opened 2020-05-08
Update to hie-bios 0.5.0

fendor:


cabal.project
```diff
@@ -0,0 +1,7 @@
+packages:
```
cocreature, line 1: This looks redundant now that you released it.

fendor: Thanks!

fendor:
We are normalising the import directories later in the pipeline (e.g. in `initSession` after `addCmdOpts`), maybe the test-suite needs to be adapted as well.

cocreature:
> maybe the test-suite needs to be adapted as well.

I don’t think it’s the test suite that needs changing here. My suspicion is that https://github.com/digital-asset/ghcide/blob/cd6f62bbedee39791c567ab7b1f01f33ace2edb0/src/Development/IDE/Core/IdeConfiguration.hs#L61is somehow return `False` now for files in `-isrc`. Probably adding a bit of debugging output there should help narrow this down.

fendor:
With some logs, I traced that `isWorkSpaceFile` is only called four times. 
Latest file registrations look like this:
```
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "/tmp/extra-dir-91410846469827/WatchedFilesMissingModule.hs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "/tmp/extra-dir-91410846469827/WatchedFilesMissingModule.lhs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "/tmp/extra-dir-91410846469827/src/WatchedFilesMissingModule.hs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "/tmp/extra-dir-91410846469827/src/WatchedFilesMissingModule.lhs")])])]))
```

while old ones look like this: 
```
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "/tmp/extra-dir-87581799130537/WatchedFilesMissingModule.hs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "/tmp/extra-dir-87581799130537/WatchedFilesMissingModule.lhs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "WatchedFilesMissingModule.hs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "WatchedFilesMissingModule.lhs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "src/WatchedFilesMissingModule.hs")])])]))
FileRegs: Just (Object (fromList [("watchers",Array [Object (fromList [("kind",Number 5.0),("globPattern",String "src/WatchedFilesMissingModule.lhs")])])]))
```

fendor:
I think I found it. The problem is that in the last release, import directories are not necessarily absolute.
In the direct cradle, we used to not make the import directory absolute, only for cabal and stack cradles.
With the latest-release all import directories are now absolute, since it is assumed that the import directories are relative to the cradle component root directory. I think that changed caused the test-failure.

jneira:
There are two failing tests, shared with mpickering/hls and the multi pr:
```
watched files
    workspace files:                                             FAIL (1.30s)
      test/exe/Main.hs:491:
      expected: 6
       but got: 4
    non workspace file:                                          FAIL (1.37s)
      test/exe/Main.hs:503:
      expected: 4
       but got: 2
```

mpickering:
Looks like the issue is that once made absolute some of the paths are the same as each other. So I think update the expected output of the test and it's fine.

cocreature:
Yeah, changing https://github.com/digital-asset/ghcide/blob/bc25ea79ab73067e4ea8d4ae4c1d7e3d9144163e/test/exe/Main.hs#L492 to a 4 seems like the right thing to do.

test/exe/Main.hs
```diff
@@ -481,26 +481,22 @@ watchedFilesTests = testGroup "watched files"
...
+      --  - /path-to-workspace/src/WatchedFilesMissingModule.lhs
+      liftIO $ length watchedFileRegs @?= 4
 
   , testSession' "non workspace file" $ \sessionDir -> do
       liftIO $ writeFile (sessionDir </> "hie.yaml") "cradle: {direct: {arguments: [\"-i/tmp\"]}}"
       _doc <- createDoc "A.hs" "haskell" "{-# LANGUAGE NoImplicitPrelude#-}\nmodule A where\nimport WatchedFilesMissingModule"
       watchedFileRegs <- getWatchedFilesSubscriptionsUntil @PublishDiagnosticsNotification
 
-      -- Expect 4 subscriptions (/tmp does not get any as it is out of the workspace):
+      -- Expect 3 subscriptions (/tmp does not get any as it is out of the workspace):
```
cocreature, line 496:
```suggestion
      -- Expect 2 subscriptions (/tmp does not get any as it is out of the workspace):
```

Issue #123, opened 2019-09-24
Add garbage collect of diagnostics


ndmitchell:
Currently if you have a file open with errors, then shut it, the errors persist. The reason is that we don't have any sophisticated removal of non-interesting files, unlike what we have with DAML.

pepeiborra:
I think that we want to at least keep the errors for dependencies of the current files of interest, those are very helpful. And I think that @mpickering may want a completely different behaviour here, as he typechecks all the targets at startup.

In implementation terms, we want to garbage collecting diagnostics which do not belong to the set of dependencies of the files of interest. This should happen after `kick` is done. There is some logic to do this already in the repo, but it is not being used and likely to be deleted as dead code and lost:
https://github.com/digital-asset/ghcide/blob/e16e841fa7e51b7950fd309196c1b16e51f95983/src/Development/IDE/Core/Shake.hs#L448-L462

mpickering:
`garbageCollect` should run when a file is deleted from the FOI HashSet, not on every kick in my opinion. (As in my opinion, nothing should run in kick as it adds costs to every call).

ndmitchell:
I agree very little should run in kick, but if you have a file of interest `A.hs` which imports `B.hs`, then edit `A.hs` so it no longer imports `B.hs`, then any warnings shown as a result of `B.hs` should disappear if it's not included by another other root. In DAML we adjusted that logic, so only if A.hs parsed successfully did we trim `B.hs`, to avoid losing warnings on every parse error.

jneira:
This continue happening with ghcide and hls master

PR #554, opened 2020-05-08
ShakeSession and shakeEnqueue

pepeiborra:
Currently we start a new Shake session for every interaction with the Shake
database, including type checking, hovers, code actions, completions, etc.
Since only one Shake session can ever exist, we abort the active session if any
in order to execute the new command in a responsive manner.

This is suboptimal in many, many ways:

- A hover in module M aborts the typechecking of module M, only to start over!
- Read-only commands (hover, code action, completion) need to typecheck all the
  modules! (or rather, ask Shake to check that the typechecks are current)
- There is no way to run non-interfering commands concurrently

This is an experiment inspired by @mpickering  `ShakeQueue` and
the follow-up discussion in https://github.com/mpickering/ghcide/issues/7

We introduce the concept of the `ShakeSession` as part of the IDE state.
The `ShakeSession` is initialized by a call to `shakeRun`, and survives until
the next call to `shakeRun`. It is important that the session is restarted as
soon as the filesystem changes, to ensure that the database is current.

The `ShakeSession` enables a new command `shakeEnqueue`, which appends work to
the existing `ShakeSession`. This command can be called in parallel without any
restriction.

`shakeEnqueue` then replaces `shakeRun` for all user actions that do not involve a change to the filesystem. It's very neat that this requires a single change in `runAction`.

The result is lightning fast performance for hover and code actions. Not so for completions, which usually arise while typing, and do not find an existing, warm `ShakeSession`. More thought is needed for these.

The hack used to inject more work in the `ShakeSession` quite possibly breaks internal Shake invariants, which is why this is a Request For Comments. I have tested it with GHC and ghcide itself, and haven't observed any issues. 

TODO

- [x] rework progress reporting mechanism


mpickering:
Am I correct that this approach doesn't guarantee that all actions are completed? How would you implement the initial loading action or the propagating to parents behaviour of the shake queue?

It also seems hard to make completions fast using this approach without implementing something quite similar to what is implemented in the shake queue approach.

pepeiborra:
> Am I correct that this approach doesn't guarantee that all actions are completed? How would you implement the initial loading action or the propagating to parents behaviour of the shake queue?
> 
> It also seems hard to make completions fast using this approach without implementing something quite similar to what is implemented in the shake queue approach.

Correct, this is a minimal change to address only the issue of suboptimal cancellations. The idea is to address the performance issues one by one, step by step. 

How does the queue make completions fast?

mpickering:
The queue does not help directly but because the queue uses stale information, the request to recompute the completions is placed into the queue after the request is dealt with to ensure that the completions information is refreshed. 

The main benefit of the queue in my opinion is that every action is guaranteed to be completed so you can remove the very suboptimal `kick`.

pepeiborra:
What is the actual impact of `kick`? 

It causes a typecheck of all the files of interest, so if I only have one file open, then `kick` is actually optimal. But even in that scenario completions take >1s to appear, and that would not get better with improvements to `kick` as it is already optimal in this situation.

mpickering:
Is your question rhetorical or would you like me to give you my opinion?

ndmitchell:
As to the Shake parts, that seems fine. Nothing breaks any internal invariants. In fact, I think you can go even further, and have `pumpAction` both continue pumping and fork off a new action in parallel. I'm not totally sure you can do that without consuming stack space though.

Something along these lines (either this approach or the one in @mpickering 's branch) seems like a good thing to do. I think we are quite quickly approaching the trade-off between correctness and performance, so maybe it's worth discussing that in words first? I believe the position of @cocreature is (was?) that we should never give incorrect information to an LSP query? It seems from this code and the branch that people (@pepeiborra and @mpickering) want to weaken that. How far? What invariants are required? What can we relax? How far can we relax it?

pepeiborra:
Thanks Neil for confirming that this trick won't break any Shake invariants. Hopefully it doesn't break any *ghcide* invariants either. 

While there is indeed a tension between accuracy and responsiveness, and I think accuracy can be relaxed in some cases (completions), I would not expect this PR to alter the current balance in ghcide. It increases the responsiveness of hovers, code actions and go-to definition, for free!

pepeiborra:
Promoting this from RFC to Merge Request

pepeiborra:
One issue with this approach is that ghcide doesn't send the `WorkDoneProgressEnd` notification timely, as the `runShake` thread stays up to support the `ShakeSession`. 

I plan to fix this by attaching the progress reporting directly to `kick`

src/Development/IDE/Core/Shake.hs
```diff
@@ -375,57 +379,114 @@ shakeProfile :: IdeState -> FilePath -> IO ()
...
+            forever $ join $ liftIO $ atomically $ readTQueue actionQueue
+
+        progressRun
+          | reportProgress = lspShakeProgress getLspId eventer inProgress
+          | otherwise = return ()
+
+        workRun restore = withAsync progressRun $ \progressThread -> do
+          let acts' =
+                [ [] <$ pumpAction
+                , parallel acts
```
pepeiborra, line 447: This call to `parallel` worries me. If it has negative performance implications, `lspShakeProgress` could be changed to auto-detect completion when the number of items `inProgress` is 0.

pepeiborra:
Hmm, progress reporting still not working as expected

src/Development/IDE/Core/Shake.hs
pepeiborra: I haven't noticed any noticeable performance degradation due to `parallel`

pepeiborra:
This is ready for final review. We have been using this branch at Strats for the past week, without issues, and:

* The perf improvements for hover, code actions and goto-def are very noticeable.
* It makes a huge difference in clients like emacs-lsp which send a hover request for every movement of the cursor. 
* `ShakeSession` removes the need for code actions and other plugins to batch their calls to `runAction`, e.g.  https://github.com/digital-asset/ghcide/blob/9129475b87f9ddf52b1599b086dc93c2f206d969/src/Development/IDE/Plugin/CodeAction.hs#L77

src/Development/IDE/Core/OfInterest.hs
```diff
@@ -79,4 +81,9 @@ modifyFilesOfInterest state f = do
     OfInterestVar var <- getIdeGlobalState state
     files <- modifyVar var $ pure . dupe . f
     logDebug (ideLogger state) $ "Set files of interest to: " <> T.pack (show $ HashSet.toList files)
-    void $ shakeRun state []
+    void $ shakeRun state [kick]
+
+kick :: Action ()
+kick = do
+    files <- getFilesOfInterest
+    void $ uses TypeCheck $ HashSet.toList files
```
ndmitchell, line 89: The reason this was outside the core was because it's wrong. In DAML we have a much better kick operation that takes care of garbage collection etc which we one day planned to move over. This was meant to be a temporary hack. But that doesn't mean it's in the wrong place now, but something to think about for the future.

src/Development/IDE/Core/Shake.hs
```diff
@@ -104,6 +105,10 @@ data ShakeExtras = ShakeExtras
     -- accumlation of all previous mappings.
     ,inProgress :: Var (HMap.HashMap NormalizedFilePath Int)
     -- ^ How many rules are running for each file
+    ,getLspId :: IO LspId
+    ,reportProgress :: Bool
```
ndmitchell, line 109: Can we get comments on these fields - I find these some of the most useful comments in the code base.

src/Development/IDE/Core/Shake.hs
```diff
@@ -220,13 +219,25 @@ type IdeRule k v =
...
+--   Keeps the 'ShakeDatabase' open, so at most one 'ShakeSession' per database.
+data ShakeSession = ShakeSession
+  { cancelShakeSession :: !(IO ())
+    -- ^ Close the Shake session
+  , runInShakeSession  :: !(forall a . Action a -> IO (IO a))
+    -- ^ Enqueue an action in the Shake session.
+  }
+
+nilShakeSession :: ShakeSession
+nilShakeSession = ShakeSession (pure ()) (\_ -> error "nilShakeSession")
```
ndmitchell, line 232: empyShakeSession?

src/Development/IDE/Core/Shake.hs
ndmitchell: It doesn't massively worry me.

src/Development/IDE/Core/Shake.hs
```diff
@@ -439,27 +503,6 @@ getHiddenDiagnostics IdeState{shakeExtras = ShakeExtras{hiddenDiagnostics}} = do
     val <- readVar hiddenDiagnostics
     return $ getAllDiagnostics val
 
--- | FIXME: This function is temporary! Only required because the files of interest doesn't work
-unsafeClearDiagnostics :: IdeState -> IO ()
-unsafeClearDiagnostics IdeState{shakeExtras = ShakeExtras{diagnostics}} =
-    writeVar diagnostics mempty
-
--- | Clear the results for all files that do not match the given predicate.
-garbageCollect :: (NormalizedFilePath -> Bool) -> Action ()
```
ndmitchell, line 448: This is useful and should stay - it relates to how kick should work.

src/Development/IDE/Core/OfInterest.hs
pepeiborra: Do these improvements still apply? If so, I can add a TODO comment if you or @cocreature can be more precise about the nature of the improvements.

src/Development/IDE/Core/Shake.hs
pepeiborra: It's not used anywhere though. Is the idea to garbage collect diagnostics for files that are no longer in the transitive closure of interest? Maybe the right thing to do then is to open a ticket to implement such behaviour (or at least to discuss whether that's the right thing to do).

src/Development/IDE/Core/Shake.hs
pepeiborra: done

src/Development/IDE/Core/Shake.hs
pepeiborra: done

src/Development/IDE/Core/Shake.hs
ndmitchell: https://github.com/digital-asset/ghcide/issues/123 already tracks the issue. It's not got much detail to it yet though. We definitely want this feature, as its absence has bitten me in the last few days.

pepeiborra: I've opened https://github.com/digital-asset/ghcide/issues/580

src/Development/IDE/Core/Shake.hs
```diff
@@ -377,57 +386,114 @@ shakeProfile :: IdeState -> FilePath -> IO ()
...
+        (_, wrapUp) <- wait workThread
+        wrapUp
+
+    -- 'runInShakeSession' is used to append work in this Shake session
+    --  The session stays open until 'cancelShakeSession' is called
+    --  This should only be necessary iff the (virtual) filesystem has changed
+    let runInShakeSession :: forall a . Action a -> IO (IO a)
+        runInShakeSession act = do
+              res <- newEmptyMVar
+              atomically $ writeTQueue actionQueue (act >>= liftIO . putMVar res)
```
cocreature, line 490: If `act` throws an exception you will never fill the MVar. Is that intentional? It seems like this could lead to a deadlock.

src/Development/IDE/Core/Shake.hs
cocreature: Please keep this and related functions for now, we still use this in `daml`.

src/Development/IDE/Core/FileStore.hs
```diff
@@ -165,17 +165,6 @@ fileStoreRules vfs = do
     getModificationTimeRule vfs
     getFileContentsRule vfs
 
-
--- | Notify the compiler service that a particular file has been modified.
---   Use 'Nothing' to say the file is no longer in the virtual file system
---   but should be sourced from disk, or 'Just' to give its new value.
-setBufferModified :: IdeState -> NormalizedFilePath -> Maybe T.Text -> IO ()
```
cocreature, line 172: Please keep this, we use it for tests in daml.

src/Development/IDE/Core/Shake.hs
```diff
@@ -314,22 +328,17 @@ shakeOpen getLspId eventer logger debouncer shakeProfileDir (IdeReportProgress r
...
     return IdeState{..}
 
-lspShakeProgress :: Hashable a => IO LSP.LspId -> (LSP.FromServerMessage -> IO ()) -> Var (HMap.HashMap a Int) -> IO ()
-lspShakeProgress getLspId sendMsg inProgress = do
+lspShakeProgress :: Hashable a => IdeTesting -> IO LSP.LspId -> (LSP.FromServerMessage -> IO ()) -> Var (HMap.HashMap a Int) -> IO ()
+lspShakeProgress (IdeTesting ideTesting) getLspId sendMsg inProgress = do
     -- first sleep a bit, so we only show progress messages if it's going to take
     -- a "noticable amount of time" (we often expect a thread kill to arrive before the sleep finishes)
-    sleep 0.1
+    unless ideTesting $ sleep 0.1
```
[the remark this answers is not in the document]
pepeiborra: There are two tests that expect progress notifications, so I think this change is useful.

src/Development/IDE/Core/Service.hs
```diff
@@ -85,21 +86,21 @@ shutdown = shakeShut
...
     bar <- newBarrier
-    res <- shakeRun ide [do v <- action; liftIO $ signalBarrier bar v; return v]
+    res <- shakeEnqueue ide (do v <- action; liftIO $ signalBarrier bar v; return v)
     -- shakeRun might throw an exception (either through action or a default rule),
     -- in which case action may not complete successfully, and signalBarrier might not be called.
     -- Therefore we wait for either res (which propagates the exception) or the barrier.
     -- Importantly, if the barrier does finish, cancelling res only kills waiting for the result,
     -- it doesn't kill the actual work
-    fmap fromEither $ race (head <$> res) $ waitBarrier bar
+    fmap fromEither $ race res $ waitBarrier bar
```
pepeiborra, line 95: I wonder if we can simplify `runAction` further. Does anyone remember what's the purpose of the Barrier here? Perhaps  it's not  needed anymore with `shakeEnqueue`.

src/Development/IDE/Core/Shake.hs
pepeiborra: I've pushed a change to propagate the exception

src/Development/IDE/Core/FileStore.hs
pepeiborra: I've rolled back all the dead code deletes

src/Development/IDE/Core/Service.hs
ndmitchell: shakeRun was asyncronous so the barrier was required to turn it back into a synchronous operation. It seems runInShakeSession already has an IO that can be joined to get the value back, so I'd do that.

src/Development/IDE/Core/Shake.hs
```diff
@@ -377,57 +395,115 @@ shakeProfile :: IdeState -> FilePath -> IO ()
...
+
+    -- 'runInShakeSession' is used to append work in this Shake session
+    --  The session stays open until 'cancelShakeSession' is called
+    --  This should only be necessary iff the (virtual) filesystem has changed
+    let runInShakeSession :: forall a . Action a -> IO (IO a)
+        runInShakeSession act = do
+              res <- newEmptyMVar
+              let act' = actionCatch @SomeException (Right <$> act) (pure . Left)
+              atomically $ writeTQueue actionQueue (act' >>= liftIO . putMVar res)
+              return (takeMVar res >>= either throwIO return )
```
ndmitchell, line 501: I would have used a barrier here for res - makes it clear the MVar is usually empty and then written to once. That might just be my style though (I hate that MVar's mean one of 3 distinct things depending on their usage, so always pick one of my three wrappers - https://neilmitchell.blogspot.com/2012/06/flavours-of-mvar_04.html

pepeiborra: I never saw the point of those concurrency abstractions until now, thanks for the link! Now I understand where they are coming from and see the value in them. 

That said, when I look at the implementation details, I'm back firmly in the `MVar` camp:
```
newtype Barrier a = Barrier (Var (Either (MVar ()) a))
```
I'm just too terrified of concurrent code to trust something like that :sweat_smile:

ndmitchell: I should fix that. I wanted the semantics that you could only signal once, but you could wait multiple times. With older GHC that was impossible to achieve directly on MVar. Now it is possible I'll put it back to being a nice simple MVar wrapper.

src/Development/IDE/Core/Service.hs
pepeiborra: it's done

src/Development/IDE/Core/Shake.hs
ndmitchell: See https://github.com/ndmitchell/extra/commit/98c2a83585d2ca0a9d961dd241c4a967ef87866a - now all the concurrency wrappers are very simple zero-cost zero-complexity over MVar - but guarantee that you can't leave it empty (Var), or can't fill it more than once (Barrier).

pepeiborra: That looks much better, and I can happily switch to a Barrier now

pepeiborra:
The benchmark suite is already paying off, it found a race condition that makes the code action experiment fail. I'll look into it further

pepeiborra:
Rebased and benchmarked again

# BEFORE
```
TOTAL hover = 3.92s (10 repetitions)
TOTAL getDefinition = 3.66s (10 repetitions)
TOTAL documentSymbols = 1.47s (100 repetitions)
TOTAL documentSymbols after edit = 3.30s (100 repetitions)
TOTAL completions after edit = 5.87s (10 repetitions)
TOTAL code actions = 0.64s (10 repetitions)
TOTAL code actions after edit = 4.78s (10 repetitions)
```
# AFTER
```
TOTAL hover = 0.57s (10 repetitions)
TOTAL getDefinition = 1.00s (10 repetitions)
TOTAL documentSymbols = 0.98s (100 repetitions)
TOTAL documentSymbols after edit = 3.14s (100 repetitions)
TOTAL completions after edit = 5.78s (10 repetitions)
TOTAL code actions = 0.19s (10 repetitions)
TOTAL code actions after edit = 5.04s (10 repetitions)

```

src/Development/IDE/Core/Shake.hs
```diff
@@ -379,57 +397,115 @@ shakeProfile :: IdeState -> FilePath -> IO ()
...
+        workRun restore = withAsync progressRun $ \progressThread -> do
+          let acts' =
+                [ [] <$ pumpAction
+                , parallel acts
+                    -- For the purposes of progress reporting, only 'acts' matters.
+                    -- When done, cancel the progressThread to indicate completion
+                    <* liftIO (cancel progressThread)
+                ]
+          res <- try (restore $ shakeRunDatabaseProfile shakeProfileDir shakeDb acts')
+          runTime <- start
```
mpickering, line 471: There is an existing bug with this logic as it includes the time taken to write the shake profile.

src/Development/IDE/Core/Service.hs
```diff
@@ -83,23 +81,14 @@ shutdown = shakeShut
...
-    bar <- newBarrier
-    res <- shakeRun ide [do v <- action; liftIO $ signalBarrier bar v; return v]
-    -- shakeRun might throw an exception (either through action or a default rule),
-    -- in which case action may not complete successfully, and signalBarrier might not be called.
-    -- Therefore we wait for either res (which propagates the exception) or the barrier.
-    -- Importantly, if the barrier does finish, cancelling res only kills waiting for the result,
-    -- it doesn't kill the actual work
-    fmap fromEither $ race (head <$> res) $ waitBarrier bar
-
+runAction ide action = join $ shakeEnqueue ide action
```
mpickering, line 84: Nice. My patch also has this!

mpickering:
I have been thinking hard about this patch for the last few weeks and how it should interact with the `ShakeQueue` idea that we had.  I think that both ideas would work well together.

What this patch does offer:

* Less redundant work recomputing information if nothing about the file system state changed.
* Less bad behaviour than as invoked by the batching mechanism, for example, long blocks whilst a big batch is being processed. 

What this patch does not offer:

* Any guarantee about prompt response to user query.
* Any guarantee that a queued action will be eventually completed.

Therefore it seems like the sensible design is this:

* Use Pepe's patch  as-is.
* Remove the queue aspect of my patch but keep a set of currently live actions in `ShakeSession` so that when any running actions are cancelled they can be requeued. This is really important for reducing the overall amount of work done by the IDE and means that it is safe to remove `kick`. 
* Keep the `IdeAction` portion of my patch, actions run in the `IdeAction` rather than the `Action` monad are intended to never block at the expense of either using old information or failing to return any information. Zubin is rewriting all these parts anyway at the moment. Queries are responded to quickly but then a request is send to shake to make sure the information is still up-to-date. Using stale information is critical in the case of completions for example, which this patch does not improve on. 

Questions for Pepe:

* Is there any way to tell shake priorities of certain actions already? For example, I want the `GetSpansInfo` rule to be dealt with before a non-critical typecheck of a parent module is triggered. 

@pepeiborra How does that sound? From my perspective the guarantee that queued actions will be completed is critical to implementing features such as initially loading a project and typechecking the parents of a modified module.

wz1000:
From what I can see, this patch will be easy to integrate with the work @mpickering and I have been doing, and will also simplify it quite a bit.

pepeiborra:
Sounds great. I agree that in the end state we may not need the external queue but we definitely need the external 'IdeAction', albeit possibly only for completions. 

For re-enqueueing cancelled actions, I wonder if this will still be desired, as this patch makes the distinction between user actions and file system actions more visible, allowing for finer cancellation policies. 

For priorities, I wonder if we need them at all. User actions always go through the fast path, and there's never more than one user action in flight since user interactions are single-threaded. Therefore there's already an implicit prioritization in place. 

For `kick`, my feeling is that we should reevaluate what further changes are needed once this patch goes through.

mpickering:
If you don't use an IdeAction for hover and any user action then they can still block for an unbounded amount of time before returning an answer, even with this patch.

pepeiborra:
> If you don't use an IdeAction for hover and any user action then they can still block for an unbounded amount of time before returning an answer, even with this patch.

Sure. But it's worth noting that there are two reasons to go through `ideAction`:

1. Avoid blocking
2. Avoid unnecessary interruptions

Reason 2 is removed now, and therefore it's worth reevaluating whether using the external `ideAction` is still worth it, as it could result in stale data.

pepeiborra:
The multi test failure in 8.6.5 is a real issue. What seems to happen:

    The first module is opened, and a new Shake session is started to type check it
    The second module is opened, and the Shake session is restarted to type check it, with a new GHC Session computed for the union of components
    The get-definition request sometimes fails, presumably because the dependent module is loaded from an interface file.

If step 3 fails, it can be fixed by manually forcing a Shake session restart by e.g. making an edit. @mpickering is it possible that the GhcSession rule needs to be run twice immediately after adding a new component ?

src/Development/IDE/Core/Shake.hs
pepeiborra: It's only an issue when shake profiling is on, right? In which case the timings are inaccurate anyway

mpickering:
@pepeiborra I think this is a timing issue if the goto definition request comes in before the new session is finished being set up. I don't understand your question though. Perhaps we should force a new shake session when a new component is added to the cache?

pepeiborra:
Is it possible (and does it make sense) to block until the new session is ready? If not, restarting the Shake session when a new component is added to the cache should do the trick.

mpickering:
Not possible I don't think with the current wild-west approach to cancellation nor desirable as we still want to respond to requests even with stale information whilst the cradles are being reloaded (ie for hover).

cocreature:
> as we still want to respond to requests even with stale information whilst the cradles are being reloaded (ie for hover).

For anything else in this PR we still report up2date information so I it seems to make sense to me to treat a new cradle being loaded the same as a file being modified and kill the current session. Let’s tackle reporting outdated info separately.

pepeiborra:
I have pushed changes to restart the Shake session when a new component added. For this to work well (and pass the test suite), I had to implement re-enqueueing of cancelled user actions too.

pepeiborra:
Nope, it's still failing occasionally, although I can no longer repro in VS Code

pepeiborra:
Bumping the delay from 5 to 6 seconds in the multi2 test makes it always pass locally.

src/Development/IDE/Core/Shake.hs
```diff
@@ -222,13 +236,25 @@ type IdeRule k v =
...
+
 -- | A Shake database plus persistent store. Can be thought of as storing
 --   mappings from @(FilePath, k)@ to @RuleResult k@.
 data IdeState = IdeState
-    {shakeDb :: ShakeDatabase
-    ,shakeAbort :: MVar (IO ()) -- close whoever was running last
-    ,shakeClose :: IO ()
-    ,shakeExtras :: ShakeExtras
+    {shakeDb         :: ShakeDatabase
+    ,shakeSession    :: MVar ShakeSession
```
mpickering, line 255: Why is `shakeSession` wrapped in an `MVar` if you are already using a `TQueue`?

src/Development/IDE/Core/Shake.hs
```diff
@@ -379,57 +403,126 @@ shakeProfile :: IdeState -> FilePath -> IO ()
...
+              let act' = actionCatch @SomeException (Right <$> act) (pure . Left)
+              atomically $ writeTQueue actionQueue (act' >>= liftIO . signalBarrier res)
+              return (waitBarrier res >>= either throwIO return)
+
+    --  Cancelling is required to flush the Shake database when either
+    --  the filesystem or the Ghc configuration have changed
+        cancelShakeSession = do
+            cancel workThread
+            atomically $ do
+                q <- flushTQueue actionQueue
```
mpickering, line 521: Can you explain what is going on here? How are things requeued?

src/Development/IDE/Core/Shake.hs
```diff
@@ -379,57 +403,126 @@ shakeProfile :: IdeState -> FilePath -> IO ()
...
+--
+--   Appropriate for user actions other than edits.
+shakeEnqueue :: IdeState -> Action a -> IO (IO a)
+shakeEnqueue IdeState{shakeSession} act =
+    withMVar shakeSession $ \s -> runInShakeSession s act
+
+-- Set up a new 'ShakeSession' with a set of initial system and user actions
+-- Will crash if there is an existing 'ShakeSession' running.
+-- Progress is reported only on the system actions.
+-- Only user actions will get re-enqueued
```
mpickering, line 452: Not sure this is the right way around. It is really the system actions we want to requeue in the end I think. This distinction seems premature anyway.

src/Development/IDE/Core/Shake.hs
pepeiborra: The `TQueue` cannot be closed, so the `MVar` is needed to ensure that no more actions are added while the session is being restarted. It's likely that this can be simplified further, this is mostly an iteration on top of the previous design.

src/Development/IDE/Core/Shake.hs
pepeiborra: 'system' and 'user' actions are not the best names, but are the best I could come up late at night, sorry about that.

- system actions: typechecks. We don't want to reenqueue typechecks since a fresh set of them are always provided when restarting the session.

- user actions: everything else. They must always be completed (unless the lsp-client cancels them) or the `clientMsgChan` thread will deadlock.

src/Development/IDE/Core/Shake.hs
pepeiborra: `cancelShakeSession` returns a list of the pending tasks, which are then fed back into the new session in line 438

PR #555, opened 2020-05-09
RFC - ShakeSession and shakeRunGently (2)

pepeiborra:
This Request For Comments builds on top of https://github.com/digital-asset/ghcide/pull/554 addressing the performance of completions.  

`shakeRunGently` enabled massive speed ups for:

- Hover
- Code Actions

But not for completions, which are usually requested immediately after a file
edit, which in turn triggers a non gentle `shakeRun`, which resets the
`ShakeSession`, and therefore `getCompletions` finds a cold `ShakeSession` and is slow
as a result.

The obvious question then, is why not make every `shakeRun` a gentle one?
The non gentle `shakeRun` disappears, and with it the ability to cancel Shake actions
Is cancelling actions an important feature?

- We can now run multiple actions concurrently, so it is no longer necessary to
  wait until the current action finishes.
  This removes the need for cancellation in user triggered actions, like hover, code
  actions and completions.
- Thanks to recent changes in ghcide, most actions have become cheap enough.
  I wonder if this is enough to remove the need for cancelling typechecks


pepeiborra:
This is a work in progress. Diagnostics, Shake profiling, and possibly other things, need further work

PR #557, opened 2020-05-09
Fix 3 space leaks and refactoring of PositionMapping

mpickering:
Please review this commit by commit. 

The first three commits are just about fixing space leaks.

The fourth commit is a comment warning about not unfixing the leak.

The fifth commit is a refactoring which makes the correctness of the first commit more obvious but seemed distracting to include in the same commit.

### DIFF
--- a/src/Development/IDE/Core/PositionMapping.hs
+++ b/src/Development/IDE/Core/PositionMapping.hs
@@ -5,6 +5,7 @@ module Development.IDE.Core.PositionMapping
   , toCurrentRange
   , fromCurrentRange
   , applyChange
+  , composeMapping
   , idMapping
   -- toCurrent and fromCurrent are mainly exposed for testing
   , toCurrent
@@ -31,10 +32,17 @@ fromCurrentRange mapping (Range a b) =
 idMapping :: PositionMapping
 idMapping = PositionMapping Just Just
 
+-- | Compose two position mappings. Composes in the same way as function
+-- composition (ie the second argument is applyed to the position first).
+composeMapping :: PositionMapping -> PositionMapping -> PositionMapping
+composeMapping (PositionMapping to1 from1) (PositionMapping to2 from2) =
+  PositionMapping (to1 <=< to2)
+                  (from1 >=> from2)
+
 applyChange :: PositionMapping -> TextDocumentContentChangeEvent -> PositionMapping
-applyChange posMapping (TextDocumentContentChangeEvent (Just r) _ t) = PositionMapping
-    { toCurrentPosition = toCurrent r t <=< toCurrentPosition posMapping
-    , fromCurrentPosition = fromCurrentPosition posMapping <=< fromCurrent r t
+applyChange PositionMapping{..} (TextDocumentContentChangeEvent (Just r) _ t) = PositionMapping
+    { toCurrentPosition = \x -> toCurrent r t =<< toCurrentPosition x
+    , fromCurrentPosition = \x -> fromCurrentPosition  =<< fromCurrent r t x
     }
 applyChange posMapping _ = posMapping
 
--- a/src/Development/IDE/Core/PositionMapping.hs
+++ b/src/Development/IDE/Core/PositionMapping.hs
@@ -2,11 +2,15 @@
 -- SPDX-License-Identifier: Apache-2.0
 module Development.IDE.Core.PositionMapping
   ( PositionMapping(..)
+  , fromCurrentPosition
+  , toCurrentPosition
+  , PositionDelta(..)
+  , addDelta
+  , mkDelta
   , toCurrentRange
   , fromCurrentRange
   , applyChange
-  , composeMapping
-  , idMapping
+  , zeroMapping
   -- toCurrent and fromCurrent are mainly exposed for testing
   , toCurrent
   , fromCurrent
@@ -15,11 +19,24 @@ module Development.IDE.Core.PositionMapping
 import Control.Monad
 import qualified Data.Text as T
 import Language.Haskell.LSP.Types
+import Data.List
 
-data PositionMapping = PositionMapping
-  { toCurrentPosition :: !(Position -> Maybe Position)
-  , fromCurrentPosition :: !(Position -> Maybe Position)
+-- The position delta is the difference between two versions
+data PositionDelta = PositionDelta
+  { toDelta :: !(Position -> Maybe Position)
+  , fromDelta :: !(Position -> Maybe Position)
   }
+
+fromCurrentPosition :: PositionMapping -> Position -> Maybe Position
+fromCurrentPosition (PositionMapping pm) = fromDelta pm
+
+toCurrentPosition :: PositionMapping -> Position -> Maybe Position
+toCurrentPosition (PositionMapping pm) = toDelta pm
+
+-- A position mapping is the difference from the current version to
+-- a specific version
+newtype PositionMapping = PositionMapping PositionDelta
+
 
 toCurrentRange :: PositionMapping -> Range -> Maybe Range
 toCurrentRange mapping (Range a b) =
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Core/PositionMapping.hs
```diff
@@ -29,20 +46,33 @@ fromCurrentRange :: PositionMapping -> Range -> Maybe Range
...
-    , fromCurrentPosition = \x -> fromCurrentPosition  =<< fromCurrent r t x
+composeDelta :: PositionDelta
+                -> PositionDelta
+                -> PositionDelta
+composeDelta (PositionDelta to1 from1) (PositionDelta to2 from2) =
+  PositionDelta (to1 <=< to2)
+                (from1 >=> from2)
+
+idDelta :: PositionDelta
+idDelta = PositionDelta Just Just
```
pepeiborra, line 62: Why not define `Monoid` and `Semigroup` instances for `PositionDelta`?

src/Development/IDE/Core/Shake.hs
```diff
@@ -536,7 +536,7 @@ usesWithStale key files = do
 
 withProgress :: (Eq a, Hashable a) => Var (HMap.HashMap a Int) -> a -> Action b -> Action b
 withProgress var file = actionBracket (f succ) (const $ f pred) . const
-    where f shift = modifyVar_ var $ return . HMap.alter (Just . shift . fromMaybe 0) file
+    where f shift = modifyVar_ var $ \x -> return (HMap.alter (Just . shift . fromMaybe 0) file x)
```
pepeiborra: The linked blog post by @nomeata explains why this can lead to performance gains, but it isn't clear that it fixes a space leak

src/Development/IDE/Core/Shake.hs
```diff
@@ -97,9 +97,11 @@ data ShakeExtras = ShakeExtras
     ,publishedDiagnostics :: Var (HMap.HashMap NormalizedUri [Diagnostic])
     -- ^ This represents the set of diagnostics that we have published.
     -- Due to debouncing not every change might get published.
-    ,positionMapping :: Var (HMap.HashMap NormalizedUri (Map TextDocumentVersion PositionMapping))
+    ,positionMapping :: Var (HMap.HashMap NormalizedUri (Map TextDocumentVersion (PositionMapping, PositionMapping)))
     -- ^ Map from a text document version to a PositionMapping that describes how to map
     -- positions in a version of that document to positions in the latest version
```
pepeiborra, line 102: Unrelated to this change, but is `positionMapping` ever flushed? It's a space leak that grows with every edit, isn't it?

src/Development/IDE/Core/Shake.hs
mpickering: I profiled a session where I repeatedly modified the file and could observe this function closure growing in the profile. I eta-expanded and the band was no longer present in the profile. So I agree it's not clear why this fixes the leak I don't think it should block merging.

src/Development/IDE/Core/Shake.hs
mpickering: Yes it grows every edit, I have some thoughts about doing some garbage collection but now the growth is only linear it is not as serious a problem. We don't end up with 1gb of residency just from the position map. 

For example, we could remove the whole position map from removing a file from FOI as well as flush out part of the `Values` map.

src/Development/IDE/Core/PositionMapping.hs
mpickering: It didn't occur to me in particular. I'm a GHC developer so I prefer having explicitly named functions!

src/Development/IDE/Core/Shake.hs
ndmitchell: Can we make it `evaluate (HMap.alter` - I suspect the space leak is removed because GHC manages to optimise a little bit more in some place. The `evaluate` would mean the operation guarantees to run in the IO monad and not stack up.

src/Development/IDE/Core/PositionMapping.hs
ndmitchell: Lies. If you were a real GHC developer you'd have used `;` `{` and your commit would have been rendered in comic sans.

src/Development/IDE/Core/Shake.hs
```diff
@@ -538,7 +538,7 @@ usesWithStale key files = do
 
 withProgress :: (Eq a, Hashable a) => Var (HMap.HashMap a Int) -> a -> Action b -> Action b
 withProgress var file = actionBracket (f succ) (const $ f pred) . const
-    where f shift = modifyVar_ var $ \x -> return (HMap.alter (Just . shift . fromMaybe 0) file x)
+    where f shift = modifyVar_ var $ \x -> return (HMap.alter (\x -> Just (shift (fromMaybe 0 x))) file x)
```
ndmitchell, line 544: Can we make this `Just $! (shift ...)`. I suspect this does fix the space leak, but by luck rather than guarantee. A `$!` would be a guarantee.

src/Development/IDE/Core/Shake.hs
```diff
@@ -538,6 +538,9 @@ usesWithStale key files = do
 
 withProgress :: (Eq a, Hashable a) => Var (HMap.HashMap a Int) -> a -> Action b -> Action b
 withProgress var file = actionBracket (f succ) (const $ f pred) . const
+    -- This functions are deliberately eta-expanded to avoid space leaks.
+    -- Do not remove the eta-expansion without profiling a session with at
+    -- least 1000 modifications.
```
ndmitchell, line 543: Once we are using `$!` and `evaluate` we can hang these comments off more solid guarantees.

ndmitchell:
Could we get a ticket for the "we reduce the space leak from quadratic to linear", since presumably eventually that linear will need fixing too.

src/Development/IDE/Core/Shake.hs
cocreature: Note that we do have some code for this in DAML https://github.com/digital-asset/daml/blob/master/compiler/damlc/daml-ide-core/src/Development/IDE/Core/Rules/Daml.hs#L946

src/Development/IDE/Core/Shake.hs
cocreature: :+1:

src/Development/IDE/Core/Shake.hs
cocreature: :+1:

mpickering:
It may take me a while to get around to making these changes as I will have to profile ghcide again in order to verify the suggestions work correctly.

mpickering:
We are somewhat blocked here because I don't want to spend any more time profiling, when I know these changes fix the leaks, but you don't want to merge because there is a suggestion other ways to fix them would be preferable.  

My preference is to merge the patch as-is and make tickets for the other suggested changes which can be picked up by other people. It is unfortunate to just leave these leaks unfixed whilst we are in this impasse. The script I used to profile is in my repo - https://github.com/mpickering/ghcide/blob/hls/test-leak

cocreature:
Sounds sensible, still need to fix the CI errors though before we can merge it.

ndmitchell:
The fixes as written are inherently fragile, relying on subtle compiler invariants, so I suspect they don't fix it on all versions and won't fix it if we breathe on the code funny. However, the points you've identified are clearly space leaks, and based on my history of hunting space leaks I can understand exactly why they are space leaks, and what the robust fix is - which I put in the comment.

My recommendation would be to switch to the robust versions, which I'd give a 90%+ probability of fixing now and continuing to fix in the future. If you're not happy with doing that in this patch I suggest I follow up with the changes I identified, but non-space-leak tested, and we merge that. I agree that retesting would be ideal, but without a CI infrastructure that spots these leaks robustly, we're somewhat flying blind and hopeful anyway, so I think anything we do is not ideal.

cocreature:
Happy to merge the more robust fixes separately.

PR #558, opened 2020-05-09
Refactor rawDependencyInformation

mpickering:
There are two reasons why this patch is good:

1. We remove the special case of the initial module from the dependency
search. It is now treated uniformly like the rest of the modules.

2. rawDependencyInformation can now take a list of files and create
dependency information for all of them. This isn't currently used but on
my fork we have a rule which gets the dependency information for the
whole project in order to create a module graph.

It seemed simplest to upstream this part first, which is already a
strict improvement to make the overal patch easier to review.

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -28,8 +28,8 @@ module Development.IDE.Core.Rules(
 
 import Fingerprint
 
-import Data.Binary
-import Data.Bifunctor (second)
+import Data.Binary hiding (get, put)
+import Data.Bifunctor (first, second)
 import Control.Monad.Extra
 import Control.Monad.Trans.Class
 import Control.Monad.Trans.Maybe
@@ -46,13 +46,11 @@ import Development.IDE.Types.Location
 import Development.IDE.GHC.Compat hiding (parseModule, typecheckModule)
 import Development.IDE.GHC.Util
 import Development.IDE.GHC.WithDynFlags
-import Data.Coerce
 import Data.Either.Extra
 import Data.Maybe
 import           Data.Foldable
 import qualified Data.IntMap.Strict as IntMap
 import Data.IntMap.Strict (IntMap)
-import qualified Data.IntSet as IntSet
 import Data.List
 import Data.Ord
 import qualified Data.Set                                 as Set
@@ -70,10 +68,12 @@ import GHC.Generics(Generic)
 import qualified Development.IDE.Spans.AtPoint as AtPoint
 import Development.IDE.Core.Service
 import Development.IDE.Core.Shake
-import Development.Shake.Classes
+import Development.Shake.Classes hiding (get, put)
 import Control.Monad.Trans.Except (runExceptT)
 import Data.ByteString (ByteString)
 import Control.Concurrent.Async (concurrently)
+
+import Control.Monad.State
 
 -- | This is useful for rules to convert rules that can only produce errors or
 -- a result into the more general IdeResult type that supports producing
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Core/Rules.hs
```diff
@@ -29,6 +29,7 @@ module Development.IDE.Core.Rules(
 import Fingerprint
 
 import Data.Binary
+import Util
```
alanz, line 32: Where does this come from?

mpickering: The `ghc` package.

mpickering:
Any chance of a review on this?

pepeiborra:
This looks good to me, but I only have a very superficial understanding of what this code is doing. It is a bit dense and hard to follow, and this might be a good chance to improve its readability while you still remember how it works.

mpickering:
@pepeiborra I think that's a fair point but my opinion of the new code is that it is *easier* to understand than the old code so it is still an improvement. It did take me at least a few hours to do this refactoring so if you have any specific parts you think could be clearer then I can try to improve that.

src/Development/IDE/Core/Rules.hs
```diff
@@ -253,51 +252,51 @@ getLocatedImportsRule =
...
+              -- Get NFPs of the imports which have corresponding files
+              let (no_file, with_file) = splitImports modImports
+                  (mns, ls) = unzip with_file
+              (fids, d', ss'') <- go (map artifactFilePath ls) (rawDepInfo', ss')
+              let moduleImports' = map (,Nothing) no_file ++ zipEqual "raw_dep" mns (map Just fids)
+              let rawDepInfo'' = insertImport fId (Right $ ModuleImports moduleImports' pkgImports) d'
+              (fids, r1, r2) <- go fs (rawDepInfo'', ss'')
+              return (fId : fids, r1, r2)
+
+    splitImports :: [(Located ModuleName, Maybe ArtifactsLocation)]
```
pepeiborra, line 313:
```suggestion
    -- Split in (package imports, local imports)
    splitImports :: [(Located ModuleName, Maybe ArtifactsLocation)]
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -253,51 +252,51 @@ getLocatedImportsRule =
...
+                 -> ([Located (ModuleName)], [(Located ModuleName, ArtifactsLocation)])
+    splitImports [] = ([], [])
+    splitImports ((m, k):is) = let (ns, ls) = splitImports is
+                               in case k of
+                                    Nothing -> (m:ns, ls)
+                                    Just a  -> (ns, (m, a):ls)
```
pepeiborra, line 319:
```suggestion
    splitImports = foldr splitImportsLoop ([],[])
    
    splitImportsLoop (import, Nothing) (ns,_) = (import:ns, ls)
    splitImportsLoop (import, Just artifact) (_, ls) = (ns, (import,artifact) : ls)
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -253,51 +252,51 @@ getLocatedImportsRule =
...
-                    -- Files that we haven’t seen before are added to the queue.
-                    let newFiles =
-                            IntSet.fromList (coerce $ mapMaybe snd modImports')
-                            IntSet.\\ IntMap.keysSet (rawImports rawDepInfo)
-                    let rawDepInfo' = insertImport fId (Right $ ModuleImports modImports' pkgImports) rawDepInfo
-                    go (newFiles `IntSet.union` fs)
-                       (rawDepInfo' { rawPathIdMap = pathIdMap }, ss')
-
-
+    go :: [NormalizedFilePath] -> (RawDependencyInformation, IntMap ArtifactsLocation)
```
pepeiborra: comment describing what `go` is doing

src/Development/IDE/Core/Rules.hs
```diff
@@ -253,51 +252,51 @@ getLocatedImportsRule =
...
+              (fids, r1, r2) <- go fs (rawDepInfo'', ss)
+              return (fId: fids, r1, r2)
+            Just (modImports, pkgImports) -> do
+              -- Get NFPs of the imports which have corresponding files
+              let (no_file, with_file) = splitImports modImports
+                  (mns, ls) = unzip with_file
+              (fids, d', ss'') <- go (map artifactFilePath ls) (rawDepInfo', ss')
+              let moduleImports' = map (,Nothing) no_file ++ zipEqual "raw_dep" mns (map Just fids)
+              let rawDepInfo'' = insertImport fId (Right $ ModuleImports moduleImports' pkgImports) d'
+              (fids, r1, r2) <- go fs (rawDepInfo'', ss'')
```
pepeiborra, line 310: add a comment explaining why this second recursive call is needed

src/Development/IDE/Core/Rules.hs
```diff
@@ -253,51 +252,51 @@ getLocatedImportsRule =
...
+              let rawDepInfo'' = insertImport fId (Left ModuleParseError) rawDepInfo'
+              (fids, r1, r2) <- go fs (rawDepInfo'', ss)
+              return (fId: fids, r1, r2)
+            Just (modImports, pkgImports) -> do
+              -- Get NFPs of the imports which have corresponding files
+              let (no_file, with_file) = splitImports modImports
+                  (mns, ls) = unzip with_file
+              (fids, d', ss'') <- go (map artifactFilePath ls) (rawDepInfo', ss')
+              let moduleImports' = map (,Nothing) no_file ++ zipEqual "raw_dep" mns (map Just fids)
+              let rawDepInfo'' = insertImport fId (Right $ ModuleImports moduleImports' pkgImports) d'
```
pepeiborra, line 308:
```suggestion
              let descriptiveName = ModuleImports moduleImports' pkgImports
                  rawDepInfo'' = insertImport fId (Right descriptiveName) d'
```

src/Development/IDE/Core/Rules.hs
pepeiborra: also, the pattern of calling `go` and consing on the first elemtn of the result tuple is used twice, maybe extract into a helper function

mpickering:
I just added quite a few comments. Does that help?

src/Development/IDE/Core/Rules.hs
ndmitchell: The GHC package exporting Util was never a good idea... Can we fix it upstream...

src/Development/IDE/Import/FindImports.hs
```diff
@@ -52,6 +54,12 @@ instance NFData Import where
   rnf (FileImport x) = rnf x
   rnf (PackageImport x) = rnf x
 
+modSummaryToArtifactsLocation :: NormalizedFilePath -> ModSummary -> ArtifactsLocation
+modSummaryToArtifactsLocation nfp ms = ArtifactsLocation nfp (ms_location ms) (isSource (ms_hsc_src ms))
```
ndmitchell, line 58:
```suggestion
modSummaryToArtifactsLocation nfp ms = ArtifactsLocation nfp (ms_location ms) (ms_hsc_src ms == HsSrcFile)
```

src/Development/IDE/Core/Rules.hs
ndmitchell: I'd go for `partitionEithers . map (\(m,k) -> maybe (Left m) (\k -> Right (m, k)) k)`

src/Development/IDE/Core/Rules.hs
```diff
@@ -253,51 +252,71 @@ getLocatedImportsRule =
...
+          -- Get a fresh FilePathId for the new file
+          let (fId, path_map) = getPathId al (rawPathIdMap rawDepInfo)
+          -- Insert the File into the bootmap if it's a boot module
+          let ss' = if isBootLocation al
+                      then IntMap.insert (getFilePathId fId) al ss
+                      else ss
+          let rawDepInfo' = rawDepInfo { rawPathIdMap = path_map }
+          case importsOrErr of
+            Nothing -> do
+            -- File doesn’t parse so add the module as a failure into the
```
ndmitchell, line 289: The use of smart quotes in comments in source bugs me. You know this will cause a unicode issue one day, and all for a pretty comment...

cocreature: I’m happy to break setups that still cannot handle unicode. It’s 2020 :slightly_smiling_face: 
This comment is also not new (and probably introduced by me) and it hasn’t caused any issues in the last year or however long it was here.

src/Development/IDE/Core/Rules.hs
```diff
@@ -253,51 +252,71 @@ getLocatedImportsRule =
...
+      -- First check to see if we have already processed the FilePath
+      -- If we have, just return its Id but don't update any of the state.
+      | Just fid <- lookupPathToId (rawPathIdMap rawDepInfo) f = do
+          (fids, r, bms) <- go fs a
+          return (fid: fids, r, bms)
+      -- Otherwise, we need to process its imports.
+      | otherwise = do
+          al <- modSummaryToArtifactsLocation f <$> use_ GetModSummary f
+          -- Try to parse the imports of the file
+          importsOrErr <- use GetLocatedImports f
```
cocreature, line 279: Maybe that’s just me but I would find it easier to read if this was in line 286 before you match on it.

src/Development/IDE/Import/FindImports.hs
cocreature: :+1:

src/Development/IDE/Core/Rules.hs
cocreature: I find @pepeiborra’s suggestion to be the easiest to understand.

src/Development/IDE/Core/Rules.hs
cocreature: The two recursive calls somewhat bug me. Previously, we had a queue that we kept processing by popping elements and adding ones that we discover. Now, the first argument isn’t actually a queue of the items we still need to process since we temporarily replace it in the first recursive call while not including `fs`.
I don’t really have a great suggestion for how to change this sadly and I do like the rest of this PR.

mpickering:
I took my favourites of all three of your suggestions. Neil's idea to use StateT was very good. The code is probably understandable now.

src/Development/IDE/Core/Rules.hs
```diff
@@ -251,53 +251,89 @@ getLocatedImportsRule =
             Nothing -> pure (concat diags, Nothing)
             Just pkgImports -> pure (concat diags, Just (moduleImports, Set.fromList $ concat pkgImports))
 
+type RawDepM a = StateT (RawDependencyInformation, IntMap ArtifactsLocation) Action a
+
+execRawDepM :: Monad m => StateT (RawDependencyInformation, IntMap a1) m a2 -> m (RawDependencyInformation, IntMap a1)
+execRawDepM act = execStateT act
+                    (RawDependencyInformation IntMap.empty emptyPathIdMap IntMap.empty
```
cocreature, line 258: Please don’t make indentation depend on identifier length. That just results in unnecessarily large diffs in the future.

mpickering: I think you should merge this PR and fix small things like this after merging.

cocreature: Making a new PR has more overhead than fixing it here. I can’t just push to master.

cocreature: Happy to push to your branch if you don’t want to fix it yourself.

mpickering:
CI failure looks unrelated.

src/Development/IDE/Core/Rules.hs
cocreature: pushed the fix, I hope you don’t mind.

PR #566, opened 2020-05-15
Upgrade to lsp-test-0.11

bubba:
lsp-test-0.11 has replaced openDoc' with createDoc which sends out workspace/didChangedWatchedFiles notifications, so the wrapper in https://github.com/digital-asset/ghcide/pull/322 shouldn't be needed anymore

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -263,7 +263,7 @@ test-suite ghcide-tests
         haskell-lsp-types,
         network-uri,
         lens,
-        lsp-test >= 0.8,
+        lsp-test >= 0.11.0.1 && < 0.12,
         parser-combinators,
         QuickCheck,
         quickcheck-instances,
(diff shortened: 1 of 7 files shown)


hie.yaml
```diff
@@ -1 +1,6 @@
-cradle: {stack: {component: "ghcide:lib"}}
```
mpickering, line 1: Not sure you meant to commit this.

mpickering:
One comment. Otherwise looks good.

hie.yaml
bubba: Whoops, I did not. But it is useful for working on the tests. Would anyone object to a separate PR for this?

bubba:
Indeed. Whilst I was creating it I thought it would be more transparent to have lsp-test implement an actual filesystem watcher, and have the test suite create physical files

PR #567, opened 2020-05-16
Log cache dir

pepeiborra:
Sometimes one needs to clear up the cache dir manually, e.g. if GHC is unable to detect that the interface files are stale and need rebuilding. This change is aimed at making the discovery of the cache dir location a bit easier. There are two changes

1. Log the cache dir location at startup,
2. Make the cache dir local to the project, as not every lsp client makes it easy to retrieve the lsp server logs

### DIFF
--- a/exe/Rules.hs
+++ b/exe/Rules.hs
@@ -38,6 +38,7 @@ import qualified Language.Haskell.LSP.Types     as LSP
 import Data.Aeson (ToJSON(toJSON))
 import Development.IDE.Types.Logger (logDebug)
 import Util
+import System.IO (hPutStrLn, stderr)
 
 -- Prefix for the cache path
 cacheDir :: String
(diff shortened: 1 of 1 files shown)


mpickering:
Cache files should live in xdg cache dir, that is the standard location and where you should look to find the cache for any program. So please do not merge this change which puts them in the local directory.

The other change looks good.

On my branch we also have added a prefix to the cache folder so you can see which project it is coming from more easily.

pepeiborra:
A prefix sounds good to me, please merge your change too!

pepeiborra:
I have dropped the change to make the cache files project-local, since I don't have a strong view on this.

mpickering:
The changes to the interface file prefix are wrapped up in multi-component so will not be straightforward to upstream that word-for-word.

Patch LGTM. Thanks for reverting your other change,

pepeiborra:
The Windows build crashed, needs rerun.
/azp run

PR #575, opened 2020-05-20
[WIP] Set hscTarget == HscInterpreted when desugaring

### LINKED ISSUE #537: ModBreaks.modBreaks_array not initialised
Tried to use ghcide on a cabal v2 project I hadn't tried it on before, in a bunch of files I now get:

```
  ModBreaks.modBreaks_array not initialised
  CallStack (from HasCallStack):
  error, called at compiler/ghci/ByteCodeTypes.hs:173:24 in ghc:ByteCodeTypes
```

when I run `ghcide src`.

My `hie.yaml` looks like:

```yaml
cradle:
  cabal:
    component: "lib:thingy"
```

so nothing too fancy there.

bgamari:
GHC 8.10.1 and earlier will fail to produce valid breakpoint information
during desugaring unless hscTarget==HscInterpreted. Specifically the
information that it *does* produce contains a bottom (namely
BreakArray). We can end up stumbling over this bottom later in the
pipeline. Avoid this by setting hscTarget appropriately during
desugaring.

Frankly, this whole situation isn't very satisfactory. The meaning of
hscTarget is incredibly squishy; it's only via a knowledge of the
internals of the compiler that I believe this to be safe. This is an
area of GHC which probably deserves some refactoring. A single global
"target" makes very little sense in a context like ghcide's.

Fixes #537.

# To-do

 * [ ] Local testing

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -179,7 +179,12 @@ compileModule (RunSimplifier simplify) packageState deps tmr =
                 let pm = tm_parsed_module tm
                 let pm' = pm{pm_mod_summary = tweak $ pm_mod_summary pm}
                 let tm' = tm{tm_parsed_module  = pm'}
-                GHC.dm_core_module <$> GHC.desugarModule tm'
+                -- N.B. If we don't set hscTarget to HscInterpreted then GHC
+                -- will fail to produce its breakpoint state, resulting in a
+                -- bottom in the desugared Core which the interpreter may later
+                -- hit. See #537.
+                let setTarget hsc_env = hsc_env { hsc_dflags = (hsc_dflags hsc_env) { hscTarget = HscInterpreted } }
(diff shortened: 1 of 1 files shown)


bgamari:
@cocreature, unfortunately I've found that this doesn't workaround the issue. I've been meaning to build a profiled compiler to work out why this is the case. However, it will likely be a few weeks before I have time to do this.

cocreature:
> However, it will likely be a few weeks before I have time to do this.

No worries, thanks a lot for looking into it!

ndmitchell:
@bgamari are you likely to get back to this? If not, any tips to add to the original ticket, ideally so someone else can pick it up?

wz1000:
Can we work around this by simply setting `mg_modBreaks` to `Nothing` after desugaring?

wz1000:
Regardless, this should not be an issue anymore because we don't run the bytecode generator since #836. So I'm closing this for now, until we decide we need to re-introduce bytecode generation.

Issue #583, opened 2020-05-25
.hi file checks are expensive


ndmitchell:
Looking at some recordings of which operations are executed during a hover request on the Shake repo (100+ real files), there are approx 2200 rules involved. Classifying those into ones that do some work (not literally zero work) there are 375. It is definitely possible to optimise those that don't do anything to have nearly zero cost, but once I do that (ideally with benchmarks...) we'll be left with those that have some cost. They break down into:

* 1 x GetFilesOfInterest - happy to keep this one around, it's literally one, so seems fine.
* 125 x GetModificationTime in the workspace - I believe we should switch these to using the VSCode watching for them. I think that's feasible.
* 124 * GetModificationTime + 124 * GetFileExists on every .hi file in `C:\\Users\\ndmit_000\\AppData\\Local\\ghcide\\44a1c205801868d589ca26f9c844cc8d55a566b8`. These are outside the workspace so don't benefit from being remembered by VS Code. My guess is that we assume these files only change outside our control when we're not running? If so, that knowledge should be lifted into the build system, and we shouldn't tell Shake to keep polling for them.

pepeiborra:
For `GetModificationTime` to rely on the lsp notification events, is #325 what you have in mind? I implemented it in PR #373 a while ago but we didn't end up merging it due to unclear perf benefits.

mpickering:
What exactly is expensive about the check? This hasn't been something which has showed up on any profiles I have done as a performance bottleneck.

pepeiborra:
For interface files, the `GetModificationTime` checks are needed  for staleness checking, and to reload them when they are regenerated. It should be easy to remove the `GetFileExists` dependency.

ndmitchell:
@pepeiborra exactly right. My guess is that the unclear perf was down to Shake not taking advantage of the knowledge it provides. The way DoesFileExist clearing is wired in also doesn't give Shake much info, but if we change that (which I'm looking at) then GetModificationTime probably would show a benefit.

@mpickering we end up doing 1 modtime and 1 existence check per file on every hover. For GHC, I imagine that's 1000 disk operations per hover. I would have imagined you wanted to get the number of disk ops on a hover down to 0.

ndmitchell:
@pepeiborra aren't we the ones that regenerate them? Thus we don't need to check the disk to see if we did it - we have the information to hand.

pepeiborra:
[...]

PR #589, opened 2020-05-29
remove unnecessary FileExists dependency in GetHiFile

pepeiborra:
It is subsumed by the GetModificationTime dependency.

One less dependency per .hi file,
one less redundant file system access,
five fewer lines of code.

This does *not* implement all the improvements discussed in #583, it's just a minor efficiency fix.

I have not measured the performance impact either, since there is no easy-to-run standard benchmark, but I expect the impact to be very small.

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -575,26 +575,22 @@ getHiFileRule = defineEarlyCutoff $ \GetHiFile f -> do
   case sequence depHis of
     Nothing -> pure (Nothing, ([], Nothing))
     Just deps -> do
-      gotHiFile <- getFileExists hiFile
-      if not gotHiFile
-        then pure (Nothing, ([], Nothing))
-        else do
-          hiVersion  <- use_ GetModificationTime hiFile
-          modVersion <- use_ GetModificationTime f
-          let sourceModified = modificationTime hiVersion < modificationTime modVersion
-          if sourceModified
-            then do
+        mbHiVersion <- use  GetModificationTime hiFile
+        modVersion  <- use_ GetModificationTime f
+        case mbHiVersion of
+            Just hiVersion
+              | modificationTime hiVersion >= modificationTime modVersion -> do
+                session <- hscEnv <$> use_ GhcSession f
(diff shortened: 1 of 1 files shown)


pepeiborra:
I think it's not a big deal, since the `.hi` file will only be missing the very first time ghcide is started on a given folder. Ever after that, the `.hi` file will exist.

pepeiborra:
Benchmark BEFORE:
```
TOTAL hover = 2.55s
TOTAL getDefinition = 2.66s
TOTAL documentSymbols = 1.45s
TOTAL documentSymbols after edit = 2.93s
TOTAL completions after edit = 4.55s
TOTAL code actions = 3.59s
```

Benchmark AFTER:
```
TOTAL hover = 2.41s
TOTAL getDefinition = 2.52s
TOTAL documentSymbols = 1.44s
TOTAL documentSymbols after edit = 2.92s
TOTAL completions after edit = 4.42s
TOTAL code actions = 3.52s
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -142,9 +141,12 @@ getHomeHieFile f = do
   mbHieTimestamp <- use GetModificationTime normal_hie_f
   srcTimestamp   <- use_ GetModificationTime f
 
-  let isUpToDate
-        | Just d <- mbHieTimestamp = comparing modificationTime d srcTimestamp == GT
-        | otherwise = False
+  let isUpToDate = case (modificationTime <$> mbHieTimestamp, modificationTime srcTimestamp) of
+          -- If the src file is in VFS, then we know it must have been typechecked
+          -- by now (due to kick) and therefore its .hie file is up to date
```
cocreature, line 146: Is this true, What happens if you just opened the file? I don’t see what guarantees that the .hie file has been generated before we get here.

pepeiborra: Hmm, you are right, we don't have a dependency on the `Typecheck` rule here. Looking at the rule for `GetModIface`, there are a couple of different scenarios depending on whether the module is a file of interest or not, etc. As I'm not keen on duplicating the logic here, I am going to declare a dependency on `GetModIface` as a tight overapproximation of what is needed here.

Issue #242, opened 2019-12-14
Add performance regression tests


cocreature:
It would be good to have some performance regression tests. We have a small wrapper around `lsp-tests` in DAML https://github.com/digital-asset/daml/blob/master/compiler/damlc/ide-debug-driver/README.md that makes it easy to script sessions which I’ve used successfully in the past to track down performance issues which might come in handy here.

domenkozar:
Refs #298

ndmitchell:
Would be good to include benchmarks, which might be things that run for infeasibly long to include in tests, but are great at measuring and improving things in the region of 5% or less.

ndmitchell:
Copying over @pepeiborra's comment:

> How to prove that they are capturing real workloads? Perhaps use haskell-lsp to record and replay real interactions with vs code, emacs, rather than modelling them with lsp-test

I looked at this a bit. I think testing on real world workloads may be very important, since otherwise we can't capture all the nuances of what is being watched via notifications etc.

pepeiborra:
@mpardalos - do you have plans to do this over the summer? It’s sorely needed!

mpardalos:
@pepeiborra Yes! I am planning on recording some sessions for this reason. 

I was planning on using lsp-test to run those sessions. I spoke with my GSoC mentor about this (@mpickering) and he told me that apparently lsp-test currently has problems with replaying sessions (it doesn't respect request timings) which might be a blocker for now? I couldn't find the relevant issue there so he probably has a better idea of what the problem is.

pepeiborra:
I'm working on a simple benchmark based on https://neilmitchell.blogspot.com/2020/05/fixing-space-leaks-in-ghcide.html

jneira:
@cocreature the project already has a benchmark, comparing branches and running in ci, courtesy of @pepeiborra, so only regression tests would be needed to close this?

cocreature:
@jneira yeah I think so.

pepeiborra:
Detecting regressions in allocations would be sweet. Steps to do this:

1. Enhance the benchmark outputs to include the total allocated bytes
2. Add a step to the bench CI script to check that the allocated bytes did not go up by more than a certain threshold

pepeiborra:
1 is done, but 2 is still very much needed

Issue #576, opened 2020-05-21
Ignore import lists for producing completions


pepeiborra:
Given a module header like the one below with explicit import lists, ghcide should provide completions for all of `Control.Monad` and not limit to just `join`:
```
module MyModule where

import Control.Monad (join)
```
It would be great if selecting a completion would also check and expand the corresponding import list if needed.

gdevanla:
@pepeiborra I would like to take a look at this one. Should I be starting in Completions.hs to find out where the list is getting restricted?

pepeiborra:
Completions are driven by the `LocalCompletions and `NonLocalCompletions` rules. To achieve the goal in this ticket, the `NonLocalCompletions` rule needs to be relaxed to generate completions for the unrestricted imports. This should be quite easy, the code is in:

https://github.com/haskell/ghcide/blob/master/src/Development/IDE/Plugin/Completions.hs#L55-L81

Basically, use `map unrestrictImport imps` instead of `imps`

gdevanla:
Great. Thanks for the pointer. Will review this item this week.

gdevanla:
I did some preliminary research on this. Before, I take a wrong path approaching this, I wanted to confirm my understanding. Any feedback on this is appreciated.

Currently, the import list is being restricted because we map over  `[GlobalRdrElt]` at  https://github.com/haskell/ghcide/blob/863392b9b94777a069a2a31e9b909d1ce45e93d4/src/Development/IDE/Plugin/Completions/Logic.hs#L257.

My understanding is that `GlobalRdrElt` already only has the restricted elements based on the explicit import list.

To address this issue, we cannot use `GlobalRdrElt` anymore. Instead, we need to get all available completions, for each `LImportDecl GhcPs`. To accomplish that, we have to access the parsed module `HsModule` from each `ImportDecl` and then access the `hsModExports` values to build the `CompItem`.

This essentially means, the logic in `cacheDataProducer` will change entirely. 

Is this a approach I can pursue?

Thanks!

pepeiborra:
This is not what I had in mind, almost no changes are needed if my suggestion below works.

We are in control of the import list used to generate the `[GlobalRdrElt]`, since we explicitly call `tcRnImportDecls`. So just modify the import list before calling: 

https://github.com/haskell/ghcide/blob/master/src/Development/IDE/Plugin/Completions.hs#L74

gdevanla:
Thanks. That makes sense.

gdevanla:
[...]

PR #590, opened 2020-05-30
Benchmark suite

pepeiborra:
The benchmark suite is designed to be run with `cabal bench`. It does not attempt to compare with past performance, I think that's better tackled as a separate task.

The module comment in the benchmark module contains some details on the benchmark methodology and experiments, which currently cover the basics:

- hover
- go to definition
- completions
- code actions
- document symbols

I have not made any attempts to integrate with CI since I don't know how consistent the timings would be across runs. 

Goes some way towards closing #242

### DIFF
--- a/bench/Main.hs
+++ b/bench/Main.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ExistentialQuantification #-}
+
 {- An automated benchmark built around the simple experiment described in:
 
   > https://neilmitchell.blogspot.com/2020/05/fixing-space-leaks-in-ghcide.html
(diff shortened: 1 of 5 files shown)


pepeiborra:
Example output:
```
$ cabal bench
Running 1 benchmarks...
Benchmark ghcide-bench: RUNNING...
starting test
Unpacking to bench/example/Cabal-3.2.0.0/
/Users/pepe/Dropbox/code/ghcide/dist-newstyle/build/x86_64-osx/ghc-8.10.1/ghcide-0.1.0/x/ghcide/build/ghcide/ghcide
    Alloc    Copied     Live     GC     GC      TOT      TOT  Page Flts
    bytes     bytes     bytes   user   elap     user     elap
ghcide version: 0.1.0 (GHC: 8.10.1) (PATH: /Users/pepe/Dropbox/code/ghcide/dist-newstyle/build/x86_64-osx/ghc-8.10.1/ghcide-0.1.0/x/ghcide/build/ghcide/ghcide) (GIT hash: b759359f457da747f5c4e8659cc2ce08c8441fcd)
Starting LSP server...
If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!
...
...
TOTAL hover = 2.21s
TOTAL getDefinition = 2.23s
TOTAL documentSymbols = 1.50s
TOTAL documentSymbols after edit = 3.56s
TOTAL completions after edit = 3.23s
TOTAL code actions = 2.71s
```

mpickering:
Looks useful @pepeiborra 

Once my branch starts getting merged in then some of these tests will become trivial lookups in a HashMap so will probably have to be modified to show anything useful.

bench/Main.hs
```diff
@@ -0,0 +1,283 @@
...
+exampleModulePath = "Distribution" </> "Simple.hs"
+
+data Config = Config
+  { verbose :: !Bool,
+    -- For some reason, the Shake profile files are truncated and won't load
+    shakeProfiling :: !(Maybe FilePath),
+    outputCSV :: !Bool
+  }
+
+type HasConfig = (?config :: Config)
```
ndmitchell, line 152: Cute.

bench/Main.hs
```diff
@@ -0,0 +1,283 @@
...
+examplePackage :: String
+examplePackage = examplePackageName <> "-" <> showVersion examplePackageVersion
+
+exampleModulePath :: FilePath
+exampleModulePath = "Distribution" </> "Simple.hs"
+
+data Config = Config
+  { verbose :: !Bool,
+    -- For some reason, the Shake profile files are truncated and won't load
+    shakeProfiling :: !(Maybe FilePath),
```
ndmitchell, line 148: Odd. Any idea where they are truncated? Would be keen to debug this, but will wait til it lands.

bench/Main.hs
```diff
@@ -0,0 +1,283 @@
...
+  writeFile
+    ("bench/example/" <> examplePackage <> "/hie.yaml")
+    ("cradle: {cabal: {component: " <> show examplePackageName <> "}}")
+
+  whenJust (shakeProfiling ?config) $ createDirectoryIfMissing True
+
+  -- print the path to ghcide (TODO platform independent)
+  when (verbose ?config) $ callCommand "which ghcide"
+
+duration2 :: MonadIO m => m a -> m (Seconds, a)
```
ndmitchell, line 272: `extra-1.7.2` has this as `duration`.

bench/Main.hs
pepeiborra: `.hlint.yaml` disallows `ImplicitParams`. Do you happen to know what's the minimal change to allow it in `bench/Main.hs`?
```
  # Shady extensions
  - {name: ImplicitParams, within: [ ]}
```

bench/Main.hs
pepeiborra: No idea. This might be a bug in ghcide

bench/Main.hs
```diff
@@ -0,0 +1,283 @@
...
+  e.g. code actions, and therefore the two methods above do not necessarily
+  produce the same results.
+
+ -}
+
+import Control.Applicative.Combinators
+import Control.Concurrent
+import Control.Exception.Safe
+import Control.Monad
+import Control.Monad.Extra (whenJust)
```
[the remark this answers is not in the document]
pepeiborra: I used to do that too, but these days I like letting ghcide insert imports for me, which currently means explicit imports. Same with `Data.List`. I don't mind them being explicit, but at the moment that's not totally satisfactory because completions are narrowed down to the identifiers in scope, which is quite at odds with explicit imports. I raised #576 for that.

bench/Main.hs
```diff
@@ -0,0 +1,283 @@
...
+  let ?config = config
+
+  output "starting test"
+
+  setup
+
+  runBenchmarks
+    [ ---------------------------------------------------------------------------------------
+      bench "hover" 10 $ \doc ->
+        isJust <$> getHover doc (Position 853 12),
```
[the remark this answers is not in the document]
pepeiborra: done - but note that the Cabal version is fixed and decoupled

pepeiborra:
> This is awesome!
> 
> The one feature I'd have liked was the ability to do profiling. That requires installing `ghcide` with profiling (a user step), passing `+RTS -p` when starting ghcide (easy to change below) and then arranging for `ghcide` to kill itself rather than be forcibly terminated by lsp-test, which I didn't have a good sense on how to do. Of course, this feature could definitely come later.

I implemented a nice way of allowing `ghcide` to exit cleanly which I think should work fine with profiling, although I have not checked that.

Another issue is that the benchmark invokes ghcide once per experiment, and afaik there is no way to redirect the profiling artifacts to a specific path, so the user would need to select only one experiment (or we change the test suite to run every experiment in a different folder).

bench/Main.hs
ndmitchell: Just delete that line. We ban it deliberately so we didn't accidentally add it.

ndmitchell:
When I tried sending a close to ghcide it didn't exit, merely stopped responding to messages, so worth validating (I could well have screwed it up). Running ghcide on a single benchmark at a time seems what people would want to do, so no problem with that.

Note that you might need to roll the stack.yaml files or explicitly override to get the latest version of extra included.

bench/Main.hs
pepeiborra: done

pepeiborra:
> When I tried sending a close to ghcide it didn't exit, merely stopped responding to messages, so worth validating (I could well have screwed it up). 

Maybe you didn't send a [`shutdown`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#shutdown) message *followed by* an [`exit`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#exit) message?

> Running ghcide on a single benchmark at a time seems what people would want to do, so no problem with that. 
That's right, but then more user steps are needed, because the benchmark suite does not support selecting experiments to run.

ndmitchell:
Ah, that would be it - I think I only sent a shutdown. Good to know, and even better that with this benchmark I don't need to know!

bench/Main.hs
```diff
@@ -0,0 +1,285 @@
...
+    return $ if res then t else -1
+  where
+    cmd =
+      unwords $
+        [ "ghcide",
+          "--lsp",
+          "--cwd",
+          dir,
+          "+RTS",
+          "-N4",
```
mpickering, line 249: Would be an improvement to not hardcode this to 4.

bench/Main.hs
```diff
@@ -0,0 +1,285 @@
...
+          "-S",
+          "-RTS"
+        ]
+          ++ concat
+            [ ["--shake-profiling", path]
```
mpickering, line 254: `--shake-profiling` can add a significant overhead to benchmarking. Use with caution I think.

bench/Main.hs
```diff
@@ -0,0 +1,285 @@
...
+        { logStdErr = verbose ?config,
+          logMessages = False,
+          logColor = False
+        }
+
+setup :: HasConfig => IO ()
+setup = do
+  alreadyExists <- doesDirectoryExist "bench/example"
+  when alreadyExists $ removeDirectoryRecursive "bench/example"
+  callCommand $ "cabal unpack -v0 " <> examplePackage <> " -d bench/example"
```
mpickering, line 271: `cabal get` replaced `cabal unpack` I think. https://github.com/haskell/cabal/issues/5410

bench/Main.hs
ndmitchell: Agreed. FWIW, the majority of time in profiling goes to writing the output file, which is sadly an unavoidable step.

mpickering:
Thanks Pepe. I left a few comments but nothing critical to stop the merge imo.

bench/Main.hs
pepeiborra: I shall remove it and let Shake auto adjust

bench/Main.hs
pepeiborra: It's turned off by default

bench/Main.hs
pepeiborra: Thanks, I wasn't aware of that

pepeiborra:
The multi test failure in 8.6.5 is a real issue. What seems to happen:

1. The first module is opened, and a new Shake session is started to type check it
2. The second module is opened, and the Shake session is restarted to type check it, with a new GHC Session computed for the union of components
3. The get-definition request sometimes fails, presumably because the dependent module is loaded from an interface file.

If step 3 fails, it can be fixed by manually forcing a Shake session restart by e.g. making an edit. @mpickering is it possible that the `GhcSession` rule needs to be run twice immediately after adding a new component ?

PR #592, opened 2020-05-31
Add Kakoune section to the README

aodhneine:


README.md
```diff
@@ -266,6 +266,20 @@ Add this to your coc-settings.json (which you can edit with :CocConfig):
...
+### Using with Kakoune
+
+Install [kak-lsp](https://github.com/ul/kak-lsp).
+
+Change `kak-lsp.toml` to include this:
+
+```toml
+[language.haskell]
+filetypes = ["haskell"]
+roots = ["Setup.hs", "stack.yaml", "*.cabal"]
```
cocreature, line 278: Should this not include `hie.yaml` and maybe `cabal.project`?

aodhneine: I'm not very well oriented in Haskell project structure, so I didn't knew about those files (besides, I just took example from [haskell-language-server](https://github.com/haskell/haskell-language-server#using-haskell-language-server-with-kakoune) and checked if it would work), but I think it shouldn't be a problem to add them.

PR #607, opened 2020-06-06
Add back a .ghci file

ndmitchell:
Allegedly it's possible to develop Ghcide without using Ghcid. I am yet to be convinced, and certainly I can't do it.

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -1,0 +1,25 @@
+:set -Wunused-binds -Wunused-imports -Worphans -Wunused-matches -Wincomplete-patterns
+
+:set -XBangPatterns
+:set -XDeriveFunctor
+:set -XDeriveGeneric
+:set -XGeneralizedNewtypeDeriving
+:set -XLambdaCase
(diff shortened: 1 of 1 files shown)


pepeiborra:
> Allegedly it's possible to develop Ghcide without using Ghcid. I am yet to be convinced, and certainly I can't do it.

What do you need Ghcid for?

ndmitchell:
> What do you need Ghcid for?

When Ghcide isn't working. Either because it fails to setup (which seems to be the case at the moment - failing to configure process through Cabal sandbox) or because I broke it while developing.

mpickering:
I don't think this file is generally useful, it was broken when I tried to run it before so I removed it.

ndmitchell:
It works for me, and without it, developing Ghcide is nearly impossible. Does its existence harm anyone else? If so, we could rename it slightly, and then it would be even more innocuous.

I note that the `hie.yaml` file is also broken and doesn't work (for me), although I'm happy for it to stay for now #610 🤣

PR #608, opened 2020-06-06
#573, make haddock errors warnings with the word Haddock in front

### LINKED ISSUE #573: Parse error on comment that ghc accepts
`ghcide` on current master errors out when trying to parse this:
```haskell
blademaster :: ~ » cat test.hs
main :: IO ()
main = do
  -- ^
  pure ()
```
with:
```
blademaster :: ~ » ghcide test.hs
ghcide version: 0.1.0 (GHC: 8.6.5) (PATH: /home/googleson78/.local/bin/ghcide) (GIT hash: a1cb4eb8fa27821a28b48ce05bb690d82792a18f)
Ghcide setup tester in /home/googleson78.
Report bugs at https://github.com/digital-asset/ghcide/issues

Step 1/6: Finding files to test in /home/googleson78
Found 1 files

Step 2/6: Looking for hie.yaml files that control setup
Found 1 cradle

Step 3/6, Cradle 1/1: Implicit cradle for /home/googleson78
Cradle {cradleRootDir = "/home/googleson78", cradleOptsProg = CradleAction: Default}

Step 4/6, Cradle 1/1: Loading GHC Session
Interface files cache dir: /home/googleson78/.cache/ghcide/da39a3ee5e6b4b0d3255bfef95601890afd80709

Step 5/6: Initializing the IDE

Step 6/6: Type checking the files
File:     /home/googleson78/test.hs
Hidden:   no
Range:    3:2-3:6
Source:   parser
Severity: DsError
Message:  parse error on input ‘-- ^’

Completed (1 file worked, 0 files failed)
```
But `ghc(i)` does fine:
```
blademaster :: ~ » ghc test.hs
[1 of 1] Compiling Main             ( test.hs, test.o )
Linking test ...
```
```
blademaster :: ~ » ghci test.hs
GHCi, version 8.6.5: http://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /home/googleson78/.ghci
[1 of 1] Compiling Main             ( test.hs, interpreted )
Ok, one module loaded.
>
```

ndmitchell:
Fixes #573. Downgrade Haddock errors to warnings, and prepend Haddock in front of them. Now we are manipulating Haddock warnings, its easier to dedupe them by range than trying to merge arbitrary diagnostics.

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -29,7 +29,7 @@ module Development.IDE.Core.Rules(
 import Fingerprint
 
 import Data.Binary hiding (get, put)
-import Data.Bifunctor (first, second)
+import Data.Tuple.Extra
 import Control.Monad.Extra
 import Control.Monad.Trans.Class
 import Control.Monad.Trans.Maybe
@@ -41,7 +41,7 @@ import Development.IDE.Import.DependencyInformation
 import Development.IDE.Import.FindImports
 import           Development.IDE.Core.FileExists
 import           Development.IDE.Core.FileStore        (getFileContents)
-import           Development.IDE.Types.Diagnostics
+import           Development.IDE.Types.Diagnostics as Diag
 import Development.IDE.Types.Location
 import Development.IDE.GHC.Compat hiding (parseModule, typecheckModule)
 import Development.IDE.GHC.Util
@@ -212,7 +212,9 @@ getParsedModuleRule = defineEarlyCutoff $ \GetParsedModule file -> do
                 haddockParse = do
                     (_, (!diagsHaddock, _)) <-
                         getParsedModuleDefinition hscHaddock opt comp_pkgs file contents
-                    return diagsHaddock
+                    -- Haddock diagnostics are confusing because they don't say Haddock, and are errors.
+                    -- Fix both those properties.
+                    return $ map (\(a,b,c) -> (a,b,fixHaddockWarning c)) diagsHaddock
 
             ((fingerPrint, (diags, res)), diagsHaddock) <-
                 -- parse twice, with and without Haddocks, concurrently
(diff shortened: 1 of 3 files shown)


src/Development/IDE/Core/Rules.hs
```diff
@@ -212,15 +212,29 @@ getParsedModuleRule = defineEarlyCutoff $ \GetParsedModule file -> do
...
+                    return $ map (\(a,b,c) -> (a,b,fixHaddockWarning c)) diagsHaddock
 
             ((fingerPrint, (diags, res)), diagsHaddock) <-
                 -- parse twice, with and without Haddocks, concurrently
                 -- we cannot ignore Haddock parse errors because files of
                 -- non-interest are always parsed with Haddocks
                 liftIO $ concurrently mainParse haddockParse
 
-            return (fingerPrint, (mergeDiagnostics diags diagsHaddock, res))
+            -- if you have a Haddock warning and a real warning at the same exact location, throw away the Haddock one
```
pepeiborra, line 225: do we need this special logic? I would think that a comment can only have Haddock warnings

ndmitchell: We run the parser twice. If you have `(` as your source file, it will be a parse error with both Haddock and without, so you'll get the same error twice. This dedupe ensures you only get the warning once.

ndmitchell:
I've had much more issues with haddock warnings in files of interest because I just typed `-- | ...` to comment out a guard, whereas haddock issues in dependencies hasn't cropped up as much. I see that it that case we don't parse twice, but only once - does that mean a Haddock in a dependency is a fatal error?

src/Development/IDE/Core/Rules.hs
pepeiborra: so the comment is a bit misleading, it's only non Haddock warnings that need deduping, right?

pepeiborra:
> I've had much more issues with haddock warnings in files of interest because I just typed `-- | ...` to comment out a guard, whereas haddock issues in dependencies hasn't cropped up as much. I see that it that case we don't parse twice, but only once - does that mean a Haddock in a dependency is a fatal error?

I would think so, yes, and the solution is similar - parse again without `OptHaddock`.

src/Development/IDE/Core/Rules.hs
ndmitchell: Comment updated. Is that clearer?

ndmitchell:
It feels odd that we have two locations where we do a parse with haddock/parse without dance, and in one we use concurrently, and in one we don't. We should probably do the same thing in both cases.

My guess is "the right thing" is to parse with Haddock, and if there are no errors, continue on. If there were errors, parse again without Haddock, and turn all errors that came from Haddock only into warnings. For files with parse errors it's slightly more latency, but the latency on a parse error is pretty low already, so that isn't a massive problem.

pepeiborra:
> It feels odd that we have two locations where we do a parse with haddock/parse without dance, and in one we use concurrently, and in one we don't. We should probably do the same thing in both cases.
> 
> My guess is "the right thing" is to parse with Haddock, and if there are no errors, continue on. If there were errors, parse again without Haddock, and turn all errors that came from Haddock only into warnings. For files with parse errors it's slightly more latency, but the latency on a parse error is pretty low already, so that isn't a massive problem.
> 
> @pepeiborra if that makes sense to you, I'll try that.

The reason we don't parse with `-haddock` always-on is that it throws away annotations (ref. @cocreature ), which can be troublesome. For interface files we don't care as the parsed AST is not stored, but for files of interest we probably do care.

ndmitchell:
OK, I'll try figuring out some sane abstractions to handle that mess.

cocreature:
> The reason we don't parse with `-haddock` always-on is that it throws away annotations (ref. @cocreature ), which can be troublesome. For interface files we don't care as the parsed AST is not stored, but for files of interest we probably do care.

Adding a bit more context here: GHC has `Opt_Haddock` and `Opt_KeepRawTokenStream` which are mutually exclusive. In theory, the latter preserves all information for haddock to work based on it afaik and there is an issue somewhere for making it work. In practice, haddock currently only works if you use `Opt_Haddock` which throws away information that tools like HLint rely on and which is included in `Opt_KeepRawTokenStream`.

ndmitchell:
So we still have two different behaviours. For `getModIface` we try with Haddock, and if that fails, fall back to without. Otherwise, we do both in parallel and merge them. I centralised the "merge them" code, so they aren't too different, but the weird corner cases around Haddock are pretty frustrating!

PR #609, opened 2020-06-06
ghc initialization error handling

pepeiborra:
There are a couple of error cases to handle as seen below.
Thanks @jneira for help discovering them all.
There used to be linking errors but I no longer see those after the multi-cradle patch

Non Nix
=========

The table below shows a couple of combinations of cradles and ghcide versions in a
non-Nix environment. All the version mismatches are now handled as follows:

- "Cannot satisfy package" -  `-package-id` flags referencing
  package versions not available (generally base)
- "bad interface" - tried to load an interface file created by a different version of ghc

cradle/ghcide | 8.6 | 8.8 | 8.10
--------------|-----|----|---
Cabal 8.6   | success | cannot satisfy package | cannot satisfy package
Cabal 8.8   | cannot satisfy package | success | cannot satisfy package
Cabal 8.10  | cannot satisfy package | cannot satisfy package | success
Stack 8.6   | success | bad-interface | bad-interfac-
Stack 8.8   | bad-interface | success | bad-interface
Stack 8.10  | bad-interface | bad-interface | success

Nix
=========

Because Nix redefines the libdir to point at the run-time ghc installation,
it's actually much easier to detect a version mismatch:
just compare the compile-time and run-time libdirs

Notes for reviewers
================
Ignore whitespace for a much smaller diff


pepeiborra:
this needs rerunning as the Windows build seems to have crashed

pepeiborra:
Rebased and resolved

Issue #610, opened 2020-06-06
`ghcide` fails for `ghcide` using cabal in windows


ndmitchell:
Using the latest HEAD, if I type `ghcide` in the `ghcide` directory then it fails:

```
Message:
  Failed to parse result of calling cabal
  Build profile: -w ghc-8.6.5 -O1
  In order, the following will be built (use -v for more details):
  - process-1.6.9.0 (lib:process) (requires build)
  - hsc2hs-0.68.7 (exe:hsc2hs) (requires build)
  - network-3.1.1.1 (lib:network) (requires build)
  - network-bsd-2.8.1.0 (lib) (requires build)
  - hslogger-1.3.1.0 (lib) (requires build)
  - haskell-lsp-0.22.0.0 (lib) (requires build)
  - ghcide-0.2.0 (lib) (first run)
  Starting process-1.6.9.0 (all, legacy fallback)

  Failed to build process-1.6.9.0. The failure occurred during the configure
  step.
  Build log (
  C:\Users\ndmit_000\AppData\Roaming\cabal\logs\ghc-8.6.5\process-1.6.9.0-a2d33b358c7e24516bfa0d05962b26aaf60ae0e6.log
  ):
  Configuring process-1.6.9.0...

  abnormal program termination
  C:\/Neil/ghcide/dist-newstyle/tmp/src-6052/process-1.6.9.0/configure: fork failed: no error [19]

  abnormal program termination
  C:\/Neil/ghcide/dist-newstyle/tmp/src-6052/process-1.6.9.0/configure: fork failed: no error [49]

  abnormal program termination
  C:\/Neil/ghcide/dist-newstyle/tmp/src-6052/process-1.6.9.0/configure: fork failed: no error [135]

  abnormal program termination
  C:\/Neil/ghcide/dist-newstyle/tmp/src-6052/process-1.6.9.0/configure: fork failed: no error [202]

  abnormal program termination
  C:\/Neil/ghcide/dist-newstyle/tmp/src-6052/process-1.6.9.0/configure: fork failed: no error [376]

  abnormal program termination
  C:\/Neil/ghcide/dist-newstyle/tmp/src-6052/process-1.6.9.0/configure: fork failed: no error [422]

  abnormal program termination
  C:\/Neil/ghcide/dist-newstyle/tmp/src-6052/process-1.6.9.0/configure: fork failed: no error [429]
[...]

PR #612, opened 2020-06-06
Use a better noRange

ndmitchell:
When there is a catastrophic failure to load a cradle (e.g. #610) the consequence is that every file is highlighted in red everywhere. If you want to continue developing with something like Ghcid it's super annoying to have to try and see the code under this sea of red. Solution is to treat cradle errors as being diagnostics spanning the first line of every file. It's still clear the error is there, you can still look at it, but it's not all you have to look at. There's no good reason to highlight an entire file in red.


mpickering:
LGTM!

PR #621, opened 2020-06-09
Restore Shake profiling

### LINKED ISSUE #620: --shake-profiling no longer works
Broken by #554

pepeiborra:
Fixes #620

### DIFF
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -177,15 +177,6 @@ instance Eq Key where
 instance Hashable Key where
     hashWithSalt salt (Key key) = hashWithSalt salt (typeOf key, key)
 
--- | The result of an IDE operation. Warnings and errors are in the Diagnostic,
---   and a value is in the Maybe. For operations that throw an error you
---   expect a non-empty list of diagnostics, at least one of which is an error,
---   and a Nothing. For operations that succeed you expect perhaps some warnings
---   and a Just. For operations that depend on other failing operations you may
---   get empty diagnostics and a Nothing, to indicate this phase throws no fresh
---   errors but still failed.
---
-
 data Value v
     = Succeeded TextDocumentVersion v
     | Stale TextDocumentVersion v
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -419,12 +419,11 @@ shakeRestart it@IdeState{shakeExtras=ShakeExtras{logger}, ..} systemActs =
               (stopTime,queue) <- duration (cancelShakeSession runner)
               res <- shakeDatabaseProfile shakeProfileDir shakeDb
               let profile = case res of
-                      Just fp ->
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Core/Shake.hs
```diff
@@ -260,15 +250,13 @@ data IdeState = IdeState
...
 -- This is debugging code that generates a series of profiles, if the Boolean is true
-shakeRunDatabaseProfile :: Maybe FilePath -> ShakeDatabase -> [Action a] -> IO ([a], Maybe FilePath)
-shakeRunDatabaseProfile mbProfileDir shakeDb acts = do
-        (time, (res,_)) <- duration $ shakeRunDatabase shakeDb acts
-        proFile <- for mbProfileDir $ \dir -> do
+shakeDatabaseProfile :: Maybe FilePath -> ShakeDatabase -> IO (Maybe FilePath)
+shakeDatabaseProfile mbProfileDir shakeDb =
+        for mbProfileDir $ \dir -> do
                 count <- modifyVar profileCounter $ \x -> let !y = x+1 in return (y,y)
-                let file = "ide-" ++ profileStartTime ++ "-" ++ takeEnd 5 ("0000" ++ show count) ++ "-" ++ showDP 2 time <.> "html"
```
ndmitchell, line 268: I found it quite useful to find the slow database runs quickly by seeing their time in the filename. Typically there are 6 0.00s profiles and one which takes 5s.

src/Development/IDE/Core/Shake.hs
```diff
@@ -429,7 +417,14 @@ shakeRestart it@IdeState{shakeExtras=ShakeExtras{logger}, ..} systemActs =
         shakeSession
         (\runner -> do
               (stopTime,queue) <- duration (cancelShakeSession runner)
-              logDebug logger $ T.pack $ "Restarting build session (aborting the previous one took " ++ showDuration stopTime ++ ")"
+              res <- shakeDatabaseProfile shakeProfileDir shakeDb
+              let profile = case res of
+                      Just fp ->
+                          let link = case filePathToUri' $ toNormalizedFilePath' fp of
+                                          NormalizedUri _ x -> x
+                          in ", profile saved at " <> T.unpack link
```
ndmitchell, line 425: Why are we packing and unpacking the filepath? Why not just dump fp into the message?

src/Development/IDE/Core/Shake.hs
pepeiborra: With the Shake session there is no longer a single run being measured, the profiles measure all what happened for the duration of the session. 

In the absence of that metric, I considered tagging the profiles by the *duration* of the session, but I figured out this would be misleading.

src/Development/IDE/Core/Shake.hs
pepeiborra: This is what the code was doing previously, in an attempt to generate a clickable link. That never worked for me, so I'm happy to remove it.

PR #623, opened 2020-06-10
Fix a bug in getHiFileRule

pepeiborra:
After merging #589 ghcide now errors out occassionally with `internal error - GetHiFile of file of interest`. 

This is great! It uncovers a real bug in `getHiFileRule`, although I'm not sure if the bug would have any user impact.

The problem is that `getHiFileRule` has direct dependencies on `GetHiFile`, which is wrong for the reasons explained in the expanded comment.


ndmitchell:
Could we rename these types as GetModIface and GetModIface_FromDisk so show their relationship more visibly?

PR #624, opened 2020-06-10
Use stale information if it's available to answer requests quickly

mpickering:
With this patch whenever any request made by the user is dealt with by an `IdeAction` which has the implicit guarantee to never block for an unbounded amount of time. 

Stale information is used if it's available to answer all requests. Information should usually be up-to-date because the `FileStore.hs` rule requests up-to-date information whenver a file changes. 

Do not review the changes to the hover/completions etc code too closely because that is all rewritten on another branch by Zubin anyway to use HieDb. 

This patch also lays the way for removing `kick` and more fine grained control over which files are typechecked.

### DIFF
--- a/src/Development/IDE/Core/Service.hs
+++ b/src/Development/IDE/Core/Service.hs
@@ -9,7 +9,7 @@
 --   using the "Shaker" abstraction layer for in-memory use.
 --
 module Development.IDE.Core.Service(
-    getIdeOptions,
+    getIdeOptions, getIdeOptionsIO,
     IdeState, initialise, shutdown,
     runAction,
     writeProfile,
@@ -20,23 +20,20 @@ module Development.IDE.Core.Service(
 
 import Data.Maybe
 import Development.IDE.Types.Options (IdeOptions(..))
-import Control.Monad
 import Development.IDE.Core.Debouncer
 import           Development.IDE.Core.FileStore  (VFSHandle, fileStoreRules)
 import           Development.IDE.Core.FileExists (fileExistsRules)
 import           Development.IDE.Core.OfInterest
-import Development.IDE.Types.Logger
+import Development.IDE.Types.Logger as Logger
 import           Development.Shake
 import qualified Language.Haskell.LSP.Messages as LSP
 import qualified Language.Haskell.LSP.Types as LSP
 import qualified Language.Haskell.LSP.Types.Capabilities as LSP
 
 import           Development.IDE.Core.Shake
+import Control.Monad
 
 
-
-newtype GlobalIdeOptions = GlobalIdeOptions IdeOptions
-instance IsIdeGlobal GlobalIdeOptions
 
 ------------------------------------------------------------
 -- Exposed API
@@ -84,10 +81,6 @@ shutdown = shakeShut
 -- This will return as soon as the result of the action is
 -- available.  There might still be other rules running at this point,
 -- e.g., the ofInterestRule.
-runAction :: IdeState -> Action a -> IO a
-runAction ide action = join $ shakeEnqueue ide action
-
-getIdeOptions :: Action IdeOptions
-getIdeOptions = do
-    GlobalIdeOptions x <- getIdeGlobalAction
-    return x
+runAction :: String -> IdeState -> Action a -> IO a
+runAction herald ide act =
+  join $ shakeEnqueue ide (mkDelayedAction herald Logger.Info act)
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE ConstraintKinds            #-}
 {-# LANGUAGE PatternSynonyms            #-}
+{-# LANGUAGE ExplicitNamespaces         #-}
 
 -- | A Shake implementation of the compiler service.
 --
@@ -30,7 +31,8 @@ module Development.IDE.Core.Shake(
     shakeRestart,
     shakeEnqueue,
     shakeProfile,
-    use, useNoFile, uses,
+    use, useNoFile, uses, useWithStaleFast, useWithStaleFast', delayedAction,
+    FastResult(..),
     use_, useNoFile_, uses_,
     useWithStale, usesWithStale,
     useWithStale_, usesWithStale_,
@@ -38,6 +40,10 @@ module Development.IDE.Core.Shake(
     getDiagnostics, unsafeClearDiagnostics,
     getHiddenDiagnostics,
     IsIdeGlobal, addIdeGlobal, addIdeGlobalExtras, getIdeGlobalState, getIdeGlobalAction,
+    getIdeGlobalExtras,
+    getIdeOptions,
+    getIdeOptionsIO,
+    GlobalIdeOptions(..),
     garbageCollect,
     setPriority,
     sendEvent,
@@ -49,10 +55,12 @@ module Development.IDE.Core.Shake(
     deleteValue,
     OnDiskRule(..),
     WithProgressFunc, WithIndefiniteProgressFunc,
-    ProgressEvent(..)
+    ProgressEvent(..),
+    DelayedAction, mkDelayedAction,
+    IdeAction(..), runIdeAction
     ) where
 
-import           Development.Shake hiding (ShakeValue, doesFileExist)
+import           Development.Shake hiding (ShakeValue, doesFileExist, Info)
 import           Development.Shake.Database
 import           Development.Shake.Classes
 import           Development.Shake.Rule
@@ -65,12 +73,12 @@ import Data.Map.Strict (Map)
 import           Data.List.Extra (partition, takeEnd)
 import qualified Data.Set as Set
 import qualified Data.Text as T
-import Data.Traversable (for)
 import Data.Tuple.Extra
 import Data.Unique
 import Development.IDE.Core.Debouncer
 import Development.IDE.Core.PositionMapping
 import Development.IDE.Types.Logger hiding (Priority)
+import qualified Development.IDE.Types.Logger as Logger
 import Language.Haskell.LSP.Diagnostics
 import qualified Data.SortedList as SL
 import           Development.IDE.Types.Diagnostics
@@ -96,6 +104,9 @@ import           System.IO.Unsafe
 import Language.Haskell.LSP.Types
 import Data.Foldable (traverse_)
 import qualified Control.Monad.STM as STM
+import Control.Monad.IO.Class
+import Control.Monad.Reader
+import Data.Traversable
 
 
 -- information we stash inside the shakeExtra field
@@ -119,12 +130,15 @@ data ShakeExtras = ShakeExtras
     -- ^ How many rules are running for each file
     ,progressUpdate :: ProgressEvent -> IO ()
     -- ^ The generator for unique Lsp identifiers
-    ,restartShakeSession :: [Action ()] -> IO ()
+    ,ideTesting :: IdeTesting
+    -- ^ Whether to enable additional lsp messages used by the test suite for checking invariants
+    ,session :: MVar ShakeSession
     -- ^ Used in the GhcSession rule to forcefully restart the session after adding a new component
     ,withProgress           :: WithProgressFunc
     -- ^ Report progress about some long running operation (on top of the progress shown by 'lspShakeProgress')
     ,withIndefiniteProgress :: WithIndefiniteProgressFunc
     -- ^ Same as 'withProgress', but for processes that do not report the percentage complete
+    ,restartShakeSession :: [DelayedAction ()] -> IO ()
     }
 
 type WithProgressFunc = forall a.
@@ -193,6 +207,19 @@ instance Eq Key where
 instance Hashable Key where
     hashWithSalt salt (Key key) = hashWithSalt salt (typeOf key, key)
 
+newtype GlobalIdeOptions = GlobalIdeOptions IdeOptions
+instance IsIdeGlobal GlobalIdeOptions
+
+getIdeOptions :: Action IdeOptions
+getIdeOptions = do
+    GlobalIdeOptions x <- getIdeGlobalAction
+    return x
+
+getIdeOptionsIO :: ShakeExtras -> IO IdeOptions
+getIdeOptionsIO ide = do
+    GlobalIdeOptions x <- getIdeGlobalExtras ide
+    return x
+
 data Value v
     = Succeeded TextDocumentVersion v
     | Stale TextDocumentVersion v
@@ -210,14 +237,20 @@ currentValue Failed = Nothing
 
 -- | Return the most recent, potentially stale, value and a PositionMapping
 -- for the version of that value.
-lastValue :: NormalizedFilePath -> Value v -> Action (Maybe (v, PositionMapping))
-lastValue file v = do
-    ShakeExtras{positionMapping} <- getShakeExtras
+lastValueIO :: ShakeExtras -> NormalizedFilePath -> Value v -> IO (Maybe (v, PositionMapping))
+lastValueIO ShakeExtras{positionMapping} file v = do
     allMappings <- liftIO $ readVar positionMapping
     pure $ case v of
         Succeeded ver v -> Just (v, mappingForVersion allMappings file ver)
         Stale ver v -> Just (v, mappingForVersion allMappings file ver)
         Failed -> Nothing
+
+-- | Return the most recent, potentially stale, value and a PositionMapping
+-- for the version of that value.
+lastValue :: NormalizedFilePath -> Value v -> Action (Maybe (v, PositionMapping))
+lastValue file v = do
+    s <- getShakeExtras
+    liftIO $ lastValueIO s file v
 
 valueVersion :: Value v -> Maybe TextDocumentVersion
 valueVersion = \case
@@ -246,14 +279,11 @@ type IdeRule k v =
 -- | A live Shake session with the ability to enqueue Actions for running.
 --   Keeps the 'ShakeDatabase' open, so at most one 'ShakeSession' per database.
 data ShakeSession = ShakeSession
-  { cancelShakeSession :: !(IO [Action ()])
+  { cancelShakeSession :: !(IO [DelayedActionInternal])
     -- ^ Closes the Shake session and returns the pending user actions
-  , runInShakeSession  :: !(forall a . Action a -> IO (IO a))
-    -- ^ Enqueue a user action in the Shake session.
+  , runInShakeSession  :: !(forall a . DelayedAction a -> IO (IO a))
+    -- ^ Enqueue an action in the Shake session.
   }
-
-emptyShakeSession :: ShakeSession
-emptyShakeSession = ShakeSession (pure []) (\_ -> error "emptyShakeSession")
 
 -- | A Shake database plus persistent store. Can be thought of as storing
 --   mappings from @(FilePath, k)@ to @RuleResult k@.
@@ -265,6 +295,7 @@ data IdeState = IdeState
     ,shakeProfileDir :: Maybe FilePath
     ,stopProgressReporting :: IO ()
     }
+
 
 
 -- This is debugging code that generates a series of profiles, if the Boolean is true
@@ -340,7 +371,7 @@ shakeOpen :: IO LSP.LspId
           -> Rules ()
           -> IO IdeState
 shakeOpen getLspId eventer withProgress withIndefiniteProgress logger debouncer
-  shakeProfileDir (IdeReportProgress reportProgress) (IdeTesting ideTesting) opts rules = mdo
+  shakeProfileDir (IdeReportProgress reportProgress) ideTesting@(IdeTesting testing) opts rules = mdo
 
     inProgress <- newVar HMap.empty
     (shakeExtras, stopProgressReporting) <- do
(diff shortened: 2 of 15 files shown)


bench/Main.hs
```diff
@@ -218,14 +218,16 @@ runBenchmarks benchmarks = do
     putStrLn $ intercalate ", " $ map (showDuration . snd) results
```
ndmitchell, line 218: How big a deal would it be to send the benchmark changes in via a different PR? Seems totally unrelated to the rest of it, and one for @pepeiborra to review.

bench/Main.hs
```diff
@@ -218,14 +218,16 @@ runBenchmarks benchmarks = do
...
 
 runBench :: HasConfig => Bench -> IO Seconds
-runBench Bench {..} = handleAny (\e -> print e >> return (-1))
+runBench Bench {..} = handleAny (\e -> print e >> return (-2))
```
ndmitchell, line 221: Having -1 as a special case was odd, but just on the edge of tolerable. Now we're having two magic negative seconds I think an enum would be better.

bench/Main.hs
```diff
@@ -236,17 +238,15 @@ runBench Bench {..} = handleAny (\e -> print e >> return (-1))
...
-    liftIO $ threadDelay 50000
-
     return $ if res then t else -1
   where
     cmd =
       unwords $
         [ "ghcide",
           "--lsp",
+          "--test",
+          "--verbose",
```
ndmitchell, line 249: Why pass verbose? Why not take this as an argument to get passed onward?

bench/Main.hs
```diff
@@ -236,17 +238,15 @@ runBench Bench {..} = handleAny (\e -> print e >> return (-1))
...
-    -- sleeep to give ghcide a chance to print the RTS stats
-    liftIO $ threadDelay 50000
-
     return $ if res then t else -1
   where
     cmd =
       unwords $
         [ "ghcide",
           "--lsp",
+          "--test",
```
ndmitchell, line 248: Why is this required?

src/Development/IDE/Core/Service.hs
```diff
@@ -50,7 +47,7 @@ initialise :: LSP.ClientCapabilities
            -> IdeOptions
            -> VFSHandle
            -> IO IdeState
-initialise caps mainRule getLspId toDiags logger debouncer options vfs =
+initialise caps mainRule getLspId toDiags logger debouncer options vfs = do
```
ndmitchell, line 50: What's this change for?

src/Development/IDE/Core/Shake.hs
```diff
@@ -419,45 +457,113 @@ withMVar' var unmasked masked = mask $ \restore -> do
...
+pattern DelayedAction :: String ->  Logger.Priority -> Action a -> DelayedAction a
+pattern DelayedAction a b c = DelayedActionX a b c
+
+mkDelayedAction :: String -> Logger.Priority -> Action a -> DelayedAction a
+mkDelayedAction = DelayedAction
+
+data DelayedActionX a = DelayedActionX { actionName :: String -- Name we show to the user
+                                      -- The queue only contrains entries for
+                                      -- unique key values.
+                                      , actionPriority :: Logger.Priority
```
ndmitchell, line 487: Why are we using log priorities here?

src/Development/IDE/Core/Shake.hs
```diff
@@ -524,6 +634,30 @@ newSession IdeState{shakeExtras=ShakeExtras{..}, ..} systemActs userActs = do
...
+logDelayedAction :: Logger -> DelayedActionInternal -> Action ()
+logDelayedAction l d  = do
+    start <- liftIO $ offsetTime
+    getAction d
+    runTime <- liftIO $ start
+    return ()
+    liftIO $ logPriority l (actionPriority d) $ T.pack $
+        "finish: " ++ (actionName d) ++ " (took " ++ showDuration runTime ++ ")"
+
+freshId :: Var Int -> IO Int
```
ndmitchell, line 657: Why not just reuse Data.Unique?

src/Development/IDE/Core/Shake.hs
```diff
@@ -524,6 +634,30 @@ newSession IdeState{shakeExtras=ShakeExtras{..}, ..} systemActs userActs = do
...
+freshId qcount = do
+    modifyVar qcount (\n -> return (n + 1, n))
```
ndmitchell, line 659: Looks like a potential space leak.

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -395,7 +395,6 @@ getCompletions ideOpts CC { allModNamesAsNS, unqualCompls, qualCompls, importabl
         = filtModNameCompls ++ map (toggleSnippets caps withSnippets
                                       . mkCompl ideOpts . stripAutoGenerated) filtCompls
                             ++ filtKeywordCompls
-
```
ndmitchell, line 398: Why delete one line in this file?

bench/Main.hs
pepeiborra: Agreed. I've got a number of changes for the benchmark suite myself, so a separate PR that can be reviewed and merged quickly would be much appreciated.

pepeiborra:
Benchmarks please!

mpickering:
Benchmarks are vacuous now but anyway.

```
TOTAL hover = 0.01s (10 repetitions)
TOTAL getDefinition = 0.03s (10 repetitions)
TOTAL documentSymbols = 0.65s (100 repetitions)
TOTAL documentSymbols after edit = 0.60s (100 repetitions)
TOTAL completions after edit = 0.16s (10 repetitions)
TOTAL code actions = 0.02s (10 repetitions)
TOTAL code actions after edit = 7.98s (10 repetitions)
```

bench/Main.hs
pepeiborra: I've pushed my benchmarks changes in #625

bench/Main.hs
wz1000: See https://github.com/digital-asset/ghcide/pull/624/files#diff-9f3c72df87c6e40a0ab15d33001d156cR739

mpickering:
I am taking a week break so probably won't get back to this until after that.

wz1000:
>     4. I’m slightly worried about the fact that we switch to a different codepath for testing in `useWithStaleFast'`. I understand why but it still makes it easy for things to creep in that aren’t covered by testing.

I would like to follow the codepath we follow for tests even when not testing, but @mpickering disagrees.

I think it is OK to block on the first request when no results have been computed. Otherwise you get some very weird behavior.

For example, suppose you have an `MaybeT IdeAction` which uses `n` rules, where none of the rules have been computed before. When you run the action the first time, it will fail after trying to get the results of the first rule, and queue up a request to compute the rule. The next time, it will fail when `useE`ing the second rule(if the first rule has finished being computed), and so on. In the best case scenario, the action will have to run at least `n` times before it succeeds.

src/Development/IDE/Core/Shake.hs
```diff
@@ -407,12 +451,72 @@ withMVar' var unmasked masked = mask $ \restore -> do
     putMVar var a'
     pure c
 
+
+-- | Actions with an ID for tracing purposes
+data DelayedActionExtra
+  = DelayedActionExtra
+  { _actionInternalId :: Unique
```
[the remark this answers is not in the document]
wz1000: I believe it is meant to be used for tracing purposes, maybe it will be helpful with the work @mpardalos is going to do.

wz1000:
> I’m less happy to do the same for things not triggered by users even if they are triggered by their editor. This includes things like codelenses, document outline and so on. There are two reasons why I don’t like it here:

One problem with this is that vscode does not send any new requests until it gets a response for these. To the user, this looks the server blocking on hover, go to definition etc.

This means that in practice, a lot of the gains of this PR will be negated. For instance, on modification, vscode will send these requests. Then when the user tries to hover on something, vscode will refuse to send any new requests until the code lenses etc. triggered by the modification return.

I don't really have a problem with this though since I don't use vscode and my editor only sends outline requests when I ask it to.

pepeiborra:
> > I’m less happy to do the same for things not triggered by users even if they are triggered by their editor. This includes things like codelenses, document outline and so on. There are two reasons why I don’t like it here:
> 
> One problem with this is that vscode does not send any new requests until it gets a response for these. To the user, this looks the server blocking on hover, go to definition etc.
> 
> This means that in practice, a lot of the gains of this PR will be negated. For instance, on modification, vscode will send these requests. Then when the user tries to hover on something, vscode will refuse to send any new requests until the code lenses etc. triggered by the modification return.
> 
> I don't really have a problem with this though since I don't use vscode and my editor only sends outline requests when I ask it to.

Are you positive that it is VSCode who doesn't send the requests concurrently? 
I believe that ghcide currently deals with requests single-threadedly, so I don't this is VSCode specific.

wz1000:
> Are you positive that it is VSCode who doesn't send the requests concurrently?
> I believe that ghcide currently deals with requests single-threadedly, so I don't this is VSCode specific.

I think so, since I got frustrated by this once and checked the logs.

pepeiborra:
> > Are you positive that it is VSCode who doesn't send the requests concurrently?
> > I believe that ghcide currently deals with requests single-threadedly, so I don't this is VSCode specific.
> 
> I think so, since I got frustrated by this once and checked the logs.

Do you mean the logs printed by ghcide? There's every chance that those would be single-threaded too

wz1000:
I'm not sure now, they might have been the haskell-lsp logs. I will check again when I'm at my computer.

wz1000:
I have simplified the API somewhat by removing `delay`, `shakeRunUser` and `shakeRunInternal`.

Now we have `shakeEnqueue` and `shakeRestart`, along with `delayedAction`, which is essentially `shakeEnqueue` for the `IdeAction` monad.

wz1000:
I believe this is ready for another look. I think I've addressed most of the comments here.

pepeiborra:
> I'm not sure now, they might have been the haskell-lsp logs. I will check again when I'm at my computer.

Did you have a chance to check whether the lap-client is able to send requests concurrently?

pepeiborra:
Since this branch has performance impacting changes, could you run and share differential benchmarks following the guidelines below

https://github.com/digital-asset/ghcide/blame/master/README.md#L313-L321

alanz:
> Since this branch has performance impacting changes

Given that this PR splits requests into a fast path and a slow path, we must be careful in analysing any performance graphs we produce.

wz1000:
> Did you have a chance to check whether the lsp-client is able to send requests concurrently?

@pepeiborra you are right, my observations were due to the single threaded event loop in ghcide. Now I'm wondering whether we should make that fork. 

Certainly making code actions synchronous now will negate many of the benefit of this PR in vscode.

wz1000:
@pepeiborra I get the following:

```
❯ stack --stack-yaml=stack88.yaml build ghcide:benchHist && stack --stack-yaml=stack88.yaml exec benchHist
# ghcide-bench (for bench-hist/upstream/documentSymbols_after_edit.csv bench-hist/upstream/documentSymbols_after_edit.benchmark-gcStats)
Error when running Shake build system:
  at want, called at bench/hist/Main.hs:92:3 in main:Main
* Depends on: all
  at need, called at bench/hist/Main.hs:113:7 in main:Main
* Depends on: bench-hist/upstream/documentSymbols_after_edit.svg
  at readFileLines, called at bench/hist/Main.hs:380:10 in main:Main
  at readFile', called at src/Development/Shake/Internal/Derived.hs:131:30 in shake-0.18.5-Lg6S0PWFuV673fv8WVPiwb:Development.Shake.Internal.Derived
  at need, called at src/Development/Shake/Internal/Derived.hs:118:15 in shake-0.18.5-Lg6S0PWFuV673fv8WVPiwb:Development.Shake.Internal.Derived
* Depends on: bench-hist/upstream/documentSymbols_after_edit.benchmark-gcStats
  at &%>, called at bench/hist/Main.hs:180:5 in main:Main
* Depends on: bench-hist/upstream/documentSymbols_after_edit.csv bench-hist/upstream/documentSymbols_after_edit.benchmark-gcStats
* Raised the exception:
ghcide-bench: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)
```

pepeiborra:
benchHist needs a `ghcide-bench` binary in the path to work. It would be nice to automate this, but I tried a few things and couldn't get them to work.

You can manually change the path in the `hist.yaml` file

wz1000:
Is `ghcide-bench` the usual `ghcide` executable? If I change `ghcide-bench` to `ghcide` in `hist.yaml`, I get:

```
Command line: ghcide --timeout=3000 --samples=100 --csv=bench-hist/upstream/documentSymbols_after_edit.csv --example-package-version=3.0.0.0 --rts=-I0.5 --ghcide=bench-hist/upstream/ghcide --select 'documentSymbols after edit' -q
Exit code: 1
Stderr:
Invalid option `--timeout=3000'
```

pepeiborra:
`ghcide-bench` is the benchmark executable

alanz:
> Certainly making code actions synchronous now will negate many of the benefits of this PR in vscode.

What do you mean by this? Only about benchmark results? I see plenty of overlapping requests from vscode, that would benefit from using cached data.

pepeiborra:
> @pepeiborra you are right, my observations were due to the single threaded event loop in ghcide. Now I'm wondering whether we should make that fork.
> 
> Certainly making code actions synchronous now will negate many of the benefits of this PR in vscode.

I am confused too, I thought this PR didn't make code action asynchronous. Does it?

It's good to have confirmation that it is ghcide that enforces a sequential lsp request handling. We should consider addressing that as an alternative to using stale data for everything.

wz1000:
Sorry, I meant that not using stale data for code actions while maintaining a synchronous request handling would negate many of the benefits of this PR. This is because vscode frequently sends code action requests, which depend on `TypeCheck`. This means we cannot answer any hover requests while ghcide is computing the response to the code action request. Ideally, we would be able to answer the hover request immediately with stale data.

Asynchronous request processing is not a alternative for this PR since individual requests would still have to wait for data to be recomputed. I think it would, however, be OK to use asynchronous processing for things like code actions to ensure that we don't block for requests like hover.

pepeiborra:
But this PR doesn't change how code actions work, they are still using Shake data, right? This is what confuses me.

wz1000:
Ah, I meant code lens. Code action doesn't use `Typecheck` any way, so it shouldn't be much of a problem.

wz1000:
Benchmark results:

```
name                       | success | samples | startup | setup | experiment | maxResidency
-------------------------- | ------- | ------- | ------- | ----- | ---------- | ------------
hover                      | True    | 10      | 6.98s   | 0.00s | 0.05s      | 57MB        
edit                       | True    | 10      | 6.73s   | 0.00s | 2.90s      | 192MB       
hover after edit           | True    | 10      | 6.77s   | 0.00s | 0.26s      | 57MB        
getDefinition              | True    | 10      | 6.95s   | 0.00s | 0.07s      | 56MB        
documentSymbols            | True    | 100     | 7.09s   | 0.00s | 0.58s      | 56MB        
documentSymbols after edit | True    | 100     | 6.74s   | 0.00s | 1.14s      | 197MB       
completions after edit     | True    | 10      | 6.67s   | 0.00s | 0.56s      | 57MB        
code actions               | True    | 10      | 6.91s   | 0.24s | 0.11s      | 57MB        
code actions after edit    | True    | 10      | 6.51s   | 0.00s | 3.61s      | 219MB       
```

wz1000:
Here is the entire output of the benchmark tool:

```
Running hover benchmark
0.04s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
Running edit benchmark
0.24s
0.24s
0.59s
0.35s
0.22s
0.22s
0.38s
0.22s
0.22s
0.22s
Running hover after edit benchmark
0.24s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
Running getDefinition benchmark
0.05s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
Running documentSymbols benchmark
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.01s
0.01s
0.01s
0.02s
0.00s
0.01s
0.01s
0.01s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.01s
0.02s
0.01s
0.01s
0.01s
0.01s
0.00s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.00s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.00s
0.01s
0.01s
0.01s
0.00s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.00s
0.00s
0.00s
0.01s
0.01s
0.00s
0.01s
0.01s
0.00s
0.00s
0.00s
0.02s
0.01s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.00s
0.01s
0.01s
0.00s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.02s
0.00s
0.00s
0.00s
0.00s
0.00s
Running documentSymbols after edit benchmark
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.02s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.16s
0.01s
0.01s
0.01s
0.01s
0.03s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.02s
0.01s
0.19s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.01s
0.03s
0.01s
0.01s
0.01s
Running completions after edit benchmark
0.26s
0.01s
0.02s
0.02s
0.02s
0.02s
0.02s
0.02s
0.18s
0.01s
Running code actions benchmark
0.08s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
0.00s
Running code actions after edit benchmark
0.30s
0.21s
0.38s
0.19s
0.17s
0.48s
0.59s
0.38s
0.53s
```

As we can see, for the results returning stale information, there is a pause for the initial run. Subsequent requests return instantaneously.

wz1000:
Here are the differential results: http://78.47.113.11/results/

wz1000:
I think the outstanding questions that remain are:

1. Should we block on the initial computation of a request (like we do when `--testing`) or fail until we have a cache? 
2. Should we respond immediately with stale information for requests like `codeLens`

pepeiborra:
1. I don't enjoy having to hover twice, go-to-def twice, etc, so I'd prefer blocking here.
2. It seems trivial to make ghcide handle requests concurrently, see https://github.com/pepeiborra/ghcide/tree/multi-threaded, so I'd prefer to be synchronous for code lenses and all other editor triggered requests

wz1000:
> * I don't enjoy having to hover twice, go-to-def twice, etc, so I'd prefer blocking here.
> 
> * It seems trivial to make ghcide handle requests concurrently, see https://github.com/pepeiborra/ghcide/tree/multi-threaded, so I'd prefer to be synchronous for code lenses and all other editor triggered requests

Addressed both points.

wz1000:
From the differential benchmarks, this result seems to be the most important/striking one:

pepeiborra:
> From the differential benchmarks, this result seems to be the most important/striking one:
> 
> 

Very nice!! That's much better than the optimisation in #670, which is bottle-necked a bit by Shake

wz1000:
See also

src/Development/IDE/Core/Shake.hs
```diff
@@ -620,6 +696,55 @@ usesWithStale_ key files = do
...
+
+askShake :: IdeAction ShakeExtras
+askShake = ask
+
+-- | A (maybe) stale result now, and an up to date one later
+data FastResult a = FastResult { stale :: Maybe (a,PositionMapping), uptoDate :: IO (Maybe a)  }
+
+-- | Lookup value in the database and return with the stale value immediately
+-- Will queue an action to refresh the value.
+-- Never blocks.
```
cocreature, line 716: This is no longer accurate. You block the first time (which is a good idea imho so just update the docs).

src/Development/IDE/Core/Shake.hs
```diff
@@ -407,12 +451,72 @@ withMVar' var unmasked masked = mask $ \restore -> do
...
+
+{-# COMPLETE DelayedAction#-}
+pattern DelayedAction :: String ->  Logger.Priority -> Action a -> DelayedAction a
+pattern DelayedAction a b c = DelayedActionX a b c
+
+mkDelayedAction :: String -> Logger.Priority -> Action a -> DelayedAction a
+mkDelayedAction = DelayedAction
+
+data DelayedActionX a = DelayedActionX
+  { actionName :: String -- ^ Name we show to the user
```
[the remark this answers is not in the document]
cocreature: Can we clarify the doc comment here please?

wz1000:
Fixed!

Issue #617, opened 2020-06-09
Benchmark issues


mpickering:
1. If you have a parent `cabal.project` file then the benchmarks fail to run. Solution, also write a cabal.project file into the benchmarks folder. 
2. The `threadDelay` in `runBench` is fragile and makes the tests fail with broken pipe errors. This is probably a bug in lsp-test but a workaround is to use `-S <file>` and then read the output of the produced file instead. 

I have patches for these locally.

pepeiborra:
1. This is a good find, thanks. Please send a PR!
2. This needs to be fixed in lsp-test I agree. It should use `waitForProcess`, perhaps with a timeout to avoid waiting forever in case of a bad server. The current bench driver does not read the `-S` output so I don't understand how `-S<file>` helps. In my branch, I have heavily changed the bench driver and it reads the `-S` output to add RTS stats to the bench result, see https://github.com/pepeiborra/ghcide/tree/more-benchmark. I'm planning to upstream later this week

mpickering:
The problem is fixed by removing the thread delay which is inserted in order to wait for `-S`. 

The benchmarks also have the issue of sometimes including initialisation time. The check to wait for a work done notification is not robust.

pepeiborra:
Ah yes, removing the timeout does fix the broken pipe errors issue, but then we lose the end output of `-S`, which is quite useful. 

I have not seen the issue with initialisation time before, maybe it is a new bug in ghcide.

pepeiborra:
I can see that the initialisation time is leaking into the first measurement. The robust way to fix this is by including the first measurement in the initialisation time. I'll add a fix to my branch and upstream it

pepeiborra:
I believe this one can be closed now

ndmitchell:
Closing, and I'll leave it to @mpickering to reopen if necessary.

PR #625, opened 2020-06-10
More benchmarks

pepeiborra:
This extends the benchmark suite with 
- two new experiments
- fixes for some of the issues reported in #617 
- Basic memory usage stats (unreliable due to https://github.com/bubba/lsp-test/issues/70)
- much better ouptut
- additional config flags, which I'm using to automate perf analysis elsewhere

### DIFF
--- a/bench/Main.hs
+++ b/bench/Main.hs
@@ -81,6 +81,17 @@ main = do
       bench "hover" 10 $ \doc ->
         isJust <$> getHover doc identifierP,
       ---------------------------------------------------------------------------------------
+      bench "edit" 10 $ \doc -> do
+        let change =
+              TextDocumentContentChangeEvent
+                { _range = Just (Range hygienicP hygienicP),
+                  _rangeLength = Nothing,
+                  _text = " "
+                }
+        changeDoc doc [change]
+        void (skipManyTill anyMessage message :: Session WorkDoneProgressEndNotification)
+        return True,
+      ---------------------------------------------------------------------------------------
       bench "hover after edit" 10 $ \doc -> do
         let change =
               TextDocumentContentChangeEvent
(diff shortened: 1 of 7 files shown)


ghcide.cabal
```diff
@@ -319,7 +319,7 @@ benchmark ghcide-bench
         extra,
         filepath,
         ghcide,
-        lsp-test < 0.12,
+        lsp-test <= 0.11.0.2 && < 0.12,
```
cocreature, line 322: That looks wrong, did you mean `>= 0.11.0.2`?

bench/Main.hs
```diff
@@ -274,14 +372,37 @@ setup = do
...
 
   return $ removeDirectoryRecursive examplesPath
 
--- | Asks the server to shutdown and exit politely
-exitServer :: Session ()
-exitServer = request_ Shutdown (Nothing :: Maybe Value) >> sendNotification Exit ExitParams
-
 --------------------------------------------------------------------------------------------
+
+-- Parse the max residency in RTS -s output
```
cocreature, line 383: Totally fine to keep this for now but once this starts to become more complex we should switch to the machine-readable output of the RTS stats.

ghcide.cabal
pepeiborra: yes, thanks
```suggestion
        lsp-test >= 0.11.0.2 && < 0.12,
```

bench/Main.hs
pepeiborra: Do you mean eventlog? It requires relinking the binary against the eventlog RTS, which would break `cabal bench`.

cocreature: No I mean the machine readable output you get from the standard RTS, see https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/runtime_control.html#rts-options-to-produce-runtime-statistics

pepeiborra: Oh, that's a good point, thanks for the tip! Definitely worth switching to that now or at some point.

PR #626, opened 2020-06-10
Add hie.yaml.stack and use none cradle for test data

jneira:
* I think ghcide is already supporting the none cradle so we will leverage it here
* The stack based hie.yaml could be handy for stack users (not sure if it should be the default one)


ndmitchell:
Awesome. I just tested this and it works great for me straight away, in contrast to the Cabal based cradle, which doesn't work at all.

Issue #603, opened 2020-06-02
Abstract away from LSP where possible


mpickering:
Most of ghcide is already agnostic to LSP but there are a few parts where the abstraction leaks. It would be nice to modify things like `sendEvent` to not directly send an LSP message but emit a message from a custom datatype which could be interpreted into an LSP message or another way. 

The reason for wanting to do this is I want to reimplement the interface of ghci but backed by ghcide. It would also be useful for other tooling which wanted to use ghcide as a library to get off the ground.

ndmitchell:
Do you imagine getting as far away as that there is a ghcide-core library that doesn't depend on LSP, and then a ghcide library that does? Or just the the tentacles shouldn't reach as far?

I imagine someone will want to implement Ghcid on top of Ghcide too, or that your Ghci could incorporate those features as standard.

mpickering:
It would be nicer if you could split the libraries up. I don't think we are that far away from that being possible right now. It's not necessary as a first step though.

alanz:
This sounds like you guys are reinventing the original hie architecture, which had the hie-plugin-api doing this, and intended supporting multiple front ends.  Interesting how the wheel goes around.

PR #627, opened 2020-06-11
Put LspFuncs in ShakeExtras, report progress for cradle setup

bubba:
This sends a progress message when consulting a cradle, and to do so passes about the `LspFuncs` in `ShakeExtras`. This replaces the need for the `eventer`/`sendFunc` and `getLspId`, since they can just be obtained via `LspFuncs`.

In future versions of `haskell-lsp`, the request id stuff will be hidden away inside `LspFuncs` so this will also help future proof it

### DIFF
--- a/src/Development/IDE/Core/FileExists.hs
+++ b/src/Development/IDE/Core/FileExists.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE OverloadedLists      #-}
 {-# LANGUAGE TypeFamilies         #-}
 {-# LANGUAGE UndecidableInstances #-}
@@ -26,6 +27,7 @@ import           Development.IDE.Types.Logger
 import           Development.Shake
 import           Development.Shake.Classes
 import           GHC.Generics
+import           Language.Haskell.LSP.Core
 import           Language.Haskell.LSP.Messages
 import           Language.Haskell.LSP.Types
 import           Language.Haskell.LSP.Types.Capabilities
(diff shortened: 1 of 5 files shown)


src/Development/IDE/Core/FileExists.hs
```diff
@@ -100,30 +104,30 @@ fileExistsRules getLspId ClientCapabilities{_workspace} vfs = do
...
+fileExistsFast :: VFSHandle -> NormalizedFilePath -> Action (Maybe BS.ByteString, ([a], Maybe Bool))
+fileExistsFast vfs file = do
     fileExistsMap <- getFileExistsMapUntracked
     let mbFilesWatched = HashMap.lookup file fileExistsMap
     case mbFilesWatched of
       Just fv -> pure (summarizeExists fv, ([], Just fv))
       Nothing -> do
         exist                   <- liftIO $ getFileExistsVFS vfs file
-        ShakeExtras { eventer } <- getShakeExtras
+        ShakeExtras{lspFuncs=lf} <- getShakeExtras
```
cocreature, line 130: I’d prefer to not abbreviate this. `lspFuncs` isn’t that long.

bubba:
Matthew (not mentioning him here as he's on holiday at the moment) mentioned over IRC that this PR might be moving in the wrong direction as "It is desirable to keep ghcide abstract from LSP as much as possible"
I could rework this so that we pass a separate `withProgress` function into `ShakeExtras` rather than pass the `LspFuncs`, but it feels rather unwieldily. Are there any thoughts on this?

pepeiborra:
> I could rework this so that we pass a separate `withProgress` function into `ShakeExtras` rather than pass the `LspFuncs`, but it feels rather unwieldily. Are there any thoughts on this?

That sounds like a good idea to me.

cocreature:
Adding `withProgress` to `ShakeExtras` sounds reasonable :+1:

bubba:
On second thought I'm still having some reservations about slicing up more stuff from haskell-lsp, since we'll end up having two separate `withProgress` and `withIndefiniteProgress`es passed through etc. I think that at some point it might be wise to integrate haskell-lsp more closely, to avoid having extend `ShakeExtras` every time a new thing is added.

This will have to be refactored at some point in the future anyway, since in @wz1000's next version of haskell-lsp it will no longer going to be possible to manually get the next request ID and construct a `FromServerMessage`, as it will be handled automatically in `sendFunc`:

https://github.com/alanz/haskell-lsp/blob/6aa60a4d72e2309c28e897d3bbe037a4b1fc5022/src/Language/Haskell/LSP/Core.hs#L216-L259

cocreature:
So you’re suggesting to merge this now and think about abstracting over the backend as @mpickering wants for #https://github.com/digital-asset/ghcide/issues/603 at a later point in time? I’m not opposed to merging this particular PR but you’re proposal to integrate even closer with haskell-lsp seems to go into exactly opposite direction of #603.

bubba:
Ah, I didn't see @mpickering's grand plan. Will patch up this PR

bubba:
https://github.com/digital-asset/ghcide/pull/644

PR #629, opened 2020-06-11
Performance analysis over time

pepeiborra:
This adds a new script `benchHist` to automate the running of `ghcide-bench` over multiple versions of the project, and produces `.csv` files and `.svg` graphs with the results. More details in the module header and the README.

By default running `stack exec benchHist` will produce an analysis comparing HEAD with upstream, assuming everything works. The analysis is configurable via a yaml file.

I have uploaded a sample output here: https://github.com/pepeiborra/ghcide/tree/bench-hist-dump/bench-hist 

Some interesting results from the analysis:
- Neil's Hashable PR really hit the leak right in the thunk:
hashable-edit

- Interface files did curb memory usage! 
interface-hover

- But they also made some things twice as slow (and I have a PR coming):
interface-hover-edit

### DIFF
--- a/bench/Hist/Main.hs
+++ b/bench/Hist/Main.hs
@@ -71,9 +71,6 @@ readConfigIO = do
 readConfig :: Action Config
 readConfig = need [config] >> liftIO readConfigIO
 
-build :: String
-build = "bench-hist"
-
 newtype GetSamples = GetSamples () deriving newtype (Binary, Eq, Hashable, NFData, Show)
 
 newtype GetExperiments = GetExperiments () deriving newtype (Binary, Eq, Hashable, NFData, Show)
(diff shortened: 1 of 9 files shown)


pepeiborra:
I forgot to mention that benchHist requires the changes in #625

bench/Hist/Main.hs
```diff
@@ -0,0 +1,472 @@
...
+findPrev :: Text -> [GitCommit] -> Text
+findPrev name (x : y : xx)
+  | humanName y == name = humanName x
+  | otherwise = findPrev name (y : xx)
+findPrev name _ = name
+
+----------------------------------------------------------------------------------------------------
+
+-- | A line in the output of -S
+data Frame = Frame
```
cocreature, line 318: Same comment as on the other PR. This would probably be cleaner using the machine readable output and now we’re getting to the amount of parsing where it pays off.  Given that this is only used during development however, I don’t mind merging it without that change.

pepeiborra: The `--machine-readable` option does not work with the detailed GC statistics provided by `-S`, so the only way around here would be event logs

cocreature: Ah that’s a shame :disappointed:

PR #630, opened 2020-06-11
Cache a ghc session per file of interest

pepeiborra:
I recently discovered that the `.hi` files PR had regressed the performance of `GetSpanInfo` quite noticeably. This is due mostly to a change in the extraction of `spanInfoDoc`s. Prior to the interfaces change, these were mined out of parsed ASTs. `.hi` files don't come with ASTs but instead bundle the haddocks for identifiers. But to be able to use them, we need to:

1. set up a GHC session with all the transitive `.hi` files 
2. rely on [getDocs](https://hackage.haskell.org/package/ghc-8.10.1/docs/GHC.html#v:getDocs) to extract them

Step 1 can be sped up easily by caching a GHC session per module for files of interest, which makes `GetSpanInfo` as speedy as it used to be, as seen in the "hover after edit" experiment.  I was hoping it would also speed up typechecking, but the edit experiment suggests otherwise so I might have missed something.  

Benchmark results (hover becomes much faster, all else stays the same):
```
version              name                         success   samples   startup              setup                 experiment            maxResidency
upstream             hover                        True      100       7.567142767000001    0.0                   0.479372591           218MB
upstream             edit                         True      100       7.953766959          0.0                   25.364602476          217MB
upstream             getDefinition                True      100       8.692988179          0.0                   0.46926602100000003   219MB
upstream             hover after edit             True      100       8.31115918           0.0                   66.021477715          237MB
upstream             completions after edit       True      100       9.410036817          0.0                   33.817070408          222MB
upstream             code actions                 True      100       10.039016802         0.44827246800000004   0.576492419           242MB
upstream             code actions after edit      False     100       7.954712423          0.0                   1.5715171460000001    245MB
upstream             documentSymbols after edit   True      100       9.284543272          0.0                   1.883872325           218MB
module-ghc-session   hover                        True      100       7.223818533          0.0                   0.13876206500000002   228MB
module-ghc-session   edit                         True      100       7.344935921          0.0                   24.669866416          219MB
module-ghc-session   getDefinition                True      100       7.237144228          0.0                   0.119550469           216MB
module-ghc-session   hover after edit             True      100       7.276310433000001    0.0                   29.694541188000002    218MB
module-ghc-session   completions after edit       True      100       7.3100683150000005   0.0                   30.394427346          222MB
module-ghc-session   code actions                 True      100       7.236867787          0.304327993           0.407078469           243MB
module-ghc-session   code actions after edit      False     100       7.397161018          0.0                   0.445487738           242MB
module-ghc-session   documentSymbols after edit   True      100       7.867092457000001    0.0                   1.7937549000000002    216MB
```

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -474,7 +474,7 @@ getModSummaryFromImports fp contents = do
         -- To avoid silent issues where something is not processed because the date
         -- has not changed, we make sure that things blow up if they depend on the date.
                 , ms_hsc_src      = sourceType
-                , ms_hspp_buf     = Nothing
+                , ms_hspp_buf     = Just contents
                 , ms_hspp_file    = fp
                 , ms_hspp_opts    = dflags
                 , ms_iface_date   = Nothing
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -474,6 +474,7 @@ getModSummaryFromImports fp contents = do
         -- To avoid silent issues where something is not processed because the date
         -- has not changed, we make sure that things blow up if they depend on the date.
                 , ms_hsc_src      = sourceType
+                -- The contents are used by the GetModSummary rule
                 , ms_hspp_buf     = Just contents
                 , ms_hspp_file    = fp
                 , ms_hspp_opts    = dflags
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -24,6 +24,7 @@ module Development.IDE.Core.Compile
   , loadInterface
   , loadDepModule
   , loadModuleHome
+  , setupFinderCache
   ) where
 
 import Development.IDE.Core.RuleTypes
@@ -116,24 +117,16 @@ computePackageDeps env pkg = do
 
 typecheckModule :: IdeDefer
                 -> HscEnv
-                -> [(ModSummary, (ModIface, Maybe Linkable))]
                 -> ParsedModule
                 -> IO (IdeResult (HscEnv, TcModuleResult))
-typecheckModule (IdeDefer defer) hsc depsIn pm = do
+typecheckModule (IdeDefer defer) hsc pm = do
     fmap (either (, Nothing) (second Just . sequence) . sequence) $
       runGhcEnv hsc $
       catchSrcErrors "typecheck" $ do
-        -- Currently GetDependencies returns things in topological order so A comes before B if A imports B.
-        -- We need to reverse this as GHC gets very unhappy otherwise and complains about broken interfaces.
-        -- Long-term we might just want to change the order returned by GetDependencies
-        let deps = reverse depsIn
-
-        setupFinderCache (map fst deps)
 
         let modSummary = pm_mod_summary pm
             dflags = ms_hspp_opts modSummary
 
-        mapM_ (uncurry loadDepModule . snd) deps
         modSummary' <- initPlugins modSummary
         (warnings, tcm) <- withWarnings "typecheck" $ \tweak ->
             GHC.typecheckModule $ enableTopLevelWarnings
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -78,6 +78,8 @@ import Control.Concurrent.Async (concurrently)
 import Control.Monad.State
 import System.IO.Error (isDoesNotExistError)
 import Control.Exception.Safe (IOException, catch)
+import FastString (FastString(uniq))
+import qualified HeaderInfo as Hdr
 
 -- | This is useful for rules to convert rules that can only produce errors or
 -- a result into the more general IdeResult type that supports producing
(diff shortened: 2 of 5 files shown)


pepeiborra:
Unfortunately github doesn't allow SVG attachments in PRs, so I have uploaded the benchmarks here: https://github.com/pepeiborra/ghcide/tree/module-ghc-session-benchmarks/bench-hist/module-ghc-session

!(hover after edit)[https://raw.githubusercontent.com/pepeiborra/ghcide/module-ghc-session-benchmarks/bench-hist/module-ghc-session/hover_after_edit.diff.svg]

pepeiborra:
Early cut-off for the `GetModSummary` rule, to bypass work when a module imports & pragmas haven't changed,e.g. GetDependencies, GetDependencyInformation, GetLocatedImports, etc.

This results in further perf improvements for hover after edit *and* for edit.


edit

hover after edit

src/Development/IDE/Core/Compile.hs
```diff
@@ -481,7 +474,7 @@ getModSummaryFromImports fp contents = do
         -- To avoid silent issues where something is not processed because the date
         -- has not changed, we make sure that things blow up if they depend on the date.
                 , ms_hsc_src      = sourceType
-                , ms_hspp_buf     = Nothing
+                , ms_hspp_buf     = Just contents
```
cocreature, line 477: Do we need this for anything now?

src/Development/IDE/Core/Rules.hs
```diff
@@ -623,12 +645,25 @@ getModIfaceFromDiskRule = defineEarlyCutoff $ \GetModIfaceFromDisk f -> do
...
+getModSummaryRule = defineEarlyCutoff $ \GetModSummary f -> do
     dflags <- hsc_dflags . hscEnv <$> use_ GhcSession f
     (_, mFileContent) <- getFileContents f
     modS <- liftIO $ evalWithDynFlags dflags $ runExceptT $
         getModSummaryFromImports (fromNormalizedFilePath f) (textToStringBuffer <$> mFileContent)
-    return $ either (,Nothing) (([], ) . Just) modS
+    case modS of
+        Right ms@ModSummary{..} -> do
+            let fingerPrint =
+                    ( moduleNameString (moduleName ms_mod)
```
cocreature, line 656: What are you basing this on? I’m slightly worried that we are missing some fields of  `ModSummary` that might be important.

src/Development/IDE/Core/Rules.hs
```diff
@@ -667,10 +702,14 @@ getModIfaceRule = define $ \GetModIface f -> do
             case mb_pm of
                 Nothing -> return (diags, Nothing)
                 Just pm -> do
-                    (diags', tmr) <- typeCheckRuleDefinition f pm DoGenerateInterfaceFiles
-                    -- Bang pattern is important to avoid leaking 'tmr'
-                    let !res = extract tmr
-                    return (diags <> diags', res)
+                    (diags, mb_hsc) <- ghcSessionDepsDefinition f
```
cocreature, line 705: Why do we not use the cached version here?

src/Development/IDE/Core/Compile.hs
pepeiborra: Yes. The `ModSummary` rule uses the contents to extract the pragmas which are then recorded in the early cutoff hash

src/Development/IDE/Core/Rules.hs
```diff
@@ -443,30 +445,30 @@ getSpanInfoRule =
...
         packageState <- hscEnv <$> use_ GhcSession file
-        deps <- maybe (TransitiveDependencies [] [] []) fst <$> useWithStale GetDependencies file
-        let tdeps = transitiveModuleDeps deps
 
 -- When possible, rely on the haddocks embedded in our interface files
 -- This creates problems on ghc-lib, see comment on 'getDocumentationTryGhc'
 #if MIN_GHC_API_VERSION(8,6,0) && !defined(GHC_LIB)
         let parsedDeps = []
 #else
+        deps <- maybe (TransitiveDependencies [] [] []) fst <$> useWithStale GetDependencies file
```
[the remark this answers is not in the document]
pepeiborra: I can't remember, will undo to minimize the diff

src/Development/IDE/Core/Rules.hs
pepeiborra: This is based on the fields of `ModSummary` as populated by `getModSummaryFromImports`. Can extract to a helper function and document its purpose and scope if it helps?

src/Development/IDE/Core/Rules.hs
pepeiborra: To avoid creating expensive cache entries for non-files of interest. Will add a comment

src/Development/IDE/Core/Rules.hs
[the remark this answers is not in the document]
pepeiborra: Actually there is a reason for it - `tdeps` is only needed in the legacy code path inside the `#else` branch.

pepeiborra:
Windows 10 build is failing as usual, please retry

PR #634, opened 2020-06-12
Fix non Haddocks diagnostics reported as Haddocks

pepeiborra:


ndmitchell:
This looks wrong. My intention was that you passed the non-Haddock diagnostics first, then the haddock ones second, which is what the code currently does. Why do you think it needs changing?

pepeiborra:
There are two call sites for this function, and one of them is wrong because I am getting real syntax errors reported as warnings

pepeiborra:
It's quite worrying that all the checks have passed if this isn't the right fix

pepeiborra:
Ah, I actually get two diagnostics in the same line, one error and one Haddock warning, so the code is working as expected. 

I wonder if previously I had missed the error, or if it wasn't reported for some reason. Oh well, I'll close the PR since the code is fine and will open an issue if the problem reoccurs.

ndmitchell:
Two syntax errors on the same line? But with different position information?

pepeiborra:
Yes:
```
data Config = Config
  { experiments :: [Unescaped String],
    samples :: Natural,
    versions :: [GitCommit],
    -- | Path to the ghcide-bench binary for the experiments
    ghcideBench :: FilePath
    -- | Output folder ('foo' works, 'foo/bar' does not)
    outputFolder :: String
  }
```

PR #637, opened 2020-06-13
Fix crash when writing to a Barrier more than once

pepeiborra:
This fixes a race condition in the ShakeSession logic for reenqueuing user actions.

### DIFF
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -496,15 +496,17 @@ newSession IdeState{shakeExtras=ShakeExtras{..}, ..} systemActs userActs = do
     --  The session stays open until 'cancelShakeSession' is called
     let runInShakeSession :: forall a . Action a -> IO (IO a)
         runInShakeSession act = do
-              res <- newEmptyMVar -- almost a Barrier, but we attempt to write more than once
+              res <- newBarrier
               let act' = do
                     -- work gets reenqueued when the Shake session is restarted
                     -- it can happen that a work item finished just as it was reenqueud
-                    -- in that case, returning the already produced value is fine
-                    alreadyDone <- liftIO $ tryReadMVar res
-                    maybe (actionCatch @SomeException (Right <$> act) (pure . Left)) return alreadyDone
-              atomically $ writeTQueue actionQueue (act' >>= liftIO . void . tryPutMVar res)
-              return (readMVar res >>= either throwIO return)
+                    -- in that case, skipping the work is fine
+                    alreadyDone <- liftIO $ isJust <$> waitBarrierMaybe res
+                    unless alreadyDone $ do
+                        x <- actionCatch @SomeException (Right <$> act) (pure . Left)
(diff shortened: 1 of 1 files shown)


src/Development/IDE/Core/Shake.hs
```diff
@@ -496,10 +496,15 @@ newSession IdeState{shakeExtras=ShakeExtras{..}, ..} systemActs userActs = do
...
-              atomically $ writeTQueue actionQueue (act' >>= liftIO . signalBarrier res)
-              return (waitBarrier res >>= either throwIO return)
+              res <- newEmptyMVar -- almost a Barrier, but we attempt to write more than once
+              let act' = do
+                    -- work gets reenqueued when the Shake session is restarted
+                    -- it can happen that a work item finished just as it was reenqueud
+                    -- in that case, returning the already produced value is fine
+                    alreadyDone <- liftIO $ tryReadMVar res
+                    maybe (actionCatch @SomeException (Right <$> act) (pure . Left)) return alreadyDone
+              atomically $ writeTQueue actionQueue (act' >>= liftIO . void . tryPutMVar res)
```
ndmitchell, line 506: I find this logic a bit confusing. Why not make act' write to the MVar too? So roughly:

```
let act' = do
     alreadyDone <- liftIO $ tryReadMVar res
     when (isNothing alreadyDone) $ do
          res <- actionCatch ... (Right <$> act) (pure . Left)
          liftIO $ void $ tryPutMVar res
```

As it happens, after that change we only write to the MVar once, so could go back to using a Barrier - but my motivation was actually to just clean up the MVar logic...

ndmitchell: Oh, and really nice find!

pepeiborra: That's much clearer, thanks!

src/Development/IDE/Core/Shake.hs
```diff
@@ -497,8 +497,15 @@ newSession IdeState{shakeExtras=ShakeExtras{..}, ..} systemActs userActs = do
     let runInShakeSession :: forall a . Action a -> IO (IO a)
         runInShakeSession act = do
               res <- newBarrier
-              let act' = actionCatch @SomeException (Right <$> act) (pure . Left)
-              atomically $ writeTQueue actionQueue (act' >>= liftIO . signalBarrier res)
+              let act' = do
+                    -- work gets reenqueued when the Shake session is restarted
+                    -- it can happen that a work item finished just as it was reenqueud
+                    -- in that case, skipping the work is fine
+                    alreadyDone <- liftIO $ isJust <$> waitBarrierMaybe res
```
cocreature, line 504: Just confirming that I get this right: We don’t need to worry about this being racy and the barrier getting updated between our check here and the `unless` since there is only one writer?

pepeiborra: Only Shake can write, correct. So if this code is running the previous Shake session has already been restarted and there is no one who can fill in the Barrier in between lines 503 and 504

PR #638, opened 2020-06-13
 Remove `Strict` from the language extensions used for code actions

tek:
This code action uses an unreliable heuristic, i.e. substring matching
for all existing language extensions.
The "quickfix" LSP action chooses an arbitrary code action from the
list, presumably the first in the list.

As an example, the redundant import

`import Data.Map.Strict`

will trigger the suggestion to add the language extension "Strict".
Moving this suggestion to the end of the list reduces the likelihood
for false positives.

----

Since selection of the quickfix action happens outside of ghcide, there's probably a better solution.
Any suggestions? A better workaround could be to check for the language extension's name's presence in the offending code.

I added a test but it fails and I don't see the definitions of `getCodeActions` and `executeCodeAction`. Some advice would be appreciated!

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -266,7 +266,11 @@ suggestAddExtension Diagnostic{_range=_range,..}
 
 -- | All the GHC extensions
 ghcExtensions :: Map.HashMap T.Text Extension
-ghcExtensions = Map.fromList . map ( ( T.pack . flagSpecName ) &&& flagSpecFlag ) $ xFlags
+ghcExtensions = Map.fromList . filter notStrictFlag . map ( ( T.pack . flagSpecName ) &&& flagSpecFlag ) $ xFlags
+  where
+    -- Strict often causes false positives, as in Data.Map.Strict imports.
+    -- See discussion at https://github.com/digital-asset/ghcide/pull/638
+    notStrictFlag (name, _) = name /= "Strict"
 
 suggestModuleTypo :: Diagnostic -> [(T.Text, [TextEdit])]
 suggestModuleTypo Diagnostic{_range=_range,..}
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -818,6 +818,28 @@ removeImportTests = testGroup "remove import actions"
             , "main = B"
             ]
       liftIO $ expectedContentAfterAction @=? contentAfterAction
+  , testSession "import containing the identifier Strict" $ do
+      let contentA = T.unlines
+            [ "module Strict where"
+            ]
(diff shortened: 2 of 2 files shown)


pepeiborra:
I agree that the suggest extension logic is a bit flaky, it's the way it is until GHC gives us structured error messages (there are already 2 GHC proposals under way). 

Lowering the priority will help the 20% of  cases where the suggestion is wrong, but unhelp the other 80% of cases where the suggestion is right. 

Based on my experience that `Strict` is the extension that triggers this most commonly (one of the few extensions that has a single word name?) I would prefer a fix that removes `Strict` from the list of `ghcExtensions`.

Those functions come from the [lsp-test package](http://hackage.haskell.org/package/lsp-test-0.11.0.2/docs/Language-Haskell-LSP-Test.html#v:getCodeActions)

tek:
> I agree that the suggest extension logic is a bit flaky, it's the way it is until GHC gives us structured error messages (there are already 2 GHC proposals under way).

awesome!

> 
> Lowering the priority will help the 20% of cases where the suggestion is wrong, but unhelp the other 80% of cases where the suggestion is right.
> 
> Based on my experience that `Strict` is the extension that triggers this most commonly (one of the few extensions that has a single word name?) I would prefer a fix that removes `Strict` from the list of `ghcExtensions`.

Sounds good to me!

> 
> Those functions come from the [lsp-test package](http://hackage.haskell.org/package/lsp-test-0.11.0.2/docs/Language-Haskell-LSP-Test.html#v:getCodeActions)

thanks!

tek:
@pepeiborra you fine with it like this?

tek:
awesome! :rocket:

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -266,7 +266,9 @@ suggestAddExtension Diagnostic{_range=_range,..}
 
 -- | All the GHC extensions
 ghcExtensions :: Map.HashMap T.Text Extension
-ghcExtensions = Map.fromList . map ( ( T.pack . flagSpecName ) &&& flagSpecFlag ) $ xFlags
+ghcExtensions = Map.fromList . filter notStrictFlag . map ( ( T.pack . flagSpecName ) &&& flagSpecFlag ) $ xFlags
+  where
+    notStrictFlag (name, _) = name /= "Strict"
```
ndmitchell, line 271: I'd add a link to the issue you initially raised here. It's one odd and unexpected exception, so having the backstory is always useful.

tek: @ndmitchell done!
there's no issue though, only this PR.

ndmitchell: Thanks - this PR is fine.

PR #640, opened 2020-06-14
Write a cabal.project file in the benchmark example

pepeiborra:
Extracted from #624 as this has started biting me too after upgrading cabal-install in my development environment

### DIFF
--- a/src/Development/IDE/GHC/Compat.hs
+++ b/src/Development/IDE/GHC/Compat.hs
@@ -16,7 +16,7 @@ module Development.IDE.GHC.Compat(
     writeHieFile,
     readHieFile,
     supportsHieFiles,
-    setDefaultHieDir,
+    setHieDir,
     dontWriteHieFiles,
 #if !MIN_GHC_API_VERSION(8,8,0)
     ml_hie_file,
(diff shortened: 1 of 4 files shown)


src/Development/IDE/GHC/Compat.hs
```diff
@@ -191,10 +190,10 @@ pattern IEThingAll a <-
     GHC.IEThingAll a
 #endif
 
-setDefaultHieDir :: FilePath -> DynFlags -> DynFlags
-setDefaultHieDir _f d =
+setHieDir :: FilePath -> DynFlags -> DynFlags
```
cocreature, line 193: This seems unrelated. Is the goal here to separate it from the directory cabal uses? That seems sensible. Do you know why it was like this in the first place?

pepeiborra: It's unrelated but in the same area (benchmarks). I added this fix to this PR to avoid starting a new one.

My thinking when I decided not to override the user setting was that `.hie` files are independent of optimisation settings and therefore completely reusable across ghcide and cabal. This is still true, but as the commit message says, it turns out that ghcide makes assumptions about the lifetime of `.hi` and `.hie` files being the same, which is only true when they are both in a file system location under the control of ghcide.

cocreature: Thanks for the explanation!

pepeiborra: As mentioned in the commit message, this breaks the getDefinition experiment because the `.hie` files live in the Cabal folder and get wiped across bench runs, whereas the `hi` files live in the cache folder and survive across

PR #643, opened 2020-06-15
Finer dependencies for GhcSessionFun

### LINKED ISSUE #619: GhcSessionFun should not be alwaysRerun
Ideally, `GhcSessionFun` would only be rerun if a new component has been added. I think this can be achieved by moving the `alwaysRerun` one level up to `GhcSessionIO`.

https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Core/Rules.hs#L565-L577

pepeiborra:
Now `GhcSession` tracks the component dependencies, and `GhcSessionIO` (no file) tracks the set of components. Since `GhcSession` has a dependency on `GhcSessionIO`, this works out as intended: if the set of components changes, all the Ghc sessions become invalidated. Otherwise, a Ghc session is invalidated only when the component dependencies (e.g. the Cabal file) change.

Another 10% perf improvement.

Fixes #619 

Benchmarks where HEAD is this patch, upstream is the previous master:
```
version    name                         success   samples   startup              setup                experiment           maxResidency
upstream   hover                        True      100       10.008805828         0.0                  1.0496357980000002   138MB
upstream   edit                         True      100       10.342245301         0.0                  40.515747643000005   137MB
upstream   getDefinition                True      100       20.589914577000002   0.0                  0.537914714          324MB
upstream   hover after edit             True      100       9.944694886          0.0                  89.364840677         156MB
upstream   completions after edit       True      100       9.465573668000001    0.0                  50.609864630000004   148MB
upstream   code actions                 True      100       9.376972059          0.6628133180000001   0.96582895           166MB
upstream   code actions after edit      True      100       10.779993297         0.0                  41.802328869         171MB
upstream   documentSymbols after edit   True      100       9.692719622          0.0                  3.1590869070000003   133MB
HEAD       hover                        True      100       10.359037769         0.0                  0.8585313830000001   137MB
HEAD       edit                         True      100       9.726820669          0.0                  31.940400373000003   137MB
HEAD       getDefinition                True      100       21.827025083000002   0.0                  0.648860783          318MB
HEAD       hover after edit             True      100       9.524636392000001    0.0                  80.23256079800001    156MB
HEAD       completions after edit       True      100       10.814269279000001   0.0                  43.996691731000006   143MB
HEAD       code actions                 True      100       9.39776919           0.595851581          0.9932212890000001   166MB
HEAD       code actions after edit      True      100       9.59566279           0.0                  38.154787442         171MB
HEAD       documentSymbols after edit   True      100       9.920586814          0.0                  3.0366749910000004   135MB
```
Graphs here: https://github.com/pepeiborra/ghcide/tree/GhcSession-alwaysRerun-benchmarks/bench-hist

Edit experiment

Hover after edit experiment

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -18,6 +18,7 @@ import Control.Concurrent.Extra
 import Control.Exception.Safe
 import Control.Monad.Extra
 import Control.Monad.IO.Class
+import Data.Bifunctor (Bifunctor(second))
 import Data.Default
 import Data.Either
 import Data.Foldable (for_)
@@ -122,7 +123,7 @@ main = do
         runLanguageServer options (pluginHandler plugins) onInitialConfiguration onConfigurationChange $ \getLspId event vfs caps wProg wIndefProg -> do
             t <- t
             hPutStrLn stderr $ "Started LSP server in " ++ showDuration t
-            sessionLoader <- loadSessionShake dir
+            sessionLoader <- loadSession dir
             let options = (defaultIdeOptions sessionLoader)
                     { optReportProgress = clientSupportsProgress caps
                     , optShakeProfiling = argsShakeProfiling
@@ -155,7 +156,7 @@ main = do
         vfs <- makeVFSHandle
         debouncer <- newAsyncDebouncer
         let dummyWithProg _ _ f = f (const (pure ()))
-        sessionLoader <- loadSessionShake dir
+        sessionLoader <- loadSession dir
         ide <- initialise def mainRule (pure $ IdInt 0) (showEvent lock) dummyWithProg (const (const id)) (logger minBound) debouncer (defaultIdeOptions sessionLoader)  vfs
 
         putStrLn "\nStep 4/4: Type checking the files"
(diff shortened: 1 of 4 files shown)


src/Development/IDE/Types/Options.hs
```diff
@@ -23,12 +24,22 @@ import           GhcPlugins                     as GHC hiding (fst3, (<>))
 import qualified Language.Haskell.LSP.Types.Capabilities as LSP
 import qualified Data.Text as T
 import Development.IDE.Types.Diagnostics
+import Control.DeepSeq (NFData(..))
+
+data IdeGhcSession = IdeGhcSession
+  { loadSessionFun :: FilePath -> IO (IdeResult HscEnvEq, [FilePath])
+  -- ^ Returns the Ghc session and the cradle dependencies
+  , sessionVersion :: !Int
```
cocreature, line 32: A comment on the semantics here would be great. It looks like you only use this as a shake key so the only requirement is that you never reuse a version but it’s probably a good choice to require it to be strictly increasing (but not necessarily in steps of 1).

exe/Main.hs
```diff
@@ -372,10 +381,11 @@ loadSession optTesting ShakeExtras{logger, eventer, restartShakeSession} dir = d
                  let res = (map (renderCradleError ncfp) err, Nothing)
                  modifyVar_ fileToFlags $ \var -> do
                    pure $ Map.insertWith HM.union hieYaml (HM.singleton ncfp (res, dep_info)) var
-                 return res
+                 return (res,[])
 
       -- This caches the mapping from hie.yaml + Mod.hs -> [String]
-      let sessionOpts :: (Maybe FilePath, FilePath) -> IO (IdeResult HscEnvEq)
+      -- Returns the Ghc session and the cradle dependencies
+      let sessionOpts :: (Maybe FilePath, FilePath) -> IO (IdeResult HscEnvEq, [FilePath])
```
cocreature, line 388: It looks like there is some duplication here now. We track the modification times outside of Shake via `getDependencyInfo` but we also call `GetModificationTime` on those files. Can we merge the two somehow or remove one of them?

pepeiborra: That's true, I just didn't have a good enough understanding of the multi cradle code to change it. 

I think the Shake tracking should stay and the cradle code changed to assume that the deps always change (since it wouldn't be called otherwise). But that seems a risky change.

pepeiborra: I'll give this a quick go and see where it leads, but maybe in a a follow-up PR?

pepeiborra:
Rebased

exe/Main.hs
cocreature: Yeah let’s leave it out for now. I’d still like a comment on `sessionVersion`, then this should be good to merge.

src/Development/IDE/Types/Options.hs
pepeiborra: Added

PR #644, opened 2020-06-16
Report progress when setting up cradle

bubba:
To do this we pass in the withProgress and withIndefiniteProgress functions from LspFuncs into ShakeExtras

Supersedes #627


bubba:
Test failure looks kind of flaky

cocreature:
Restarted, 8.10 is pretty broken on Windows. Should probably just add some retries.

PR #649, opened 2020-06-19
Send WorkDoneProgressEnd only when work is done

pepeiborra:
Recent changes to startup and progress reporting have resulted in multiple `NotWorkDoneProgressEnd` notifications being sent at startup, which makes automation (benchmarks, tests) challenging and seems meaningless from a user pov.

The goal of this PR is to eliminate those `NotWorkDoneProgressEnd` notifications sent when a `kick` is aborted before completion.

But what should the lsp server send to notify that the work got cancelled? I cannot find anything in the lsp specification for that scenario, so I've opted for not sending anything. 

Thoughts?


ndmitchell:
I've no idea what LSP would like here. CC @alanz who might.

alanz:
The [work done progress](https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#workDoneProgressBegin) is managed via tokens.  For each token there should be a `begin`, a series of zero or more `update`s, and then a "deallocation" via an `end`.

I have not looked into the current incarnation of the progress stuff in ghcide recently, but if I recall at one stage there was a "work in progress" counter synthesised that kept track of how many things are still happening.

If that is the case, then doing something like sending an `end` when the counter gets to zero, an `update` if it is progress time and there are still tasks and a token active, and a `begin` otherwise.

As I understand it we do not have enough insight into the work being done via the shake graph, and it is too dynamic, to be able to do more than that.  This should result in a "busy" indicator while there is actually work being done.

pepeiborra:
The specific question here is what to do when the activity tracked by a progress token is cancelled. The LSP specification doesn't say anything in this regard for server-initiated progress, and it doesn't seem to mandate that every `begin` is followed by an end. 

There is a third alternative, which is to continue reporting on the same token when a `kick` is restarted. The problem with this option is that progress will no longer be monotonically increasing.

Summarizing the three options:

|Model|kick|kick|done|
|----|---|---|---|
|HEAD|begin(1)|end(1),begin(2)|end(2)|
|this PR|begin(1)|begin(2)|end(2)|
|3rd alt.|begin(1)||end(1)|

The 3rd alternative seems saner.
Does that make sense?

alanz:
I think monotonic progress is not feasible for ghcide, the graph is too dynamic.  Also, given the speed of processing, I am not sure that showing specific messages (which is what unique tokens gives) adds any value.

According to https://github.com/microsoft/language-server-protocol/issues/923#issuecomment-588146124, the begin is used to allocate resources (in vscode at least), so it seems logical that and end is needed to free them.

alanz:
FYI here is a screenshot of the very real resources allocated in vscode for progress

Screenshot 2020-06-19 at 14 31 34

These are long running processes doing indexing for erlang.

pepeiborra:
The resources shared by @alanz suggest that the third alternative is the better one. I'll keep this PR open for further discussion but move it to draft to prevent merging before it is updated.

pepeiborra:
Updated with my implementation for the third alternative

src/Development/IDE/Core/Rules.hs
```diff
@@ -675,6 +675,7 @@ getModSummaryRule = defineEarlyCutoff $ \GetModSummary f -> do
 
 getModIfaceRule :: Rules ()
 getModIfaceRule = define $ \GetModIface f -> do
+#if MIN_GHC_API_VERSION(8,6,0) && !defined(GHC_LIB)
```
cocreature, line 678: What’s up with this? What changed here?

src/Development/IDE/Core/Shake.hs
```diff
@@ -346,6 +348,11 @@ shakeOpen getLspId eventer withProgress withIndefiniteProgress logger debouncer
         publishedDiagnostics <- newVar mempty
         positionMapping <- newVar HMap.empty
         let restartShakeSession = shakeRestart ideState
+        progressQueue <- newTVarIO KickCompleted
```
cocreature, line 351: It’s a bit surprising that this is calle `Queue` but then it’s just a `TVar` instead of a queue.

src/Development/IDE/Core/Shake.hs
```diff
@@ -346,6 +348,11 @@ shakeOpen getLspId eventer withProgress withIndefiniteProgress logger debouncer
...
+        let progressUpdate = atomically . writeTVar progressQueue
+        when reportProgress $
+            void $ async $ progressThread progressQueue inProgress
```
cocreature, line 354: Afaict this thread will keep running even if you call `shakeShut`. While I think this probably doesn’t matter much for ghcide since we usually only call this much it matters when you use ghcide as a library and might run multiple sessions within a single process.

src/Development/IDE/Core/Rules.hs
pepeiborra: The benchmark tests revealed that `getDefinition` was broken in GHC 8.4. This fixes it. I can add more detail if needed.

src/Development/IDE/Core/Shake.hs
pepeiborra: good point - it was originally a queue but then I realized only the most recent value matters. Will change.

src/Development/IDE/Core/Shake.hs
pepeiborra: good point, will fix

src/Development/IDE/Core/Rules.hs
cocreature: No need to add anything more, I was just curious how this was related to this PR. Thanks!

test/exe/Main.hs
```diff
@@ -2241,6 +2244,20 @@ nonLspCommandLine = testGroup "ghcide command line"
...
+benchmarkTests =
+    let ?config = Bench.defConfig
+            { Bench.verbosity = Bench.Quiet
+            , Bench.repetitions = Just 3
+            , Bench.buildTool = Bench.Stack
+            } in
+    withResource Bench.setup id $ \_ -> testGroup "benchmark experiments"
+    [ testCase (Bench.name e) $ do
+        res <- Bench.runBench runInDir e
+        assertBool "did not successfully complete 5 repetitions" $ Bench.success res
```
cocreature, line 2257: This has just failed on master (no additional output). Do you expect this to be flaky?

wz1000: For me it was failing locally because it ran stack, which then tried to download `ghc-8.6.5`, which led to a timeout.

pepeiborra: This is flaky, yes. I'll have a fix in a few minutes

pepeiborra: It defaults to `stack` because that's what CI uses, but it can be made to work with Cabal as well. I just haven't figured out how to autodetect which build tool is being used

PR #651, opened 2020-06-19
stack810.yaml: bump

domenkozar:
Needs `diagrams-svg` release.

### DIFF
--- a/stack810.yaml
+++ b/stack810.yaml
@@ -1,5 +1,4 @@
-resolver: nightly-2020-02-13
-compiler: ghc-8.10.1
+resolver: nightly-2020-06-19
 allow-newer: true
 packages:
 - .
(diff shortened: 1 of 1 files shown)


stack810.yaml
```diff
@@ -1,6 +1,4 @@
-resolver: nightly-2020-02-13
-compiler: ghc-8.10.1
-allow-newer: true
```
cocreature, line 3: Can we just keep the `allow-newer` for now? Clearly `diagrams-svg` needs it.

pepeiborra:
It might be a good idea to make `benchHist` non buildable by default, or turn it into a benchmark. I believe @alanz did this for the hls fork of ghcide

domenkozar:
It builds now but benchmarking has a lot of these:

```
Warning: Failed to decode module interface: /home/ielectric/dev/ghcide/.stack-work/dist/x86_64-linux-nix/Cabal-3.2.0.0/build/ghcide/ghcide-tmp/Utils.hi Decoding failure: Invalid magic: e49ceb0f
```

cocreature:
> It builds now but benchmarking has a lot of these:

That’s a stack bug https://github.com/commercialhaskell/stack/issues/5134

PR #653, opened 2020-06-20
Code action: add constraint

### LINKED ISSUE #345: Code action: add additional constraint
I just got:

```
C:\Neil\model\Main.hs:20:19: error:
    • No instance for (Hashable a) arising from a use of ‘hash’
      Possible fix:
        add (Hashable a) to the context of
          the type signature for:
            apply :: forall a. Int -> RW a -> Map.Map a Int -> Map.Map a Int
    • In the expression:
        hash (name, [(v, mp Map.! v) | v <- Set.toAscList $ Set.union r w])
      In an equation for ‘input’:
          input
            = hash
 ```

That could probably have been a code action.

### PR DESCRIPTION
**Pull request by DenisFrezzato**
This PR adds code actions to generate suggestions for missing constraints, both for `instance`s declaration and type signatures. I've handled cases where no, one or more other constraints are already declared.

I've also introduced operators like `>>>`, `&`, `<&>` in order to compose from left to right (like the code flows from left to right) and to have code more easier to read and to reason about. I'm mentioning this because I didn't see them used in the code (at least not in the files I changed), so if for some reason they are not welcome, I'll change them.

This closes https://github.com/digital-asset/ghcide/issues/345.

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -472,8 +472,8 @@ suggestInstanceConstraint contents Diagnostic {..} missingConstraint
   | otherwise = []
     where
       findExistingConstraints :: T.Text -> T.Text
-      findExistingConstraints =
-        T.lines >>> flip (!!) 1 >>> T.strip >>> T.replace "from the context: " ""
+      findExistingConstraints t =
+        T.replace "from the context: " "" . T.strip $ T.lines t !! 1
 
       readPositionNumber :: T.Text -> Int
       readPositionNumber = T.unpack >>> read >>> pred
@@ -529,7 +529,7 @@ suggestFunctionConstraint contents Diagnostic{..} missingConstraint
       findExistingConstraints :: T.Text -> Maybe T.Text
       findExistingConstraints message =
         if message =~ ("from the context:" :: String)
-           then matchRegex message "\\. ([^=]+)" <&> head <&> T.strip
+           then fmap (T.strip . head) $ matchRegex message "\\. ([^=]+)"
            else Nothing
 
       buildNewConstraints :: T.Text -> Maybe T.Text -> T.Text
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -56,6 +56,7 @@ import Outputable (ppr, showSDocUnsafe)
 import DynFlags (xFlags, FlagSpec(..))
 import GHC.LanguageExtensions.Type (Extension)
 import System.Time.Extra (showDuration, duration)
+import Data.Function
 
 plugin :: Plugin c
 plugin = codeActionPluginWithRules rules codeAction <> Plugin mempty setHandlersCodeLens
@@ -155,6 +156,7 @@ suggestAction dflags packageExports ideOptions parsedModule text diag = concat
     , suggestModuleTypo diag
     , suggestReplaceIdentifier text diag
     , suggestSignature True diag
+    , suggestInstanceConstraint text diag
     ] ++ concat
     [  suggestNewDefinition ideOptions pm text diag
     ++ suggestRemoveRedundantImport pm text diag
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Development/IDE/Plugin/CodeAction.hs [hunk 1]
```diff
@@ -375,6 +380,132 @@ suggestSignature isQuickFix Diagnostic{_range=_range@Range{..},..}
...
+
+-- | Suggests a constraint for an instance declaration for which a constraint is missing.
+suggestInstanceConstraint :: Maybe T.Text -> Diagnostic -> [(T.Text, [TextEdit])]
+suggestInstanceConstraint contents Diagnostic {..}
+-- Suggests a constraint for an instance declaration with no existing constraints.
+-- • No instance for (Eq a) arising from a use of ‘==’
+--   Possible fix: add (Eq a) to the context of the instance declaration
+-- • In the expression: x == y
+--   In an equation for ‘==’: (Wrap x) == (Wrap y) = x == y
+--   In the instance declaration for ‘Eq (Wrap a)’
```
fendor, line 403: Drive-by-comment: I think it would be nicer if this awesome comment is part of the haddock documentation, since hover will show it. Just my personal taste, though.

DenisFrezzato: I'm not sure what you mean here. I've added a documentation comment at line 395. The one you're referring to just explains what case the branch handles and an example of the missing constraint error from GHC (like it has been done for other code actions). Can you show me an example of what you mean?

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -375,6 +380,132 @@ suggestSignature isQuickFix Diagnostic{_range=_range@Range{..},..}
...
+--     add (Eq b) to the context of
+--       the type signature for:
+--         eq :: forall a b. Eq a => Pair a b -> Pair a b -> Bool
+-- • In the second argument of ‘(&&)’, namely ‘y == y'’
+--   In the expression: x == x' && y == y'
+--   In an equation for ‘eq’:
+--       eq (Pair x y) (Pair x' y') = x == x' && y == y'
+  | Just c <- contents
+  , True <- _message =~ ("the type signature for:" :: String)
+  , Just constraint <- findMissingConstraint _message
```
pepeiborra, line 480: `findMissingConstraint _message` can get evaluated up to 3 times, which seems a bit wasteful. It would be good to reorganize things so that it only needs to be called once.

DenisFrezzato: Good point! I'll try to address this.

src/Development/IDE/Plugin/CodeAction.hs
fendor, line 403:
```suggestion
-- | Suggests a constraint for an instance declaration for which a constraint is missing.
--
-- Suggests a constraint for an instance declaration with no existing constraints.
-- • No instance for (Eq a) arising from a use of ‘==’
--   Possible fix: add (Eq a) to the context of the instance declaration
-- • In the expression: x == y
--   In an equation for ‘==’: (Wrap x) == (Wrap y) = x == y
--   In the instance declaration for ‘Eq (Wrap a)’
suggestInstanceConstraint :: Maybe T.Text -> Diagnostic -> [(T.Text, [TextEdit])]
suggestInstanceConstraint contents Diagnostic {..}
```

I mean this. This way, on hover on this function, you can see all comments.
But if you think that is a bad idea, that is fine, too.

src/Development/IDE/Plugin/CodeAction.hs
DenisFrezzato: I've created an entrypoint for missing constraint suggestion in order to perform `findMissingConstraint _message` only once and then routing to the relevant implementation.

src/Development/IDE/Plugin/CodeAction.hs
DenisFrezzato: I see. The comment you moved refers only to the branch of the pattern matching, so it's relevant to an implementation detail. Besides, the first line would be kind of redundant, wouldn't it? There is also another comment similar to this one for the branch of the pattern match right below. If we moving both of them out of that context they would loose their meaning, since they strictly refer to the branches. What do you think?

fendor: > the first line would be kind of redundant, wouldn't it?

Yeah, but I dont think this is really bad.

> There is also another comment similar to this one 

Yeah, I would move that, too.

> What do you think?

Your concern seems valid to me. *Personally*, I would like it to have it available as haddock comments. *But*, if you think it is too much of an implementation detail to actually show in the documentation, you are probably right. So, whatever feels right to you!

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -375,6 +379,140 @@ suggestSignature isQuickFix Diagnostic{_range=_range@Range{..},..}
...
+        endOfConstraint = Position typeSignatureLine $
+          typeSignatureFirstChar + maybe 0 T.length mExistingConstraints
+        range = Range startOfConstraint endOfConstraint
+     in [(actionTitle missingConstraint typeSignatureName, [TextEdit range newConstraint])]
+  | otherwise = []
+    where
+      findExistingConstraints :: T.Text -> Maybe T.Text
+      findExistingConstraints message =
+        if message =~ ("from the context:" :: String)
+           then matchRegex message "\\. ([^=]+)" <&> head <&> T.strip
```
cocreature, line 503:
```suggestion
           then fmap (T.strip . head) $ matchRegex message "\\. ([^=]+)"
```
Not a huge fan of reverse function application especially if it’s mixed with regular function application in the same expression.

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -375,6 +379,140 @@ suggestSignature isQuickFix Diagnostic{_range=_range@Range{..},..}
...
+        startOfConstraint = Position instanceLine constraintFirstChar
+        endOfConstraint = Position instanceLine $
+          constraintFirstChar + T.length existingConstraints
+        range = Range startOfConstraint endOfConstraint
+     in [(actionTitle missingConstraint, [TextEdit range newConstraints])]
+  | otherwise = []
+    where
+      findExistingConstraints :: T.Text -> T.Text
+      findExistingConstraints =
+        T.lines >>> flip (!!) 1 >>> T.strip >>> T.replace "from the context: " ""
```
cocreature, line 447:
```suggestion
      findExistingConstraints :: T.Text -> T.Text
      findExistingConstraints t =
        T.replace "from the context: " "" . T.strip $ T.lines t !! 1
```
No need to code golf this into pointfree style.

cocreature:
This PR needs rebasing on master since it has conflicted with #657. If you run into issues resolving the conflicts, I’m happy to help.

PR #657, opened 2020-06-21
Delete unused top level binding code action

### LINKED ISSUE #54: code action for deleting unused definitions
This is probably a good first issue, unless working out exactly to delete proves hard.

If a symbol isn't used and isn't exported, GHC emits this warning.
```
C:\msys64\home\ganesh\temp\x\Foo.hs:3:1: warning:
    Defined but not used: ‘foo’
```
It'd be great to have a code action to delete it.

Complicating factors:
 - the text range from the warning only points to the name itself, not the entire definition, so we need to find the full definition somehow. I guess GHC's parse tree should be available?
 - you get the same warning for local definitions and unused parameters. Local definitions can also just be deleted but it may be harder to work out exactly what to delete. Unused parameters should be either turned into or prefixed with '_'.

serhiip:
This is my first contribution. My goal is contribute to haskell-language-server / ghcide constantly. This adds a code action that deletes unused top level binding.

Please, let me know how this should be improved.

This closes #54 (I think)

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -191,6 +191,7 @@ suggestDeleteTopBinding ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecls}
                         . sortLocated
                         $ hsmodDecls
           sameName = filter (matchesBindingName (T.unpack name) . snd) allTopLevel
+    , not (null sameName)
             = [("Delete ‘" <> name <> "’", flip TextEdit "" . toNextBinding allTopLevel . fst <$> sameName )]
     | otherwise = []
     where
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -48,6 +48,7 @@ import Data.List.Extra
 import qualified Data.Text as T
 import Data.Tuple.Extra ((&&&))
 import HscTypes
+import SrcLoc
 import Parser
 import Text.Regex.TDFA ((=~), (=~~))
 import Text.Regex.TDFA.Text()
@@ -185,9 +186,11 @@ suggestDeleteTopBinding :: ParsedModule -> Diagnostic -> [(T.Text, [TextEdit])]
 suggestDeleteTopBinding ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecls}} Diagnostic{_range=_range,..}
 -- Foo.hs:4:1: warning: [-Wunused-top-binds] Defined but not used: ‘f’
     | Just [name] <- matchRegex _message ".*Defined but not used: ‘([^ ]+)’"
-    , let
-        allTopLevel = filter (isTopLevel . fst) $ map (\(L l b) -> (srcSpanToRange l, b)) hsmodDecls
-        sameName = filter (matchesBindingName (T.unpack name) . snd) allTopLevel
+    , let allTopLevel = filter (isTopLevel . fst)
+                        . map (\(L l b) -> (srcSpanToRange l, b))
+                        . sortLocated
+                        $ hsmodDecls
+          sameName = filter (matchesBindingName (T.unpack name) . snd) allTopLevel
             = [("Delete ‘" <> name <> "’", flip TextEdit "" . toNextBinding allTopLevel . fst <$> sameName )]
     | otherwise = []
     where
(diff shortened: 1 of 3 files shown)


test/exe/Main.hs
```diff
@@ -1144,6 +1145,40 @@ insertNewDefinitionTests = testGroup "insert new definition actions"
...
+                     <$> getCodeActions docId (R 0 0 0 0)
+
+      liftIO $ actionTitle @?= "Delete ‘f’"
+      executeCodeAction action
+      contentAfterAction <- documentContents docId
+      liftIO $ contentAfterAction @?= T.unlines [
+        "{-# OPTIONS_GHC -Wunused-top-binds #-}"
+        , "module A (some) where"
+        , ""
+        , ""
```
ndmitchell, line 1176: A very minor remark, but beforehand they had 1-line gaps, now they have 2-line gaps. Should we expand one line after if it's blank? Or maybe this is getting ahead of ourselves and there is no need.

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -180,6 +181,24 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
...
+suggestDeleteTopBinding :: ParsedModule -> Diagnostic -> [(T.Text, [TextEdit])]
+suggestDeleteTopBinding ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecls}} Diagnostic{_range=_range,..}
+-- Foo.hs:4:1: warning: [-Wunused-top-binds] Defined but not used: ‘f’
+    | Just [name] <- matchRegex _message ".*Defined but not used: ‘([^ ]+)’"
+    , (_character . _start) _range == 0
+    , let sameName = filter (\(L l b) -> isTopLevel l && matchesBindingName b (T.unpack name)) hsmodDecls
+            = [("Delete ‘" <> name <> "’", [TextEdit (nextLine . srcSpanToRange $ l) "" | (L l _) <- sameName])]
+    | otherwise = []
+    where
+      isTopLevel l = (_character . _start) (srcSpanToRange l) == 0
```
cocreature, line 193: You are basically duplicating this logic in line 188. How about we change this to accept a range and then we can use it for both cases (composing with `srcSpanToRange` at one call site).

test/exe/Main.hs
cocreature: Makes sense to me. What do you think @serhiip?

serhiip:
> This looks really good. Thanks for your contribution!
> 
> My guess is this doesn't work for operators, or functions defined like `` x `f` y = ... ``? Not a problem, as they are vastly more rare, and this is a code action.

Good point. And actually there was a bug :). Luckily it was easy to fix. I pushed the fixed version and added corresponding test (just in case). It was a one line fix so I've incorporated it together with other changes I did.

src/Development/IDE/Plugin/CodeAction.hs
serhiip: Indeed! There was a bug with infix bindings, so I've eradicated this line to fix that bug

test/exe/Main.hs
serhiip: For sure expanding the text edit until the next binding in a file is a good thing. I was thinking about this initially but only after your comment I've figured how to actually do it: now the code works by finding the next binding after current one and expands the deletion until that next binding (if any). To make that happen I rearranged my code a bit. Hope it makes more sense now. Thanks for suggestion

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -180,6 +181,30 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
...
+        sameName = filter (matchesBindingName (T.unpack name) . snd) allTopLevel
+            = [("Delete ‘" <> name <> "’", flip TextEdit "" . toNextBinding allTopLevel . fst <$> sameName )]
+    | otherwise = []
+    where
+      isTopLevel l = (_character . _start) l == 0
+
+      forwardLines lines r = r {_end = (_end r) {_line = (_line . _end $ r) + lines, _character = 0}}
+
+      toNextBinding bindings r@Range { _end = Position {_line = l} }
+        | Just (Range { _start = Position {_line = l'}}, _) <- find ((> l) . _line . _start . fst) bindings
```
cocreature, line 199: Is `hsmodDecls` guaranteed to be sorted by ranges? Looking at haddock’s `collectDocs` it always seems to be called after `sortLocated` suggesting that we probably need this here as well.

serhiip: Order in `hsmodDecls` doesn't matter really - I use `find` to locate next binding within all the top level bindings. I might have missed the point why `hsmodDecls` must be sorted

cocreature: `find` will give you the first match. where the condition is true, consider what happens if you have 3 bindings. You want to delete the first. Assume `hsModDecls` has the order `1, 3, 2`. Then your `find` here will find `3` and you will end up deleting `2` as well

serhiip: My bad. That is indeed an issue - I've pushed an update. Thanks for catching this

PR #661, opened 2020-06-22
Retry GHC 8.10 on Windows

cocreature:
It keeps crashing and annoying everyone. The issue is in GHC not in
our code and I believe it’s fixed in HEAD already but that doesn’t
help us so let’s add some retries for now.


jneira:
The ghc bug is weird but it is even more when three retries can do the workaround 🤦

jrp2014:
I too get lots of random Windows crashes in another (plugin) application.

In my case, the issue seems to be related to the runtime dynamic linker and its adverse impact on DynFlags.

PR #662, opened 2020-06-23
Fix debouncer for 0 delay

pepeiborra:
The indirection caused by `async (sleep 0 >> fire)` was causing non-deterministic ordering of progress done messages and diagnostics, which makes the code actions benchmark experiment fail randomly.

### DIFF
--- a/src/Development/IDE/Core/Debouncer.hs
+++ b/src/Development/IDE/Core/Debouncer.hs
@@ -39,10 +39,11 @@ newAsyncDebouncer = Debouncer . asyncRegisterEvent <$> newVar Map.empty
 -- Events are run unmasked so it is up to the user of `registerEvent`
 -- to mask if required.
 asyncRegisterEvent :: (Eq k, Hashable k) => Var (HashMap k (Async ())) -> Seconds -> k -> IO () -> IO ()
-asyncRegisterEvent d 0 k fire = modifyVar_ d $ \m -> mask_ $ do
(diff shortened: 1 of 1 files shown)


src/Development/IDE/Core/Debouncer.hs
```diff
@@ -39,6 +39,10 @@ newAsyncDebouncer = Debouncer . asyncRegisterEvent <$> newVar Map.empty
 -- Events are run unmasked so it is up to the user of `registerEvent`
 -- to mask if required.
 asyncRegisterEvent :: (Eq k, Hashable k) => Var (HashMap k (Async ())) -> Seconds -> k -> IO () -> IO ()
+asyncRegisterEvent d 0 k fire = modifyVar_ d $ \m -> mask_ $ do
+    whenJust (Map.lookup k m) cancel
+    fire
```
pepeiborra, line 44: Oops, forgot to unmask here

cocreature: Should we just disable the debouncer in the tests?

pepeiborra: But this is a real issue. If we start disabling things for test, we end up testing the wrong thing.

cocreature: Makes sense :+1:

pepeiborra: fixed

PR #666, opened 2020-06-24
Make BenchHist non buildable by default and save logs

pepeiborra:
As the title says


bench/hist/Main.hs
```diff
@@ -178,37 +178,37 @@ main = shakeArgs shakeOptions {shakeChange = ChangeModtimeAndDigest} $ do
...
         let ghcide = dropFileName outcsv </> "ghcide"
             ghcpath = dropFileName outcsv </> "ghc.path"
         need [ghcide, ghcpath]
         ghcPath <- readFile' ghcpath
-        verb <- getVerbosity
         withResource ghcideBenchResource 1 $ do
           Stdout res <-
             command
-              [ EchoStdout True,
+              [ EchoStdout False,
```
ndmitchell, line 195: Have you seen `FileStdout` https://hackage.haskell.org/package/shake-0.19.1/docs/Development-Shake.html#v:FileStdout? Could that be used to avoid reading it as a buffer, then writing it to a file?

pepeiborra: Oh, that's useful, thanks!

pepeiborra: I really like `cmd_` and friends btw, they are super ergonomic.

ndmitchell: Thanks! I think I've spent about as long on cmd_ as I have on all of the rest of Shake :)

pepeiborra:
I think I've figured out a nicer way to turn benchHist into a benchmark. Moving to draft until I've tested it on both stack and Cabal ....

PR #677, opened 2020-06-30
Use a global namecache to read `.hie` files

wz1000:
This backports [this GHC MR](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/3199) in preparation for a larger role for `.hie` files in ghcide.

We now have a global shared `NameCache` referenced by `ShakeExtras`, which we use to read `.hie` files.

### DIFF
--- a/src-ghc810/Development/IDE/GHC/HieBin.hs
+++ b/src-ghc810/Development/IDE/GHC/HieBin.hs
@@ -1,0 +1,393 @@
+{-
+Binary serialization for .hie files.
+-}
+{-# LANGUAGE ScopedTypeVariables #-}
+module Development.IDE.GHC.HieBin ( readHieFile, readHieFileWithVersion, HieHeader, writeHieFile, HieName(..), toHieName, HieFileResult(..), hieMagic, hieNameOcc,NameCacheUpdater(..)) where
+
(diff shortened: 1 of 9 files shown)


src-ghc810/Development/IDE/GHC/HieBin.hs
```diff
@@ -0,0 +1,394 @@
+{-
```
cocreature, line 1: I’m increasingly starting to wonder if GHC is the right location for the `.hie` files code. We seem to copy paste basically everything since we need to patch it. Hopefully, things will stabilize at some point :disappointed:

cocreature:
Needs rebasing.

src-ghc810/Development/IDE/GHC/HieBin.hs
wz1000: It's used by haddock, so there isn't much of an alternative.

wz1000:
I found a bug introduced by this, fixed it and modified the tests to exercise this code path. See the commit message of https://github.com/digital-asset/ghcide/pull/677/commits/96d1f30ff84dc87d59118630d6b6dbfb for an explanation.

cocreature:
Thanks for the bugfix!

PR #678, opened 2020-06-30
Fix spaninfo Haddocks for local modules

pepeiborra:
The regression was introduced in #630.

I added `GhcSessionDeps` with the idea of reusing the typecheck GHC session
for computing the SpanInfo, instead of rebuilding it from scratch.

But I forgot to actually reuse it, or maybe the change got lost during the merge.

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -468,7 +468,7 @@ getSpanInfoRule :: Rules ()
 getSpanInfoRule =
     define $ \GetSpanInfo file -> do
         tc <- use_ TypeCheck file
-        packageState <- hscEnv <$> use_ GhcSession file
+        packageState <- hscEnv <$> use_ GhcSessionDeps file
 
 -- When possible, rely on the haddocks embedded in our interface files
 -- This creates problems on ghc-lib, see comment on 'getDocumentationTryGhc'
--- a/test/data/hover/Bar.hs
+++ b/test/data/hover/Bar.hs
@@ -1,3 +1,4 @@
 module Bar (Bar(..)) where
 
+-- | Bar Haddock
 data Bar = Bar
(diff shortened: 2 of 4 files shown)


ndmitchell:
Is there any test that will stop us breaking it again in the future?

pepeiborra:
Benchmarks here - https://github.com/pepeiborra/ghcide/tree/fix-spaninfo-benchmarks/bench-hist

Sadly the `edit` experiment regresses quite a bit - it takes twice as long. 
Keep in mind that edits now ask for `GetSpanInfo` too.
This happens because now the GHC session does contain matches for the doc lookups.
I suppose asking GHC for benchmarks is really slow! 
Or more likely, our code to convert Haddock markups to markdown. 
Will need more investigation (but perhaps in another PR)

https://github.com/pepeiborra/ghcide/blob/fix-spaninfo-benchmarks/bench-hist/results.csv

pepeiborra:
> Is there any test that will stop us breaking it again in the future?

Good question. There are tests for this, but they fall short of checking for comments!
Should be an easy fix, to be added in this PR

pepeiborra:
I have extended the tests that were intended to verify for this condition, and checked that they would fail without this patch

pepeiborra:
The profiler points towards GHC, although there are no details below `ioMsgMaybe`
[profile of edit experiment (1000 samples).zip](https://github.com/digital-asset/ghcide/files/4852150/profile.of.edit.experiment.1000.samples.zip)

pepeiborra:
I have looked at the performance regression after switching to `GhcSessionDeps` which has the effect of correctly loading all the dependencies in the Ghc session: and it's attributed to:

1. The `getDocumentationTryGhc`, which wasn't doing much previously for home package names.
2. The `lookupName` call, idem, and
3. The `desugarExpr` call, which becomes much more expensive too for some reason. 

I suspect that all three cases are simply providing more answers and doing more work now that the Ghc session is correctly loaded with the dependencies. 

There are some low hanging fruits in cases 2 and 3 which I'll address in a separate PR.

cocreature:
Thanks for the analysis. Happy to leave this for a separate PR.

PR #680, opened 2020-07-01
Code action add default type annotation to remove `-Wtype-defaults` warning

serhiip:
This PR adds a code action to add default type annotation suggested by GHC when `-Wtype-defaults` is enabled. 

I noticed that during development, when I use `traceShow` from `Debug.Trace` I often get annoyed by sudden warning displayed, so I usually urged to add that default type annotation manually. These changes add a way to do that via a hotkey.

I hope this is useful to someone else.

These changes handle only literals, not expressions. For example for binding `f = 2 + 2` (without type signature defined) the warning will be shown, but no code action produced.

I've tried to handle all the cases of the appearance of this warning, and I am interested in finding more examples to handle (if there is some).

If this doesn't seem like a useful addition, I am fine with not merging it

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -1210,9 +1210,7 @@ deleteUnusedDefinitionTests = testGroup "delete unused definition action"
       liftIO $ contentAfterAction @?= expectedResult
 
     extractCodeAction docId actionPrefix = do
-      Just (CACodeAction action@CodeAction { _title = actionTitle })
-                  <- find (\(CACodeAction CodeAction{_title=x}) -> actionPrefix `T.isPrefixOf` x)
-                     <$> getCodeActions docId (R 0 0 0 0)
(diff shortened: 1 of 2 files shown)


test/exe/Main.hs
```diff
@@ -2639,19 +2735,25 @@ openTestDataDoc path = do
...
+
+findCodeActions' :: (T.Text -> T.Text -> Bool) -> String -> TextDocumentIdentifier -> Range -> [T.Text] -> Session [CodeAction]
+findCodeActions' op errMsg doc range expectedTitles = do
   actions <- getCodeActions doc range
   let matches = sequence
         [ listToMaybe
           [ action
           | CACodeAction action@CodeAction { _title = actionTitle } <- actions
-          , actionTitle == expectedTitle ]
+          , expectedTitle `op` actionTitle]
```
serhiip, line 2750: Here I swapped the order of application of arguments. That matters only for `T.isPrefixOf` (the version which I added), not for `==`

PR #681, opened 2020-07-02
Performance improvements for GetSpanInfo

pepeiborra:
Avoiding duplicate work and batching ghc-api lookups for both `getDocs` and `lookupName` are the main sources of improvement. 

As a bonus I also removed usage of the GHCi code paths. Since there is no way for me to test if this fixes the ghc-lib issues, I'm happy to put the CPP pragmas back if that's preferred.

edit

Full benchmarks in: https://github.com/pepeiborra/ghcide/tree/perf-spaninfo-benchmark

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -43,10 +43,10 @@ import Development.IDE.Types.Options
 import Development.IDE.Types.Location
 
 #if MIN_GHC_API_VERSION(8,6,0)
-import           DynamicLoading (initializePlugins)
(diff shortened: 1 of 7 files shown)


cocreature:
Looks like this needs rebasing.

pepeiborra:
Rebased.

I'd rather put back the CPP exclusions, merge this now, and then send a separate PR to remove them which can be tested in due time. Might be able to do this later today

pepeiborra:
Thanks! I'll submit a PR to restore the CPP sections later today or tomorrow.

PR #690, opened 2020-07-12
Fix `.hie` file location for `.hs-boot` files

wz1000:


pepeiborra:
What are `.hie.boot` files used for? Is there any way to write a test that will prevent this from breaking silently?

wz1000:
They are the same as `.hie` files, but are generated for `.hs-boot` files instead of `.hs` files. I will try to write a test for this, but it might be difficult since we don't use `.hie` files for much as of now.

pepeiborra:
Goto definition is driven by `.hie` files, so maybe the test could leverage that

wz1000:
I discovered a bug where we were trying to load the wrong interface for `hs-boot` files (loading the `hi` file instead of the `hi-boot` file) which meant that the interfaces for boot files were regenerated every time.

To fix this, I moved all the logic for adding the `-boot` suffice to artifacts directly to where the `ModSummary`s are generated. 

I've also added a testcase.

src/Development/IDE/GHC/Compat.hs
```diff
@@ -380,6 +384,14 @@ instance HasSrcSpan (GenLocated SrcSpan a) where
...
     catch (Right <$> Hdr.getImports a b c d)
           (return . Left . srcErrorMessages)
+
+-- | Add the @-boot@ suffix to all output file paths associated with the
+-- module, not including the input file itself
+addBootSuffixLocnOut :: GHC.ModLocation -> GHC.ModLocation
+addBootSuffixLocnOut locn
+  = locn { ml_hi_file  = Module.addBootSuffix (ml_hi_file locn)
+         , ml_obj_file = Module.addBootSuffix (ml_obj_file locn)
+         }
```
pepeiborra, line 394: Why is the field for hie files (`ml_hie_file`) not handled in the same way?

wz1000: If you expand the fold a bit above, you will find that this bit is in the `#else` block of a `#if MIN_GHC_API_VERSION(8,8,0)`. So the answer is that `.hie` files (and the `ml_hie_file` field) don't exist in this piece of code.

See also the definition of `ml_hie_file` as a function for GHC < 8.8 above.

pepeiborra (approved):

PR #691, opened 2020-07-12
Show documentation on hover for symbols defined in the same module

wz1000:
When parsing a module, if parsing haddocks succeeds, then use them
Previously, even though we were parsing modules twice, with and without
haddocks, we were just returning the result of parsing without haddocks.

The reason for this was that Opt_KeepRawTokenStream and Opt_Haddock do
not interact nicely. We decided that for now it was better to fix an
actual issue and then solve the problem when hlint requires a module
with Opt_KeepRawTokenStream.

### DIFF
--- a/src/Development/IDE/Core/RuleTypes.hs
+++ b/src/Development/IDE/Core/RuleTypes.hs
@@ -61,6 +61,9 @@ type instance RuleResult GetKnownFiles = HS.HashSet NormalizedFilePath
 -- that module.
 data TcModuleResult = TcModuleResult
     { tmrModule     :: TypecheckedModule
+    -- ^ warning, the ModIface in the tm_checked_module_info of the
+    -- TypecheckedModule will always be Nothing, use the ModIface in the
+    -- HomeModInfo instead
     , tmrModInfo    :: HomeModInfo
     , tmrDeferedError :: !Bool -- ^ Did we defer any type errors for this module?
     }
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -233,6 +233,12 @@ priorityGenerateCore = Priority (-1)
 priorityFilesOfInterest :: Priority
 priorityFilesOfInterest = Priority (-2)
 
+-- | IMPORTANT FOR HLINT INTEGRATION:
+-- We currently parse the module both with and without Opt_Haddock, and
+-- return the one with Haddocks if it -- succeeds. However, this may not work
+-- for hlint, and we might need to save the one without haddocks too.
+-- See https://github.com/digital-asset/ghcide/pull/350#discussion_r370878197
+-- and https://github.com/mpickering/ghcide/pull/22#issuecomment-625070490
 getParsedModuleRule :: Rules ()
 getParsedModuleRule = defineEarlyCutoff $ \GetParsedModule file -> do
     sess <- use_ GhcSession file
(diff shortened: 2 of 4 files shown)


src/Development/IDE/Core/RuleTypes.hs
```diff
@@ -50,6 +50,9 @@ type instance RuleResult GetDependencies = TransitiveDependencies
 -- that module.
 data TcModuleResult = TcModuleResult
     { tmrModule     :: TypecheckedModule
+    -- ^ warning, the ModIface in the tm_checked_module_info of the
+    -- TypecheckedModule will always be Nothing, use the ModIface in the
+    -- HomeModInfo instead
```
pepeiborra, line 55: This is very helpful, thanks!

pepeiborra:
(a test would be very nice)

wz1000:
@pepeiborra the test already existed :)

(good to see a CI failure because of `unexpected success`)

jneira:
In hls [we are about to add hlint support](https://github.com/haskell/haskell-language-server/pull/166) including refactoring and it will be broken with this. Could be return both version of the parsed modules and make hlint (and other tools that needs the original parsed module version) use the existing one? 
It would duplicate the memory used (not sure it it could be an issue), but only if you are using hlint.

I think @shayne_fletcher [mentioned](https://github.com/haskell/haskell-language-server/pull/166#issuecomment-647783953) some wip to make compatible the parsing with haddock options though.

wz1000:
@jneira I'm confused, this patch is already included in the branch of ghcide that hls is using.

jneira:
Yeah, but as we have to make it optional for daml hlint i was proposing another way to do it, make two versions available at runtime instead only one or another using cpp.
I have to admit i've still not hit a issue in my tests of the hlint plugin that i can confirm it is caused by the haddock version of the parsed module.

pepeiborra:
@wz1000 I think we need a configuration option before this can be merged to avoid a breaking change, unless @cocreature no longer needs it

wz1000:
Ok, I can do that. However, I think @jneira has hlint working in the presence of this patch.

wz1000:
@pepeiborra done and rebased.

jneira:
@wz1000 i've not seen any error using the hlint plugin + GetParsedModule in hls so far, using a ghcide version with this patch.
But @cocreature mentioned that they were incompatible, maybe cause they are using an older version?

pepeiborra (approved):

cocreature:
They are not fully incompatible. But the haddock parsing options doesn’t include some stuff, e.g., extensions iirc so HLint won’t show you the corresponding hints if you use haddock parsing. That said, I don’t care all that much about those hints (neither here nor in DAML) so it’s probably reasonable to go for this approach for now. Hopefully, this will be addressed on GHC’s side at some point.

wz1000:
Is it not possible to manually tell hlint about extensions etc. by extracting them from the dynflags?

/cc @ndmitchell

ndmitchell:
You can tell HLint that the code should be parsed with the OverloadedStrings or GADTs extension just fine. So HLint will still be able to parse the file.

You can't tell HLint that the `{-# LANGUAGE GADTs #-}` pragma was used other than by it being in the source code. HLint can warn if certain extensions are unused, or duplicates, or imply other hints, so you wouldn't get those hints.

pepeiborra:
Hlint also supports ANN pragmas for annotations which presumably need full tokens

PR #692, opened 2020-07-12
Code action: remove redundant constraints for type signature

### LINKED ISSUE #60: Code action: remove redundant constraints
Should be a good first issue if identifying the right bit of code to remove is simple enough (as weith #54).

If the `-Wredundant-constraints` warning is on, GHC emits warnings like:

```
C:\msys64\home\ganesh\temp\x\Foo.hs:3:1: warning:
    • Redundant constraint: Ord a
    • In the type signature for:
           foo :: forall a. Ord a => a -> Int
```

Unfortunately the source range covers all of the type signature for `foo`, not just the redundant constraint itself.

DenisFrezzato:
This closes #60.


src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -586,6 +587,78 @@ suggestFunctionConstraint contents Diagnostic{..} missingConstraint
...
+      findRedundantConstraints :: T.Text -> Maybe [T.Text]
+      findRedundantConstraints t = t
+        & T.lines
+        & head
+        & T.strip
+        & (`matchRegex` "Redundant constraints?: (.+)")
+        <&> (head >>> parseConstraints)
+
+      findConstraints :: T.Text -> T.Text -> T.Text
+      findConstraints contents typeSignatureName = contents
```
DenisFrezzato, line 634: On a second thought, I've realised that this will fail if the type signature is not formatted as expected, for example with an arbitrary number of spaces or written over multiple lines (see also lines 608 and 611).

While I don't reckon it could brake the code, this is rather poorly handled. I see two ways to improve it:
* trivial one: make it a bit safer and do nothing if the code is not formatted as expected;
* trickier: improve the logic in order to handle any format. I'm not sure this would be achievable with regular expressions and `Text` functions though, so maybe some text parsing should be involved?

What do you think?

ndmitchell: Usually do the trivial one now, and then handle the trickier one in a follow up PR if its a problem in practice.

cocreature: Doing the trivial one now is totally fine but we should do nothing if the formatting doesn’t match our expectations instead of crashing which the `last` and `head` calls will do.

DenisFrezzato:
@cocreature Yes, that is exactly what I was planning to do, but I didn't have time yet. Since I'm hanging here, is it fine with you if I add `safe` as a dependency?

cocreature:
Yes `safe` is fine.

PR #694, opened 2020-07-12
Populate ms_hs_date in GetModSummary rule

pepeiborra:
This is a straightforward fix that came up during https://github.com/haskell/haskell-language-server/pull/191

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -84,8 +84,6 @@ import           System.IO.Extra
 import Control.DeepSeq (rnf)
 import Control.Exception (evaluate)
 import Exception (ExceptionMonad)
-import Data.Time (getCurrentTime)
-import System.IO.Error (isDoesNotExistError)
 import TcEnv (tcLookup)
 
 
@@ -460,10 +458,6 @@ getModSummaryFromImports
   -> Maybe SB.StringBuffer
   -> ExceptT [FileDiagnostic] m ModSummary
 getModSummaryFromImports fp contents = do
-    fileModTime <- liftIO $ getModificationTime fp `catch` \e ->
-        if isDoesNotExistError e
-            then getCurrentTime -- Virtual file case
-            else throwIO e
     (contents, dflags) <- preprocessor fp contents
     (srcImports, textualImports, L _ moduleName) <-
         ExceptT $ liftIO $ first (diagFromErrMsgs "parser" dflags) <$> GHC.getHeaderImports dflags contents fp fp
@@ -482,7 +476,7 @@ getModSummaryFromImports fp contents = do
 #if MIN_GHC_API_VERSION(8,8,0)
                 , ms_hie_date     = Nothing
 #endif
-                , ms_hs_date      = fileModTime
+                , ms_hs_date      = error "use GetModSummary instead of GetModSummaryWithoutTimestamps"
                 , ms_hsc_src      = sourceType
                 -- The contents are used by the GetModSummary rule
                 , ms_hspp_buf     = Just contents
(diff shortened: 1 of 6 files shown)


src/Development/IDE/Core/Compile.hs
```diff
@@ -451,6 +453,10 @@ getModSummaryFromImports
   -> Maybe SB.StringBuffer
   -> ExceptT [FileDiagnostic] m ModSummary
 getModSummaryFromImports fp contents = do
+    fileModTime <- liftIO $ getModificationTime fp `catch` \e ->
```
cocreature, line 456: Is there a reason why this is not using the `GetModificationTime` rule? I’m somewhat hesitant to introduce a `fileModTime` that will change even if the document version does not change for a virtual file.

pepeiborra: That's a good point, this should be using the rule. Will fix

pepeiborra:
I extended this change in the following directions:

1. Platform dependent code to convert `FileVersion` to `Maybe UTCTime`. Further improvements possible in Mingw platforms to avoid `UTCTime -> FileVersion -> Maybe UTCTime`
2. Add timestamps to the `ModSummary` code path for parsing: we need both parsing and interfaces mod summaries to have timestamps
3. Added a new Shake rule `ModSummaryWithoutTimestamp` to preserve the existing early cutoff semantics and avoid a perf regression.

I tested the changes with the Eval plugin in haskell-language-server, branch: http://github.com/pepeiborra/ide/tree/eval

Please take another look and let me know

src/Development/IDE/Core/Compile.hs
pepeiborra: Sorted

pepeiborra:
Benchmarks showing no perf regressions: https://github.com/pepeiborra/ghcide/tree/getmodsummary-modtime-benchmarks/bench-hist

src/Development/IDE/Core/Rules.hs
```diff
@@ -642,6 +646,8 @@ ghcSessionDepsDefinition file = do
 
 getModIfaceFromDiskRule :: Rules ()
 getModIfaceFromDiskRule = defineEarlyCutoff $ \GetModIfaceFromDisk f -> do
+    -- We do want the timestamps in the ModSummary for the interactive evaluation functions
```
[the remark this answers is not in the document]
pepeiborra: The main goal is to avoid inserting bottom modification times in the ModSummary values that we craft, because the interactive evaluation functions in the ghc api perform their own recompilation checking and would crash.

For files of interest, we over approximate the modification time - it is correct but it is not precise - I don't think it matters too much. For other files we get it from the file system.

src/Development/IDE/Core/Rules.hs
```diff
@@ -679,15 +685,30 @@ isHiFileStableRule = define $ \IsHiFileStable f -> do
...
-        getModSummaryFromImports (fromNormalizedFilePath f) (textToStringBuffer <$> mFileContent)
-    case modS of
-        Right ms -> do
-            return ( Just (computeFingerprint f dflags ms), ([], Just ms))
-        Left diags -> return (Nothing, (diags, Nothing))
+getModSummaryRule = do
+    defineEarlyCutoff $ \GetModSummary f -> do
+        dflags <- hsc_dflags . hscEnv <$> use_ GhcSession f
+        (fv, mFileContent) <- getFileContents f
+        modTime <- maybe (liftIO getCurrentTime) return $ modificationTime fv
```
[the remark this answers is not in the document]
pepeiborra: I have moved the logic to `getFileContents` since all the call sites want it.`GetModIfaceFromDisk` is the only use site that really requires timestamps for the interactive evaluation to work

PR #695, opened 2020-07-12
Fix bugs with top level binding deletion code action and comments

serhiip:
This PR fixes a couple of bugs with code changes introduced in #657 : 
  * Top-level binding deletion code action produces code edit that also deletes the haddock comment (or any comment) for the successive binding (which should not happen)
  * Top-level binding deletion code action produces code edit that fails to remove the haddock comment associated with the current binding

I tried using GHC API to fix those problems, but I found it is not possible to get the nodes for comments or haddock comments: `DocD` never appeared in a `hsmodDecls` (as per documentation https://hackage.haskell.org/package/ghc-8.10.1/docs/GHC-Hs-Decls.html#t:HsDecl)

I added a code that does a reversal on an array of lines un current file - it's linear - this addition should not hit performance hard.

I will be glad to hear how I can improve this

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -150,7 +150,7 @@ suggestAction dflags packageExports ideOptions parsedModule text diag = concat
     [  suggestNewDefinition ideOptions pm text diag
     ++ suggestRemoveRedundantImport pm text diag
     ++ suggestNewImport packageExports pm diag
-    ++ suggestDeleteTopBinding pm diag
+    ++ suggestDeleteTopBinding text pm diag
     | Just pm <- [parsedModule]]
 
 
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -173,27 +173,33 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
...
 
-suggestDeleteTopBinding :: ParsedModule -> Diagnostic -> [(T.Text, [TextEdit])]
-suggestDeleteTopBinding ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecls}} Diagnostic{_range=_range,..}
+suggestDeleteTopBinding :: Maybe T.Text -> ParsedModule -> Diagnostic -> [(T.Text, [TextEdit])]
+suggestDeleteTopBinding sourceOpt ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecls}} Diagnostic{_range=_range,..}
 -- Foo.hs:4:1: warning: [-Wunused-top-binds] Defined but not used: ‘f’
     | Just [name] <- matchRegex _message ".*Defined but not used: ‘([^ ]+)’"
-    , let allTopLevel = filter (isTopLevel . fst)
+    , let allTopLevel = filter ((== 0) . (_character . _start) . fst)
                         . map (\(L l b) -> (srcSpanToRange l, b))
```
pepeiborra, line 181: I would fuse the `filter` and `map` using a list comprehension or `mapMaybe`

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -173,27 +173,33 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
...
     | otherwise = []
 
-suggestDeleteTopBinding :: ParsedModule -> Diagnostic -> [(T.Text, [TextEdit])]
-suggestDeleteTopBinding ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecls}} Diagnostic{_range=_range,..}
+suggestDeleteTopBinding :: Maybe T.Text -> ParsedModule -> Diagnostic -> [(T.Text, [TextEdit])]
+suggestDeleteTopBinding sourceOpt ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecls}} Diagnostic{_range=_range,..}
 -- Foo.hs:4:1: warning: [-Wunused-top-binds] Defined but not used: ‘f’
     | Just [name] <- matchRegex _message ".*Defined but not used: ‘([^ ]+)’"
-    , let allTopLevel = filter (isTopLevel . fst)
+    , let allTopLevel = filter ((== 0) . (_character . _start) . fst)
```
pepeiborra, line 180: I would bring back the `isTopLevel` helper

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -173,27 +173,33 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
...
                         . map (\(L l b) -> (srcSpanToRange l, b))
                         . sortLocated
                         $ hsmodDecls
           sameName = filter (matchesBindingName (T.unpack name) . snd) allTopLevel
     , not (null sameName)
-            = [("Delete ‘" <> name <> "’", flip TextEdit "" . toNextBinding allTopLevel . fst <$> sameName )]
+            = [("Delete ‘" <> name <> "’", flip TextEdit T.empty . toNextNonEmptyLine allTopLevel . fst <$> sameName )]
```
pepeiborra, line 186: I would break this long line down by extracting a `mkTextEdit = flip TextEdit T.empty` helper

cocreature:
@pepeiborra raises a good point. This is the difference between `Opt_Haddock` and `Opt_KeepRawTokenStream`. For the former you get `DocD` nodes whereas for the latter you get annotations

serhiip:
Small update: was able the start seeing some parsed comment tokens inside my code action handler after I added Opt_KeepRawTokenStream flag (thanks for the link). Current issue is that not all tokens have the location attached to it (SrcSpan) so I see `[(<no location info>,\n  [AnnLineComment \"-- | This should be deleted\",\n   AnnBlockComment \"{-# OPTIONS_GHC -Wunused-top-binds #-}\"])]`.

Still trying exactly when this happens and if there is way to avoid getting that `<no location info>`

alanz:
@serhiip the comments are normally attached to the nearest enclosing AST element.  The ones at the top and bottom of the file cannot be associated with anything specific, so they are  captured against the <no-location-info> span.  Which is effectively the top of the file.

alanz:
And you might be better off using ghc-exactprint for this, which already manages comments, alignment and so on. And is used for various formatters and retrie too.

serhiip:
@alanz indeed, when I changed my test ta have following input

```
    (T.unlines [ "{-# OPTIONS_GHC -Wunused-top-binds #-}"
               , "module A (some, some2) where"
               , ""
               , "some2 = ()"
               , ""
               , "-- | This should be deleted (new)"
               , "documented :: Int -> Int -> Int"
               , "documented _ b = let a = 1"
               , "                 -- comment"
               , "                 in a + b"
               , ""
               , "some = ()"
               , "-- irrelevant"
               ])
```

it gives me back 

```
[(/private/var/folders/rv/fvmp1h3x2yn40ptnsbpz2fl40000gn/T/extra-dir-45990137941/A.hs:(8,18)-(10,25),\n  [AnnLineComment \"-- comment\"]),\n (<no location info>,\n  [AnnLineComment \"-- irrelevant\",\n   AnnLineComment \"-- | This should be deleted (new)\",\n   AnnBlockComment \"{-# OPTIONS_GHC -Wunused-top-binds #-}\"])]
```

Which is not really helpful for what I am trying to do (delete top level binding and it's documentation annotation).

I will look into `ghc-exactprint` (looks promising)

Thanks for clarification and suggestion

src/Development/IDE/Plugin/CodeAction.hs
wz1000: While this may work for most haskell files out in the wild, it is not strictly correct. For instance, the following is a valid haskell module:

```haskell
module Main (main) where
  main :: IO ()
  main = putStrLn "hello"

```

(I understand this bug was not introduced by this PR)

pepeiborra:
Note that $710 reimplements the deletion code action and presumably fixes all the issues, except that it doesn't try to address comments

pepeiborra:
closing due to inactivity

PR #696, opened 2020-07-13
Obtain the GHC libdir at runtime

bubba:
#693 needs to be merged first

This replaces hardcoding the GHC libdir path with ghc-paths and instead
gets it at runtime through the hie-bios cradle. This means that the
ghcide binary should be a bit more distributable now, since it won't
rely on paths baked at compile time that are local to the machine it was
compiled on

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -518,7 +518,7 @@ setCacheDir logger prefix hscComponents comps dflags = do
 
 
 renderCradleError :: NormalizedFilePath -> CradleError -> FileDiagnostic
-renderCradleError nfp (CradleError _ec t) =
+renderCradleError nfp (CradleError _ _ec t) =
   ideErrorWithSource (Just "cradle") (Just DsError) nfp (T.unlines (map T.pack t))
 
 -- See Note [Multi Cradle Dependency Info]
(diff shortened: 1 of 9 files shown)


exe/Main.hs
```diff
@@ -192,21 +195,35 @@ showEvent lock e = withLock lock $ print e
...
+        -- Same here
         CradleNone -> return (Left [])
 
-emptyHscEnv :: IORef NameCache -> IO HscEnv
-emptyHscEnv nc = do
-    libdir <- getLibdir
-    env <- runGhc (Just libdir) getSession
+emptyHscEnv :: IORef NameCache -> FilePath -> IO HscEnv
+emptyHscEnv nc libDir = do
+    -- We always
```
pepeiborra, line 225:
```suggestion
```

exe/Main.hs
```diff
@@ -250,181 +267,184 @@ loadSession dir = do
...
+              -- scratch again (for now)
+              -- It's important to keep the same NameCache though for reasons
+              -- that I do not fully understand
+              logInfo logger (T.pack ("Making new HscEnv" ++ show inplace))
+              hscEnv <- emptyHscEnv ideNc libDir
+              newHscEnv <-
+                -- Add the options for the current component to the HscEnv
+                evalGhcEnv hscEnv $ do
+                  _ <- setSessionDynFlags df
+                  checkSession logger ghcLibCheck
```
pepeiborra, line 334: I believe that the `ghcLibCheck` is no longer necessary, as the installation check should be accurate enough now that we are comparing the ghc installation that was used to compile *ghcide* with the one that is being used to run it.

PR #701, opened 2020-07-17
Save source with HIE files

### LINKED ISSUE #671: Generated .HIE files incompatible with STAN
Stan doesnt work with HIE files generated by Haskell Language Server / ghcide.

The issue was described in reddit and the explanation of what is missing:
>I think I know what happens. When contributing GHC-8.10 support in ghcide, I noticed that they create HIE files artificially and they erase source content from HIE files completely. That's why you see an error about too large index — there's simply no source code info in HIE files.
>
> We can check indexes on our side in Stan, but you will see blank lines in report, which won't make it that useful to users.
>
> If this is indeed what is happening, what you can do is to try to discuss this issue with the Haskell IDE engine team, reporting the bug to them. I see how this issue can break the integration of different tools and that resolving it can help different tools not to interfere with each other. I believe it's a good idea, in general, to build tooling around the same API and same expectations, otherwise we will have the broken ecosystem in area of IDE.

I confirmed in ghcide that the source bytestring is empty: https://github.com/digital-asset/ghcide/blob/master/src/Development/IDE/Core/Compile.hs#L281 and https://github.com/digital-asset/ghcide/blob/master/src-ghc88/Development/IDE/GHC/HieAst.hs#L94

Issue originally posted to https://github.com/kowainik/stan/issues/327#issue-646721049

I attempted to patch the linked code to use `getFileContents` to write the source to the HIE files, but it didnt work (e.g. the same error was thrown). I am happy to attempt a PR, but I am wondering if it is a sensible approach to use `getFileContents`.

fendor:
Closes #671 

Only manual testing done, yet. Not sure how this can be tested, suggestions welcome!

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -73,6 +73,7 @@ import Control.Monad.Extra
 import Control.Monad.Except
 import Control.Monad.Trans.Except
 import Data.Bifunctor                           (first, second)
+import qualified Data.ByteString as BS
 import qualified Data.Text as T
 import           Data.IORef
 import           Data.List.Extra
@@ -280,13 +281,13 @@ atomicFileWrite targetPath write = do
   (tempFilePath, cleanUp) <- newTempFileWithin dir
   (write tempFilePath >> renameFile tempFilePath targetPath) `onException` cleanUp
 
-generateAndWriteHieFile :: HscEnv -> TypecheckedModule -> IO [FileDiagnostic]
-generateAndWriteHieFile hscEnv tcm =
+generateAndWriteHieFile :: HscEnv -> TypecheckedModule -> BS.ByteString -> IO [FileDiagnostic]
+generateAndWriteHieFile hscEnv tcm source =
   handleGenerationErrors dflags "extended interface generation" $ do
     case tm_renamed_source tcm of
       Just rnsrc -> do
         hf <- runHsc hscEnv $
-          GHC.mkHieFile mod_summary (fst $ tm_internals_ tcm) rnsrc ""
+          GHC.mkHieFile mod_summary (fst $ tm_internals_ tcm) rnsrc source
         atomicFileWrite targetPath $ flip GHC.writeHieFile hf
       _ ->
         return ()
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -58,6 +58,7 @@ import Data.IntMap.Strict (IntMap)
 import Data.List
 import qualified Data.Set                                 as Set
 import qualified Data.Text                                as T
+import qualified Data.Text.Encoding                       as T
 import           Development.IDE.GHC.Error
 import           Development.Shake                        hiding (Diagnostic)
 import Development.IDE.Core.RuleTypes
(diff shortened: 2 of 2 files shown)


src/Development/IDE/Core/Rules.hs
```diff
@@ -185,11 +186,24 @@ getHomeHieFile f = do
...
         pm <- use_ GetParsedModule f
-        typeCheckRuleDefinition hsc pm DoGenerateInterfaceFiles
+        source <- getSourceFileSource f
+        typeCheckRuleDefinition hsc pm DoGenerateInterfaceFiles source
       _ <- MaybeT $ liftIO $ timeout 1 wait
       ncu <- mkUpdater
       liftIO $ loadHieFile ncu hie_f
 
+getSourceFileSourceRule :: Rules ()
+getSourceFileSourceRule = define $ \GetSourceFileSource nfp -> do
```
pepeiborra, line 196: What is the motivation to make this a Shake rule, i.e. why should we cache the source file contents? I cannot think of any good reason

src/Development/IDE/Core/Compile.hs
```diff
@@ -276,13 +277,13 @@ atomicFileWrite targetPath write = do
...
-generateAndWriteHieFile :: HscEnv -> TypecheckedModule -> IO [FileDiagnostic]
-generateAndWriteHieFile hscEnv tcm =
+generateAndWriteHieFile :: HscEnv -> TypecheckedModule -> BS.ByteString -> IO [FileDiagnostic]
+generateAndWriteHieFile hscEnv tcm source =
   handleGenerationErrors dflags "extended interface generation" $ do
     case tm_renamed_source tcm of
       Just rnsrc -> do
         hf <- runHsc hscEnv $
-          GHC.mkHieFile mod_summary (fst $ tm_internals_ tcm) rnsrc ""
+          GHC.mkHieFile mod_summary (fst $ tm_internals_ tcm) rnsrc source
```
pepeiborra, line 286: Not relevant but just to point out that you could get the preprocessed source from the `ModSummary` inside the `tm_parsed_module`. However that's not what we want here, since GHC embed the unpreprocessed source in the `hie` files it produces and we want to do the same thing as GHC.

src/Development/IDE/Core/Rules.hs
fendor: It mainly felt right to add it to the Shake Graph. But I do not have any other motivation, I am not really familiar with the shake system, so I am happy to learn :)
I thought it would be always good to cache it.

fendor:
I think there is no real way of knowing the hie file location from outside ghcide. It looks to me like we would need unit-tests that can directly access inner rules to realy test it.

src/Development/IDE/Core/Rules.hs
```diff
@@ -185,11 +186,21 @@ getHomeHieFile f = do
...
       ncu <- mkUpdater
       liftIO $ loadHieFile ncu hie_f
 
+getSourceFileSource :: NormalizedFilePath -> Action BS.ByteString
+getSourceFileSource nfp = do
+    (_, msource) <- getFileContents nfp
+    case msource of
+        Nothing -> do
+            source <- readFile' (fromNormalizedFilePath nfp)
+            let bsSource = T.encodeUtf8 $ T.pack source
```
pepeiborra, line 201: This seems like a very roundabout way of reading a file: String -> Text -> BS. 
Why not use `BS.readFile` like GHC does? 

https://gitlab.haskell.org/ghc/ghc/-/blob/master/compiler/GHC/Iface/Ext/Ast.hs#L284

fendor: Good point, thank you! 

Side question, is there some invariant that this filepath exists in the filesystem? Do we need more exception handling here?

pepeiborra: Hmm, I think existing error handling will deal with this if the file does not exist.

wz1000:
I think a better way to achieve this would be to save the file contents along with the parsed module, so that we can be sure that it is never out of sync.

pepeiborra:
> I think a better way to achieve this would be to save the file contents along with the parsed module, so that we can be sure that it is never out of sync.
> 
> This can be done along with the `getFileContents` rule in the parsed module definition.

The Shake graph already guarantees this property I think

pepeiborra (approved):

PR #707, opened 2020-07-24
Fix Binary instance of Q to handle empty file paths

cocreature:


src/Development/IDE/Core/Shake.hs
```diff
@@ -792,7 +794,14 @@ isBadDependency x
 newtype Q k = Q (k, NormalizedFilePath)
     deriving (Eq,Hashable,NFData, Generic)
 
-instance Binary k => Binary (Q k)
+instance Binary k => Binary (Q k) where
+    put (Q (k, fp)) = put (k, fp)
+    get = do
+        (k, fp) <- get
+        -- The `get` implementation of NormalizedFilePath
```
ndmitchell, line 801: Could we not fix this upstream? Would be good to either have a link to the issue tracking it, or some more detailed explanation of why it's never possible to fix upstream.

cocreature: Upstream doesn’t want to allow empty filepaths since they’re technically not URLs (at least that was the result last time this came up). I did open https://github.com/alanz/haskell-lsp/pull/249 but it doesn’t fix it since the conversion from filepaths to URIs doesn’t work on empty file paths so while it decodes the filepath correctly it cannot convert it.

PR #711, opened 2020-07-27
Codeaction for exporting unused top-level bindings

sureyeaah:
This codeaction lets you export unused top-level functions, data types, newtypes, type synonyms, type classes, type families and pattern synonyms. 
- When exporting data types/newtypes/typeclasses/type families, we export all (e.g. `Foo(..)`).
- We don't support extracting the data constructor/fields when the corresponding type is already exported. e.g.
```haskell
module X(Foo) where
Foo = Foo {foo :: ()}
```
This is done because it would require us to find the corresponding type in the export list.

Fixes https://github.com/haskell/haskell-language-server/issues/157


sureyeaah:
I did that because it was easier to implement but now that I think about it, it would be just as easy to append it at the end (in the same line as the closing bracket). I'll update the PR.

sureyeaah:
I just realized there is an edge case when appending to the export list -`module X(foo,) where` and `module X(foo) where`. We'll have to look for that last comma.

test/exe/Main.hs
```diff
@@ -1657,6 +1658,204 @@ addSigActionTests = let
...
+              [ "{-# OPTIONS_GHC -Wunused-top-binds #-}" 
+              , "module A (foo,bar) where"
+              , "foo = id"
+              , "bar = foo"])
+    , testSession "multi line explicit exports" $ template
+        (T.unlines
+              [ "{-# OPTIONS_GHC -Wunused-top-binds #-}" 
+              , "module A"
+              , "  ("
+              , "    foo) where"
```
cocreature, line 1733: I wonder how clever we should be in terms of inserting line breaks. At some point we venture into autoformatter territory which a code action is definitely not the right place for but maybe there is some middleground. E.g., insert a line break if the opening and closing parentheses are on different lines.

Let’s merge it like this but keep this in mind for a possible future extension.

sureyeaah: When writing codeactions, I prefer to offload as much work as possible to the formatter.
If we were to insert a line break, we would also have to get the indentation right.

PR #713, opened 2020-07-28
Use InitializeParams.rootUri for initial session setup

sureyeaah:
Pass the `config` param to `runLanguageServer` so that we can extract the directory from `InitializeRequest`. 

Fixes https://github.com/digital-asset/ghcide/issues/646

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -96,10 +96,10 @@ main = do
         t <- offsetTime
         hPutStrLn stderr "Starting LSP server..."
         hPutStrLn stderr "If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!"
-        runLanguageServer options (pluginHandler plugins) onInitialConfiguration onConfigurationChange $ \getLspId event vfs caps wProg wIndefProg getConfig -> do
+        runLanguageServer options (pluginHandler plugins) onInitialConfiguration onConfigurationChange $ \getLspId event vfs caps wProg wIndefProg getConfig rootPath -> do
             t <- t
             hPutStrLn stderr $ "Started LSP server in " ++ showDuration t
-            sessionLoader <- loadSession dir
+            sessionLoader <- loadSession $ fromMaybe dir rootPath
             config <- fromMaybe defaultLspConfig <$> getConfig
             let options = (defaultIdeOptions sessionLoader)
                     { optReportProgress = clientSupportsProgress caps
--- a/src/Development/IDE/LSP/LanguageServer.hs
+++ b/src/Development/IDE/LSP/LanguageServer.hs
@@ -46,7 +46,7 @@ runLanguageServer
     -> (InitializeRequest -> Either T.Text config)
     -> (DidChangeConfigurationNotification -> Either T.Text config)
     -> (IO LspId -> (FromServerMessage -> IO ()) -> VFSHandle -> ClientCapabilities
-        -> WithProgressFunc -> WithIndefiniteProgressFunc -> IO (Maybe config) -> IO IdeState)
+        -> WithProgressFunc -> WithIndefiniteProgressFunc -> IO (Maybe config) -> Maybe FilePath -> IO IdeState)
     -> IO ()
 runLanguageServer options userHandlers onInitialConfig onConfigChange getIdeState = do
     -- Move stdout to another file descriptor and duplicate stderr
@@ -133,7 +133,7 @@ runLanguageServer options userHandlers onInitialConfig onConfigChange getIdeStat
         handleInit exitClientMsg clearReqId waitForCancel clientMsgChan lspFuncs@LSP.LspFuncs{..} = do
 
             ide <- getIdeState getNextReqId sendFunc (makeLSPVFSHandle lspFuncs) clientCapabilities
-                               withProgress withIndefiniteProgress config
+                               withProgress withIndefiniteProgress config rootPath
 
             _ <- flip forkFinally (const exitClientMsg) $ forever $ do
                 msg <- readChan clientMsgChan
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -88,6 +88,7 @@ main = do
     , benchmarkTests
     , ifaceTests
     , bootTests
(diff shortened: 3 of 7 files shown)


sureyeaah:
Is there a better way to extract something from the config besides the `InitCBConfig` type class?

src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -131,9 +132,9 @@ runLanguageServer options userHandlers onInitialConfig onConfigChange getIdeStat
     where
         handleInit :: IO () -> (LspId -> IO ()) -> (LspId -> IO ()) -> Chan (Message config) -> LSP.LspFuncs config -> IO (Maybe err)
         handleInit exitClientMsg clearReqId waitForCancel clientMsgChan lspFuncs@LSP.LspFuncs{..} = do
-
+            cfg <- config
             ide <- getIdeState getNextReqId sendFunc (makeLSPVFSHandle lspFuncs) clientCapabilities
-                               withProgress withIndefiniteProgress
+                               withProgress withIndefiniteProgress cfg
```
bubba, line 137: I think the rootUri is available through `LspFuncs` as `rootPath`, which is a bit confusingly admittedly. Can we just pass that here instead? Instead of passing through the config.

exe/Main.hs
```diff
@@ -78,8 +78,11 @@ main = do
     command <- makeLspCommandId "typesignature.add"
 
     let plugins = Completions.plugin <> CodeAction.plugin
-        onInitialConfiguration = const $ Right ()
-        onConfigurationChange  = const $ Right ()
+        
+        -- we use `Maybe InitializeParams` as the config type.
+        -- Change this if you want to parse InitializeRequest/DidChangeConfigurationNotification differently.
+    let onInitialConfiguration RequestMessage{_params=initParams} = Right $ Just initParams
+        onConfigurationChange = const $ Right Nothing
```
bubba, line 85: The `config` type is the settings that users can customise from the language clients, not something internal to the language server. `haskell-lsp` handles this a bit weirdly in that it lets us parse it ourselves from the initialisation message here but it's probably best to not use it as a mechanism for passing about information from the initialization request

bubba:
To test this I would take a look at `runInDir` https://github.com/digital-asset/ghcide/blob/826e886adac32378cb8a9f1ad735ca9234806eab/test/exe/Main.hs#L3043-L3064 and probably make another version that doesn't set the `--cwd` flag and calls `runSession` in a specific directory, and seeing if it can load documents relative to the specified root directory

sureyeaah:
I open ghcide in `dirA` and run the session in the tests with root as `dirB`. Both these directories have different `hie.yaml` files. But regardless of whether I use `InitializeParams.rootUri`, the `dirB/hie.yaml` file is always used.

sureyeaah:
@pepeiborra thanks I'll have a look. I followed @bubba 's advice and retrieved the path from LSPFuncs instead. Should I go back to the parsing?

Also, do you have any idea why my testing approach isn't working?

pepeiborra:
> @pepeiborra thanks I'll have a look. I followed @bubba 's advice and retrieved the path from LSPFuncs instead. Should I go back to the parsing approach?

If you can follow @bubba 's advice with minimal changes, then I believe that should be fine. 

> Also, do you have any idea why my testing approach isn't working?

I believe it doesn't work because ghcide calls [findCradle](https://hackage.haskell.org/package/hie-bios-0.6.1/docs/HIE-Bios.html#v:findCradle) see Main.hs:264

sureyeaah:
Aaha that makes sense. So then what is the change in behaviour that using rootUri should give us (and that i can test for)?

bubba:
@pepeiborra @sureyeaah just to make sure I understand this, the `findCradle` logic is independent of the root directory and only depends on the file being loaded. But `findCradle` I think is really misnamed, since it only finds the `hie.yaml` file. The root directory is used for the implicit cradle instead, as `dir` here:

https://github.com/digital-asset/ghcide/blob/765967d19be4ca64abc999173029a6bfe6aec113/session-loader/Development/IDE/Session.hs#L199-L205

So perhaps a file structure like:

```
- a/
  - Foo.hs
  - a.cabal
- b/
  - Bar.hs
  - b.cabal
```

Launch ghcide in `a/`, but set the rootDir to `../b`. Then try to load Bar.hs and see what happens?

sureyeaah:
Thanks @pepeiborra @bubba! Ready for review.

src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -133,7 +133,7 @@ runLanguageServer options userHandlers onInitialConfig onConfigChange getIdeStat
         handleInit exitClientMsg clearReqId waitForCancel clientMsgChan lspFuncs@LSP.LspFuncs{..} = do
 
             ide <- getIdeState getNextReqId sendFunc (makeLSPVFSHandle lspFuncs) clientCapabilities
-                               withProgress withIndefiniteProgress
+                               withProgress withIndefiniteProgress lspFuncs
```
pepeiborra, line 136: A bunch of the arguments to `getIdeState` are contained inside `lspFuncs`. Please simplify

test/exe/Main.hs
```diff
@@ -2781,9 +2782,22 @@ benchmarkTests =
         , Bench.name e /= "edit" -- the edit experiment does not ever fail
     ]
 
+-- | checks if we use InitializeParams.rootUri for loading session
+rootUriTests :: TestTree
+rootUriTests = testCase "use rootUri" . withoutStackEnv . runTest "dirA" "dirB" $ \dir -> do
```
pepeiborra, line 2787: Does this test fail without the new code?

sureyeaah: Yeah it does.

src/Development/IDE/LSP/LanguageServer.hs
sureyeaah: Didn't realize. Fixed.

pepeiborra:
rebase needed

sureyeaah:
@pepeiborra rebased

PR #720, opened 2020-08-04
Add a test case involving -fno-warn-missing-signatures

zliu41:
I was trying to run ghcide on https://github.com/agda/agda, and the majority of the modules failed. What I observed was
- `Agda.cabal` declares missing signatures as fatal (`-Werror=missing-signatures`)
- `Lexer.hs` (a generated file) has missing signatures, but it disables the warning by `{-# OPTIONS_GHC -fno-warn-missing-signatures #-}`.
- However, ghcide still failed to typecheck `Lexer.hs`, citing missing signatures. `Lexer.hs` is transitively imported by a lot of modules.

This PR fixes it by only enabling non-fatal warnings in `DynFlags`.


wz1000:
Seems reasonable to me, but I think a better way to achieve the same goal would be to strip out `-Werror` entirely. Maybe we can still "upgrade" the `-Werror` diagnostics to errors when presenting them to the user, like we do for type errors.

wz1000:
Closing since this should be fixed by #738. Thanks for bringing attention to this!

pepeiborra:
I agree that #738 is a better solution but the test in this PR is still relevant. 
@zliu41 would you be able to edit the PR down to the tests, and rebase?

pepeiborra (changes requested):
Edit the PR down to the tests

zliu41:
Ok cool, I confirmed that the problem is resolved in HEAD.
Reverted the change and only keeping the test.
Thanks @wz1000 @pepeiborra !

pepeiborra (approved):

PR #721, opened 2020-08-05
Generate doc file URL via LSP (to fix it for Windows)

DunetsNM:
Part of the fix for haskell/haskell-language-server#247

Windows understands only file:/// but not file:// (at least it's true for VS Code)


DunetsNM:
replaced "file://" hardcode with LSP normalizer, this fixes both three slashes for windows and replaces backslashes with forward slash (as mentioned in haskell/vscode-haskell/pull/270) and makes it work for Windows

wz1000 (approved):

pepeiborra:
Thanks you, and apologies for the long delay in merging your PR!

PR #723, opened 2020-08-06
Dynamically load libm on Linux for each new session

bubba:
This fixes the issue on Linux where the binary was statically linked and Template Haskell (or the eval plugin on haskell-language-server) tried to evaluate some code. It would previously fail to load ghc-prim because it couldn't lookup symbols from libm which are usually dynamically linked in.

See https://github.com/haskell/haskell-language-server/issues/221

Unfortunately we can't just statically link in `libm` and have the binary `dlsym` on itself apparently: https://stackoverflow.com/questions/14289488/use-dlsym-on-a-static-binary

So now the statically linked binary actually dynamically links in `libm` on a "best-effort" case: From what I can tell if `libm` is not available this will just print an error to stderr and carry on as usual (albeit it will trip up on Template Haskell), and if the binary dynamically linked then this will have no effect.

cc @pepeiborra

### DIFF
--- a/session-loader/Development/IDE/Session.hs
+++ b/session-loader/Development/IDE/Session.hs
@@ -56,6 +56,7 @@ import System.Info
 import System.IO
 
 import GHC
+import GHCi
 import DynFlags
 import HscTypes
 import Linker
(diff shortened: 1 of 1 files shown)


pepeiborra:
A fair point to make is that currently this is only relevant for HLS, and therefore maybe it belongs in the HLS Main module

bubba:
There was a way I saw originally to load in libm without the `HscEnv` which we could do, I'm not 100% sure what the difference is though. At some point we should also revisit this and see what is the correct way to fix this in the long term

pepeiborra:
Can you rebase or otherwise retrigger CI somehow? I need a green light to merge!

cocreature:
@pepeiborra /azp run should work for anyone with write access to the repo I believe.

cocreature:
hm /azp run probably got lost in the repo transfer. Try it out on new PRs.

PR #725, opened 2020-08-10
Fix haddock to markdown conversion

georgefst:
Previously, `haddockToMarkdown` would make a mess of docs with an identifier inside a monospaced block, as it would attempt to insert backticks twice.

For example, [createDirectoryIfMissing](https://hackage.haskell.org/package/directory-1.3.6.1/docs/src/System.Directory.html#createDirectoryIfMissing) would generate:

```markdown
`\`createDirectoryIfMissing\`  parents dir`  creates a new directory
 `dir`  if it doesn't exist.
```

which renders as:

`\`createDirectoryIfMissing\`  parents dir`  creates a new directory
 `dir`  if it doesn't exist.

This PR corrects that by escaping backticks in all strings upfront, then removing non-escaped backticks inside a monospaced block before surrounding with them.

This is not the cleanest solution. Perhaps we should just rewrite `haddockToMarkdown` to use a custom intermediate data structure, or a library such as `cmark`. Although really, the whole idea of machine-generated markdown makes me wince a little - I assume the LSP protocol doesn't allow us to just send HTML?

Also I should probably add some tests...

### DIFF
--- a/src/Development/IDE/Plugin/Completions.hs
+++ b/src/Development/IDE/Plugin/Completions.hs
@@ -2,7 +2,11 @@
 {-# LANGUAGE TypeFamilies #-}
 #include "ghc-api-version.h"
 
-module Development.IDE.Plugin.Completions(plugin) where
+module Development.IDE.Plugin.Completions
+    (
+      plugin
+    , getCompletionsLSP
+    ) where
 
 import Control.Applicative
 import Language.Haskell.LSP.Messages


georgefst:
> I assume the LSP protocol doesn't allow us to just send HTML?

Actually, microsoft/vscode#97793 enables some usage of HTML tags. Unsure about other editors. Of course, HTML can be embedded in Markdown, so previously they were going out of there way not to support it.

I'll investigate what we might be able to do with that...

georgefst:
FWIW, I've been using this patch for the last week (working with Haskell full time plus some evenings) and not seen any malformed hovers. I'll report back here if I ever do.

wz1000 (approved):

pepeiborra:
Apologies for the long delay in merging this. Can you rebase or otherwise retrigger CI somehow? I don't think I can do it myself (and need a green light to merge)

wz1000:
I'm sorry, I tried to push to your branch after rebasing, but I pushed the ghcide master commit instead.  Can you push https://github.com/wz1000/ghcide/tree/georgefst/master to your master branch and reopen this PR? I don't have permissions to push to your branch anymore, since github decided to close this PR once it was empty.

georgefst:
@wz1000 Done, but AFAICT I don't have any option to reopen either... Shall I just open a fresh PR?

wz1000:
Sure, please do.

PR #727, opened 2020-08-13
Handle multiple user actions concurrently

pepeiborra:
This improves on the single thread pumping user actions right now

### DIFF
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -584,9 +584,9 @@ newSession ShakeExtras{..} shakeDb acts = do
     let
         -- A daemon-like action used to inject additional work
         -- Runs actions from the work queue sequentially
-        pumpActionThread = forever $ do
+        pumpActionThread = do
             d <- liftIO $ atomically $ popQueue actionQueue
-            run d
+            void $ parallel [run d, pumpActionThread]
 
         run d  = do
             start <- liftIO offsetTime
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -533,6 +533,7 @@ data DelayedAction a = DelayedAction
   , actionPriority :: Logger.Priority -- ^ Priority with which to log the action
   , getAction :: Action a -- ^ The payload
   }
+  deriving Functor
 
 type DelayedActionInternal = DelayedAction ()
 
@@ -549,9 +550,8 @@ delayedAction a = do
 
 -- | Restart the current 'ShakeSession' with the given system actions.
 --   Any computation running in the current session will be aborted,
---   but user actions (added via 'shakeEnqueue') will be requeued.
---   Progress is reported only on the system actions.
-shakeRestart :: IdeState -> [DelayedAction a] -> IO ()
+--   but actions added via 'shakeEnqueue' will be requeued.
+shakeRestart :: IdeState -> [DelayedAction ()] -> IO ()
 shakeRestart IdeState{..} systemActs =
     withMVar'
         shakeSession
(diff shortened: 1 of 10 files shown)


wz1000:
This looks suspiciously similar to some of @mpickering's code I deleted from the stale info PR :)

What about your patch to handle requests concurrently? I've been feeling lack of that lately.

pepeiborra:
> This looks suspiciously similar to some of @mpickering's code I deleted from the stale info PR :)

Did he implement this too? Great minds think alike. Please share the link so that I can check if anything's missing.
 
> What about your patch to handle requests concurrently? I've been feeling lack of that lately.

I haven't looked into that yet.

src/Development/IDE/Types/Action.hs
```diff
@@ -0,0 +1,69 @@
+module Development.IDE.Types.Action
+  (DelayedAction(..)
+  ,DelayedActionInternal
+  ,ActionQueue
+  ,newQueue
+  ,pushQueue
+  ,popQueue
+  ,doneQueue
+  ,peekInProgress
```
rayshih, line 9: `  , peekInProgress` 

space

src/Development/IDE/Types/Action.hs
```diff
@@ -0,0 +1,69 @@
...
+    newActions <- newTQueue
+    inProgress <- newTVar mempty
+    return ActionQueue{..}
+
+pushQueue :: DelayedActionInternal -> ActionQueue -> STM ()
+pushQueue act ActionQueue{..} = writeTQueue newActions act
+
+-- | You must call 'doneQueue' to signal completion
+popQueue :: ActionQueue -> STM DelayedActionInternal
+popQueue ActionQueue{..} = do
```
rayshih, line 58: pop sounds like dequeue but it's still in `ActionQueue`

pepeiborra: pop is intended to mean dequeue. The item is in the inProgress list, but no longer in the queue. You cannot pop the same element twice.

src/Development/IDE/Types/Action.hs
pepeiborra: I am layout and spacing blind! Let me run ormolu

src/Development/IDE/Core/Shake.hs
```diff
@@ -565,86 +554,74 @@ shakeRestart IdeState{..} systemActs =
...
-                    return (logDelayedAction logger act)
-                liftIO $ atomically $ writeTVar actionInProgress Nothing
+        pumpActionThread = do
+            d <- liftIO $ atomically $ popQueue actionQueue
+            void $ parallel [run d, pumpActionThread]
+
+        run d  = do
+            start <- liftIO offsetTime
+            getAction d
+            liftIO $ atomically $ doneQueue d actionQueue
```
wz1000, line 590: There can be an exception after `getAction d` but before `doneQueue`. Should the `popQueue`, `getAction` and `doneQueue` calls be wrapped in `bracket`?

wz1000: Ok, so `bracket` won't be appropriate since we need to requeue, but perhaps we can `mask` exceptions everywhere but in `getAction d`.

pepeiborra: I'm tempted to leave as it is. The outcome of not calling doneQueue is just redundant work: the action will be reexecuted in the next restart, then the check in line 630 will not signal the barrier, and finally the action will be removed from the queue

wz1000: OK, makes sense.

wz1000 (approved):

Issue #710, opened 2020-07-27
bug in "Delete top-binding" codeaction


sureyeaah:
Consider the following:

```haskell
{-# OPTIONS_GHC -Wunused-top-binds #-}
{-# OPTIONS_GHC -Wunused-binds #-}
module X (foo, bar) where

foo = id
  where
    bar = ()
//  ^^^
bar = id
```
1. The codeaction is offered for `bar` (it should not be!).
2. It deletes the top-level `bar` (it should delete the local `bar` instead).

see https://github.com/digital-asset/ghcide/pull/711 for fix to same problem in different codeaction.

rayshih:
Just created a WIP commit for this: https://github.com/rayshih/ghcide/pull/1

Will polish that and recreate a real PR

pepeiborra:
Think we can close this now

PR #728, opened 2020-08-13
Fix issue #710: fix suggest delete binding 

rayshih:
By generalizing code action: suggestions of delete unused **top** binding to suggestions of delete unused binding.

This is basically a reimplementation which traverses the AST.

suggest delete unused binding


pepeiborra:
Merged, thank you Ray!

PR #731, opened 2020-08-16
Store the lsp client settings in shakeExtras and create a Rule to get them

jneira:
* The motivation is to expose the client settings in shake rules, concretely my goal is to enable or disable [hlint diagnostics](https://github.com/haskell/haskell-language-server/pull/166) using a client setting, changing diagnostics on the fly if the user change it in the editor
  * I hope being able to configure the rules (when parsing or typechecking for example?) using the client settings could be useful in general
*  The config is stored in the json raw format and updated using the `InitializeParams` and the `didChangeConfigurationParamsHandler` notification
  * The first one is empty using vscode, it notifies a configuration change when the lsp server is started. I guess other editors will use the first one (???). I assume both should update the client settings cause the `onInitialConfig` and `onConfigChange` callbacks returns the same type (it is a type param in ghcide)
* The client settings has an specific rule, which uses `alwaysRerun`, inspired in #688. There @pepeiborra advised it can degrade performance, but i guess there will be not much configuration changes.
* With this patch a Rule that depends on `GetClientSettings` (for example the hlint rule that produces hlint diagnostics) is rerun on a file modification or opening a new one. Ideally i would like to trigger it within the configuration change itself (maybe restarting the shake session?)

Sorry i am missing something obvious in the ghcide or shake concepts, this is my first pr touching their internals.

### DIFF
--- a/src/Development/IDE/Core/IdeConfiguration.hs
+++ b/src/Development/IDE/Core/IdeConfiguration.hs
@@ -2,6 +2,7 @@
 module Development.IDE.Core.IdeConfiguration
   ( IdeConfiguration(..)
   , registerIdeConfiguration
+  , getIdeConfiguration
   , parseConfiguration
   , parseWorkspaceFolder
   , isWorkspaceFile
@@ -61,9 +62,9 @@ modifyWorkspaceFolders ide f = modifyIdeConfiguration ide f'
   where f' (IdeConfiguration ws initOpts) = IdeConfiguration (f ws) initOpts
 
 modifyClientSettings
-  :: IdeState -> (Value -> Value) -> IO ()
+  :: IdeState -> (Maybe Value -> Maybe Value) -> IO ()
 modifyClientSettings ide f = modifyIdeConfiguration ide f'
-  where f' (IdeConfiguration ws clientSettings) = IdeConfiguration ws (f <$> clientSettings)
+  where f' (IdeConfiguration ws clientSettings) = IdeConfiguration ws (f clientSettings)
 
 modifyIdeConfiguration
   :: IdeState -> (IdeConfiguration -> IdeConfiguration) -> IO ()
--- a/src/Development/IDE/Core/IdeConfiguration.hs
+++ b/src/Development/IDE/Core/IdeConfiguration.hs
@@ -14,6 +14,7 @@ where
 
 import           Control.Concurrent.Extra
 import           Control.Monad
+import           Data.Hashable                  (Hashed, hashed, unhashed)
 import           Data.HashSet                   (HashSet, singleton)
 import           Data.Text                      (Text, isPrefixOf)
 import           Data.Aeson.Types               (Value)
@@ -26,7 +27,7 @@ import           System.FilePath (isRelative)
 -- | Lsp client relevant configuration details
 data IdeConfiguration = IdeConfiguration
   { workspaceFolders :: HashSet NormalizedUri
-  , clientSettings :: Maybe Value
+  , clientSettings :: Hashed (Maybe Value)
   }
   deriving (Show)
 
@@ -51,7 +52,7 @@ parseConfiguration InitializeParams {..} =
       <> (foldMap . foldMap)
            (singleton . parseWorkspaceFolder)
            _workspaceFolders
-  clientSettings = _initializationOptions
+  clientSettings = hashed _initializationOptions
 
 parseWorkspaceFolder :: WorkspaceFolder -> NormalizedUri
 parseWorkspaceFolder =
@@ -65,7 +66,8 @@ modifyWorkspaceFolders ide f = modifyIdeConfiguration ide f'
 modifyClientSettings
   :: IdeState -> (Maybe Value -> Maybe Value) -> IO ()
 modifyClientSettings ide f = modifyIdeConfiguration ide f'
-  where f' (IdeConfiguration ws clientSettings) = IdeConfiguration ws (f clientSettings)
+  where f' (IdeConfiguration ws clientSettings) =
+            IdeConfiguration ws (hashed . f . unhashed $ clientSettings)
 
 modifyIdeConfiguration
   :: IdeState -> (IdeConfiguration -> IdeConfiguration) -> IO ()
@@ -86,4 +88,4 @@ isWorkspaceFile file =
           workspaceFolders
 
 getClientSettings :: Action (Maybe Value)
-getClientSettings = clientSettings <$> getIdeConfiguration
+getClientSettings = unhashed . clientSettings <$> getIdeConfiguration
--- a/src/Development/IDE/Core/IdeConfiguration.hs
+++ b/src/Development/IDE/Core/IdeConfiguration.hs
@@ -8,6 +8,7 @@ module Development.IDE.Core.IdeConfiguration
   , isWorkspaceFile
   , modifyWorkspaceFolders
   , modifyClientSettings
+  , getClientSettings
   )
 where
 
@@ -84,3 +85,6 @@ isWorkspaceFile file =
         any
           (\root -> toText root `isPrefixOf` toText (filePathToUri' file))
           workspaceFolders
+
+getClientSettings :: Action (Maybe Value)
+getClientSettings = clientSettings <$> getIdeConfiguration 
--- a/src/Development/IDE/Core/IdeConfiguration.hs
+++ b/src/Development/IDE/Core/IdeConfiguration.hs
@@ -6,6 +6,7 @@ module Development.IDE.Core.IdeConfiguration
   , parseWorkspaceFolder
   , isWorkspaceFile
   , modifyWorkspaceFolders
+  , modifyClientSettings
   )
 where
 
@@ -13,6 +14,7 @@ import           Control.Concurrent.Extra
 import           Control.Monad
 import           Data.HashSet                   (HashSet, singleton)
 import           Data.Text                      (Text, isPrefixOf)
+import           Data.Aeson.Types               (Value)
 import           Development.IDE.Core.Shake
 import           Development.IDE.Types.Location
 import           Development.Shake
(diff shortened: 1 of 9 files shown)


src/Development/IDE/Core/IdeConfiguration.hs
```diff
@@ -39,24 +44,35 @@ getIdeConfiguration =
...
+  where f' (IdeConfiguration ws clientSettings) = IdeConfiguration ws (f clientSettings)
+
+modifyIdeConfiguration
+  :: IdeState -> (IdeConfiguration -> IdeConfiguration) -> IO ()
+modifyIdeConfiguration ide f = do
   IdeConfigurationVar var <- getIdeGlobalState ide
-  IdeConfiguration    ws  <- readVar var
-  writeVar var (IdeConfiguration (f ws))
+  ideConfig <- readVar var
+  writeVar var (f ideConfig)
```
pepeiborra, line 75: Let's take the chance to fix the potential race condition in this code by using `modifyVar`

src/Development/IDE/Core/RuleTypes.hs
```diff
@@ -221,3 +222,11 @@ data GetModSummary = GetModSummary
...
 instance NFData   GetModSummary
 instance Binary   GetModSummary
+
+data GetClientSettings = GetClientSettings
+    deriving (Eq, Show, Typeable, Generic)
+instance Hashable GetClientSettings
+instance NFData   GetClientSettings
+instance Binary   GetClientSettings
+
+type instance RuleResult GetClientSettings = Maybe Value
```
pepeiborra, line 232: Please add a comment describing the ClientSettings and what they are used for

src/Development/IDE/Core/Rules.hs
```diff
@@ -784,6 +785,12 @@ isFileOfInterestRule = defineEarlyCutoff $ \IsFileOfInterest f -> do
     let res = f `elem` filesOfInterest
     return (Just (if res then "1" else ""), ([], Just res))
 
+
+getClientSettingsRule :: Rules ()
+getClientSettingsRule = defineNoFile $ \GetClientSettings -> do
+  alwaysRerun
```
pepeiborra, line 791: An `alwaysRerun` rule needs early cutoff - please use `defineEarlyCutoffNoFile`. 
Moreover, following my advice in #688, please change the `RuleResult GetClientSettings` to `Hashed (Maybe Value)`

jneira:
@pepeiborra many thanks for your comments and insights, i was a little bit lost in the ghcide codebase.

src/Development/IDE/Core/Rules.hs
pepeiborra: This isn't fully resolved, please change the RuleResult GetClientSettings to Hashed (Maybe Value)

jneira: yeah, I was about to change it in another commit, is it for not recomputing the hash every time or is there another reason?

pepeiborra: that's right

jneira: I would like to keep `clientSettings IdeConfiguration` with type `Maybe Value`, to not leak the performance related wrapper to client code, would be it possible? (i guess it is not, after taking a look to #688)

jneira: nvm, i am happy just keeping `getClientSettings` and `modifyClientSettings` with the same signature, without `Hashed`

jneira:
> Could we have some tests to at least check that the Shake session is restarted when the config does change?

I'll try to add it asap. Only comment that [i've cherrypicked changes here in hls](https://github.com/haskell/haskell-language-server/pull/166/commits/86f07725f7cd73210520b64843ef5daaa398285d) and now hlint suggestions honour happily the client setting 😄

pepeiborra:
needs rebase (and tests!)

jneira:
rebased! that was easy, implement the tests will take me some time thought cause i am not used to ghcode test suite, any tip will be welcomed 😄

pepeiborra:
The test should do something along these lines:

1. Open a file
2. Wait for progress
3. Change the lsp client settings
4. Observe the session did restart. This is the tricky bit, it might be observable by waiting for progress. I recommend uncommenting the lines 3323-3327 in test/Main.hs so that you can see the LSP traffic

jneira:
Yeah, i was thinking in add a test in hls with hlint, cause there it is easy to observe how hlints diagnostics come and go.

pepeiborra:
I think we want to have a test here as well please. You can send custom messages to the lsp client if necessary, see how cradle loading is tested

jneira:
I cant make progress with the tests cause the test suite hangs in windows. hls one too so i suspect some change in lsp-test around the shutdown of the process is involved. 😟

jneira:
After fixing the hang of the test suite in windows and add some logging notifications in the Shake module i have been able to add the test cases.

wz1000:
rebase needed

jneira:
There is a failing test for ghc-8.6 and 8.10 (but no for 8.8 🤔 )
```
2020-09-20T21:57:49.0532485Z   Interface loading tests
2020-09-20T21:57:49.2935216Z     iface-error-test-1:                                                                                   FAIL
2020-09-20T21:57:49.2959459Z       Exception: NamedParserException "NotificationMessage ServerMethod PublishDiagnosticsParams" (NamedParserException "NotificationMessage ServerMethod PublishDiagnosticsParams" (NamedParserException "NotificationMessage ServerMethod PublishDiagnosticsParams" (NamedParserException"
.............. (NamedParserException "Response for id: IdInt 1" (NamedParserException "NotificationMessage ServerMethod (ProgressParams WorkDoneProgressBeginParams)" (BothFailed (Unexpected "ConduitParser.empty") (Unexpected "ConduitParser.empty"))))))))))))))))))))))))))))) (Unexpected "ConduitParser.empty")))))))))))))))))))))))))))))
```

I guess the new log notifications for shake session broke it. 😟

src/Development/IDE/Core/Shake.hs
```diff
@@ -567,16 +567,22 @@ shakeRestart IdeState{..} acts =
               let profile = case res of
                       Just fp -> ", profile saved at " <> fp
                       _ -> ""
-              logDebug (logger shakeExtras) $ T.pack $
-                  "Restarting build session (aborting the previous one took " ++
-                  showDuration stopTime ++ profile ++ ")"
+              let msg = T.pack $ "Restarting build session (aborting the previous one took "
+                              ++ showDuration stopTime ++ profile ++ ")"
+              logDebug (logger shakeExtras) msg
+              eventer shakeExtras (notifyLogMessage msg)
```
pepeiborra, line 573: we should probably only send this notification when `--test`

If the goal is to observe session restarts, shouldn't this notification be sent from `newSession`?

https://github.com/haskell/ghcide/blob/20ce9d3a46e92b57ab3e559d20d8f6b4686c8bfa/src/Development/IDE/Core/Shake.hs#L605

src/Development/IDE/Core/Shake.hs
```diff
@@ -616,18 +622,22 @@ newSession ShakeExtras{..} shakeDb acts = do
             getAction d
             liftIO $ atomically $ doneQueue d actionQueue
             runTime <- liftIO start
-            liftIO $ logPriority logger (actionPriority d) $ T.pack $
-                "finish: " ++ actionName d ++ " (took " ++ showDuration runTime ++ ")"
+            let msg = T.pack $ "finish: " ++ actionName d
+                            ++ " (took " ++ showDuration runTime ++ ")"
+            liftIO $ do
+                logPriority logger (actionPriority d) msg
+                eventer $ notifyLogMessage msg
```
pepeiborra, line 629: only when `--test`

src/Development/IDE/Core/Shake.hs
```diff
@@ -616,18 +622,22 @@ newSession ShakeExtras{..} shakeDb acts = do
...
           let res' = case res of
                       Left e -> "exception: " <> displayException e
                       Right _ -> "completed"
-
-          let wrapUp = logDebug logger $ T.pack $ "Finishing build session(" ++ res' ++ ")"
-          return wrapUp
+          let msg = T.pack $ "Finishing build session(" ++ res' ++ ")"
+          return $ do
+              logDebug logger msg
+              eventer $ notifyLogMessage msg
```
pepeiborra, line 640: only when `--test`

pepeiborra (commented):
> There is a failing test for ghc-8.6 and 8.10 (but no for 8.8 🤔 )
> 
> ```
> 2020-09-20T21:57:49.0532485Z   Interface loading tests
> 2020-09-20T21:57:49.2935216Z     iface-error-test-1:                                                                                   FAIL
> 2020-09-20T21:57:49.2959459Z       Exception: NamedParserException "NotificationMessage ServerMethod PublishDiagnosticsParams" (NamedParserException "NotificationMessage ServerMethod PublishDiagnosticsParams" (NamedParserException "NotificationMessage ServerMethod PublishDiagnosticsParams" (NamedParserException"
> .............. (NamedParserException "Response for id: IdInt 1" (NamedParserException "NotificationMessage ServerMethod (ProgressParams WorkDoneProgressBeginParams)" (BothFailed (Unexpected "ConduitParser.empty") (Unexpected "ConduitParser.empty"))))))))))))))))))))))))))))) (Unexpected "ConduitParser.empty")))))))))))))))))))))))))))))
> ```
> 
> I guess the new log notifications for shake session broke it. 😟

You can probably simplify the failing test as below, which might help it pass:

```
    res <- skipManyTill anyMessage $ responseForId li
```

src/Development/IDE/Core/Shake.hs
jneira: Mmm,  i have followed closely the existing log messages, although i did not for "finishing" ones.
The "Restarting build session..." message seemed the right one to know if, well, the session was being restarted. 🤔

pepeiborra: ignore me, the notification is indeed in the right place

jneira:
Really tired of fpcomplete timeouts 😢

jneira:
:crossed_fingers:

jneira:
finally ci is green! @pepeiborra i think last requested changes are done, thanks for your patience :smile:

pepeiborra:
Eureka!

pepeiborra (approved):

PR #734, opened 2020-08-20
Use Bios targets in FindImports

### LINKED ISSUE #733: Use hie-bios targets for locating modules
With modern hie-bios versions, we already know where every module in the known component set lives. #688 already puts this knowledge in the Shake graph. 

On the other hand, the code for locating a module considers all the possible locations, as given by the `-i` flags passed to the cradle. If we could replace that search by a simple lookup in the known module set, I think we would no longer need the file system watchers.

https://github.com/digital-asset/ghcide/blob/765967d19be4ca64abc999173029a6bfe6aec113/src/Development/IDE/Import/FindImports.hs#L113-L118

pepeiborra:
Do not return any module that doesn't belong to the targets set.
If the targets set is empty, then lift this restriction.

The file existence check is probably redundant but keeping it around for now.

Fixes #733

### DIFF
--- a/session-loader/Development/IDE/Session.hs
+++ b/session-loader/Development/IDE/Session.hs
@@ -58,6 +58,7 @@ import Linker
 import Module
 import NameCache
 import Packages
+import qualified Data.HashSet as Set
 
 -- | Given a root directory, return a Shake 'Action' which setups an
 -- 'IdeGhcSession' given a file.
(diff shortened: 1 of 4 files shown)


pepeiborra:
Using the targets of the `HscEnvEq` doesn't work because they don't span across the whole component set. I'll rebase on top of #688 and use `KnownFiles` instead.

pepeiborra:
https://github.com/wz1000/ghcide/pull/10

PR #741, opened 2020-08-23
Restore identifiers missing from hi file

### LINKED ISSUE #614: interface-file errors when using Template Haskell since ghcide 0.2.0
I have a project that compiles with cabal and often works without a problem with ghcide.
But sometimes I see issues like this:

```
 [typecheck] [E] • Can't find interface-file declaration for data constructor SortPosition
    Probable cause: bug in .hi-boot file, or inconsistent .hi file
    Use -ddump-if-trace to get an idea of which file caused the error
```

I have seen this at different positions in my code, but sadly, also because of the heisenbugishness of this bug, I couldn‘t construct a minimal failing example yet.

This issue didn‘t happen on ghcide 0.1.0.

On thing of note: I have only met this issue on data constructors that where called like the Type they inhabit. Renaming the Type fixed the problem for me. But that might just have been luck.

**EDIT:** The Type is generated via Template Haskell.

maralorn:
This fixes part of #614 by introducing a workaround for ghc droping some bindings that we still need.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -130,13 +130,14 @@ typecheckModule (IdeDefer defer) hsc pm = do
             dflags = ms_hspp_opts modSummary
 
         modSummary' <- initPlugins modSummary
-        (warnings, tcm) <- withWarnings "typecheck" $ \tweak ->
+        (warnings, tcm1) <- withWarnings "typecheck" $ \tweak ->
             GHC.typecheckModule $ enableTopLevelWarnings
                                 $ demoteIfDefer pm{pm_mod_summary = tweak modSummary'}
+        tcm2 <- liftIO $ fixDetailsForTH tcm1
         let errorPipeline = unDefer . hideDiag dflags
             diags = map errorPipeline warnings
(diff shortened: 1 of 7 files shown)


pepeiborra (approved):
Awesome, thank you! 

Looks like some compat. changes are needed for GHC <8.8. I would suggest moving `fixDetailsForTH` to `Development.IDE.GHC.Compat` and use CPP to allow for multiple implementations. You can choose to use `return` for the <8.8 implementation if it turns out to be too hard to fix.

maralorn:
I am sorry, that I have not looked into this further. I realized that this only one and not _the_ TH issues I have. I will address the fixes, when I find the time.

pepeiborra:
@maralorn which is the TH issue *you* have?

maralorn:
Well that's what I reported at #614. I have a failing testcase linked there.
It is honestly not a blocker for this PR. Because this definitely fixes a bug.

wz1000:
@maralorn can you add your failing test case here with "known broken"?

maralorn:
(I will rebase the PR when it's declared ready to merge.)

maralorn:
All checks green. \o/

wz1000:
Thanks!

PR #743, opened 2020-08-27
Suggest open imports

expipiplus1:
Fixes https://github.com/haskell/haskell-language-server/issues/353


expipiplus1:
Duplicate of https://github.com/digital-asset/ghcide/pull/740

PR #746, opened 2020-08-30
Include module path in progress message

pepeiborra:
We use a multi-project bios cradle at work, and currently the progress message is always the same "Setting up project fbcode", which isn't very helpful. This simple change will help at least determine what prompted the cradle call


ndmitchell (approved):
More information here looks like a great idea. Although I'm curious why project is always the same, and not a more refined message?

pepeiborra:
> More information here looks like a great idea. Although I'm curious why project is always the same, and not a more refined message?

There isn't enough information to craft a better message really

PR #748, opened 2020-08-31
Refinement holes

### LINKED ISSUE #639: Improved fill hole suggestions
[A lot of work](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/glasgow_exts.html#typed-holes) has gone into making GHC able to produce suggestions and the current code action falls short of making the most of it. The possibilities include:

- Should it use `-fshow-hole-constraints` to make added constraints explicit?
- Should it use `-fshow-docs-of-hole-fits` to show the docs of the suggestion?
- Should it use `-frefinement-level-hole-fits=1` to show almost-valid fits?
- Should it use `-fsort-by-subsumption-hole-fits` ?

pepeiborra:
Fixes #639

### DIFF
--- a/src/Development/IDE/GHC/Compat.hs
+++ b/src/Development/IDE/GHC/Compat.hs
@@ -330,7 +330,9 @@ setUpTypedHoles ::DynFlags -> DynFlags
 #if MIN_GHC_API_VERSION(8,6,0)
 setUpTypedHoles df
   = flip gopt_unset Opt_AbstractRefHoleFits    -- too spammy
+#if MIN_GHC_API_VERSION(8,8,0)
   $ flip gopt_unset Opt_ShowDocsOfHoleFits     -- not used
+#endif
   $ flip gopt_unset Opt_ShowMatchesOfHoleFits  -- nice but broken (forgets module qualifiers)
   $ flip gopt_unset Opt_ShowProvOfHoleFits     -- not used
   $ flip gopt_unset Opt_ShowTypeAppOfHoleFits  -- not used
--- a/src/Development/IDE/GHC/Compat.hs
+++ b/src/Development/IDE/GHC/Compat.hs
@@ -327,12 +327,27 @@ dontWriteHieFiles d =
 #endif
 
 setUpTypedHoles ::DynFlags -> DynFlags
-setUpTypedHoles df
-  = df
 #if MIN_GHC_API_VERSION(8,6,0)
-  { refLevelHoleFits = Just 2  -- GHC diagnostics become too slow for level > 2
+setUpTypedHoles df
+  = flip gopt_unset Opt_AbstractRefHoleFits    -- too spammy
+  $ flip gopt_unset Opt_ShowDocsOfHoleFits     -- not used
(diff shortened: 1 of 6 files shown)


wz1000 (approved):

bubba (approved):

pepeiborra:
This change needs a bit more work, as I found some more complex examples of refinement hole fits are not handled by the new regex approach

pepeiborra:
My second pass is ready for re-review:

- Set a number of GHC flags that make parsing the diagnostics much simpler, ensure consistent behaviour and improve sorting
- Simplified the parsing code
- Sorted the code actions by priority, to prevent long lists of typed holes from burying other suggestions
- Add a hook to customise the DynFlags in case we want to make the typed hole settings available to the user. Refinement holes are very powerful but make type checking an order of magnitude slower.

wz1000:
Is this OK to merge?

pepeiborra:
Yes, good to merge if changes requested after re-review

src/Development/IDE/Types/Options.hs
```diff
@@ -94,6 +94,9 @@ data IdeOptions = IdeOptions
     --   Otherwise, return the result of parsing without Opt_Haddock, so
     --   that the parsed module contains the result of Opt_KeepRawTokenStream,
     --   which might be necessary for hlint.
+  , optCustomDynFlags :: DynFlags -> DynFlags
```
wz1000, line 97: Is this for hls and other downstream users of ghcide?

pepeiborra: yep

src/Development/IDE/GHC/Compat.hs
```diff
@@ -325,6 +326,31 @@ dontWriteHieFiles d =
...
+  = flip gopt_unset Opt_AbstractRefHoleFits    -- too spammy
+#if MIN_GHC_API_VERSION(8,8,0)
+  $ flip gopt_unset Opt_ShowDocsOfHoleFits     -- not used
+#endif
+  $ flip gopt_unset Opt_ShowMatchesOfHoleFits  -- nice but broken (forgets module qualifiers)
+  $ flip gopt_unset Opt_ShowProvOfHoleFits     -- not used
+  $ flip gopt_unset Opt_ShowTypeAppOfHoleFits  -- not used
+  $ flip gopt_unset Opt_ShowTypeAppVarsOfHoleFits -- not used
+  $ flip gopt_unset Opt_ShowTypeOfHoleFits     -- massively simplifies parsing
+  $ flip gopt_set   Opt_SortBySubsumHoleFits   -- very nice and fast enough in most cases
```
Tritlo, line 341: This line slows the whole thing by a lot, since it does pairwise comparison between every single possible fit, and everything grinds to a halt. I'd *strongly* recommend you turn it off!

pepeiborra: You mean sort by subsumption right? It wasn't too bad in my empirical tests. It's quadratic complexity but the number of items is small, so only in degenerate cases like `_ :: a` does it grind to a halt, and ghcide will cancel it as soon as an edit is made anyway.

Overall, there is no set of flags that will make everyone happy, so we should eventually expose this flags as a user configurable setting.

PR #749, opened 2020-08-31
Minimal nix-shell script

pepeiborra:
Ideally we should have a Nix CI job too, but I haven't got around to that yet.


shell.nix
```diff
@@ -0,0 +1,107 @@
...
+          time
+          transformers
+          typed-process
+          unordered-containers
+          utf8-string
+          yaml
+         ]);
+in
+stdenv.mkDerivation {
+  name = "haskell-language-server";
```
adamse, line 88: ghcide?

shell.nix
```diff
@@ -0,0 +1,107 @@
...
+
+{ sources ? import nix/sources.nix,
+  nixpkgs ? import sources.nixpkgs {},
+  compiler ? "default",
+  hoogle ? false
+ }:
+with nixpkgs;
+
+let defaultCompiler = "ghc" + lib.replaceStrings ["."] [""] haskellPackages.ghc.version;
+    haskellPackagesForProject = p:
```
adamse, line 26: this bit was a bit confusing, but I see why. Otoh, why not let nix build the dependecies?

adamse (approved):
I tested it: `nix-shell` and then `cabal update && cabal build exe:ghcide`.

It seems to work.

(I could also load the ghcide source with the build product)

adamse (commented):
xxx

shell.nix
pepeiborra: We let Nix download the dependencies! We don't want Nix to build them, it's slow and often wrong, since it doesn't do constraint solving.

shell.nix
pepeiborra: Thanks!

wz1000 (approved):

mpickering:
It might get a bit annoying having to manually update the list of haskell dependencies.. the shell.nix file which I use locally is quite a lot like this but doesn't provide a compiler with any packages so cabal builds the Haskel dependencies.

PR #750, opened 2020-09-01
CI: remove (internal) DA Slack notifications

garyverhaegen-da:
With the repo now officially transferred from Digital Asset's control to the open-source Haskell organization, there is no good reason for master builds to ping DA anymore. This (the corresponding Slack token "secret") is also the only piece of non-open-source configuration in the existing build process, so it needs to be removed before the CI can be transferred.


ndmitchell (approved):

cocreature (approved):
Thanks!

PR #751, opened 2020-09-02
GHC 8.8.4 & 8.10.2

domenkozar:


pepeiborra:
Thanks! Did you investigate if any extra deps can be removed?

pepeiborra (approved):

pepeiborra:
CI rerun seems broken - could you rebase or push an empty commit to force a rerun please?

domenkozar:
Might be related to https://www.stackage.org/blog/2020/08/ghc-8-10-2-windows-workaround

domenkozar:
Yup: `zlib             > ghc.exe: could not execute: C:/GitLabRunner/builds/2WeHDSFP/0/ghc/ghc/inplace/mingw/bin/ld.exe`

pepeiborra:
ping @jneira help!

jneira:
ugh, that workaround would be doable with some `sed` magic over the ghc config file but i think the ghc team ia gonna release a new version 8.10.3 with the fix so maybe the easy way is to exclude 8.10 from ci temporary:

https://github.com/haskell/ghcide/blob/84af588fa00717ad5f624dd600bfd0b0c6060090/.azure/windows-stack.yml#L8-L9

... or keep a stack8101.yaml and use it for windows until 8.10.3 is released (more user firendly)

pepeiborra:
I'm happy with a temporary exclusion @jneira, why don't you send a separate PR to do that? 
Or alternatively @domenkozar you can make the change in this PR.

jneira:
Yeah, will do but i would prefer to keep a stack yaml for ghc-8.10.1

domenkozar:
Hopefully this will pass now :)

pepeiborra:
Merged, thanks!

PR #760, opened 2020-09-04
Write ifaces on save

wz1000:
To achieve this we augment the FileOfInterest store with a flag that indicates whether a file is saved on disk or not.

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -46,7 +46,7 @@ import System.FilePath
 import System.Time.Extra
 import Paths_ghcide
 import Development.GitRev
-import qualified Data.HashSet as HashSet
+import qualified Data.HashMap.Strict as HashMap
 import qualified Data.Aeson as J
 
 import HIE.Bios.Cradle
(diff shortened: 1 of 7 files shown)


wz1000:
The `mtime` change in the last commit isn't very relevant right now, but it is important for checking the integrity of the database in my hiedb branch.

pepeiborra (approved):
Changes look good to me. Can we have a test too?

wz1000:
I'm not sure how to test this change.

pepeiborra:
Hmmm I cannot think of a nice way either. 

Why do we want to write interface files on save? Is that for hiedb?

wz1000:
Yes, it is useful for hiedb, but also addresses some of your concerns here: https://github.com/haskell/ghcide/pull/688#discussion_r453326292

wz1000:
A way to test this might be to check if the interface file exists after saving. But I don't think we can get the path the the ghcide cache directory currently.

Maybe we should also introduce unit tests?

pepeiborra:
I was thinking that maybe it would become easier to test this when the hiedb changes land

pepeiborra:
@wz1000 are you planning to add tests to this? You can know find the location of the cache directory for a given module

wz1000:
I think this is ready

PR #763, opened 2020-09-04
Use hie-bios-0.7.1

jneira:
* To ensure we are able to set the stack.yaml in hie.yaml stack config

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -86,7 +86,7 @@ library
         ghc-check >=0.5.0.1,
         ghc-paths,
         cryptohash-sha1 >=0.11.100 && <0.12,
-        hie-bios >= 0.6.0 && < 0.8.0,
+        hie-bios >= 0.7.1 && < 0.8.0,
         base16-bytestring >=0.1.1 && <0.2
     if os(windows)
       build-depends:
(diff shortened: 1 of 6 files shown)


ghcide.cabal
```diff
@@ -276,7 +276,7 @@ executable ghcide
         hashable,
         haskell-lsp,
         haskell-lsp-types,
-        hie-bios >= 0.6.0 && < 0.8,
+        hie-bios >= 0.7.1 && < 0.8,
```
fendor, line 279: We should probably remove those as they are redundant

fendor (approved):
LGTM

ghcide.cabal
jneira, line 279:
```suggestion
        hie-bios,
```

jneira: Yeah, they are already in the lib

runeksvendsen:
Looks like `hie-bios` specifies a version bound of `>= 0.8.32` for `yaml` while depending on its `Data.Yaml.Internal` module exporting `Warning(..)` but this requires at least `0.10.0`.

runeksvendsen:
I've reported this as a `hie-bios` bug: https://github.com/mpickering/hie-bios/issues/246

jneira:
I think we could merge this, without waiting to the hie-bios fix

fendor:
@jneira It is fixed on the hie-bios end with hackage revisions. The version bounds are fixed for version `0.7.0` and `0.7.1`, sorry if I did not communicate that clearly

jneira:
Dont worry, i did not check the linked issue neither.

PR #764, opened 2020-09-04
Allow for implicit (single file) cradles

pepeiborra:
I recently fixed a bug in the logic that figures out the path to an imported module arising when modules A B belong to different cradles C_A and C_B (https://github.com/wz1000/ghcide/pull/10/files). That fix assumes that the cradles list all the targets, an assumption that we already make in many other places, and apparently introduces a regression in implicit cradle handling.

This fix is intended to partially address that regression

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -322,7 +322,8 @@ getLocatedImportsRule =
         let dflags = addRelativeImport file (moduleName $ ms_mod ms) $ hsc_dflags env
         opt <- getIdeOptions
         let getTargetExists nfp
-                | HashSet.null targets || nfp `HashSet.member` targets = getFileExists nfp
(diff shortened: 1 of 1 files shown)


wz1000:
Looks good. Comment explaining would be nice though.

Also, single files should already work (since they don't import any other `.hs` file). I believe this fixes things for when there is an implicit cradle for a multi-file project which lists no explicit targets, but we(hie-bios) count the file we are loading as a target any way.

Also, this approach might lead to problems when there is a single file component. Though this will be rare, since for the issue to trigger, there must be a single file component that has an overlap in include paths with a different component (as described here https://github.com/haskell/ghcide/pull/734), but the user has only loaded the single file component.

pepeiborra:
Added a description. Though I agree that this fix is unsatisfactory, so I will probably close this PR altogether.

pepeiborra:
Instead it might be more productive to detect the situation where a cradle doesn't list any targets and fail early.

PR #766, opened 2020-09-05
Fix documentation (or source) link when html file is less specific than module

DunetsNM:
Below is an example when documentation is available but wasn't located because file name is less specific than module. This PR fixes it

**Documentation:**
Note the difference in doc file name and module name.

image

**Source:**
In this case both src and module name are the same.

image

**Other notes:**

* Also all directories attempted (not sure if it can possibly be more than one directory but original code is trying only first anyway)
* Just in case I'm on Windows (10)
* I'm experiencing more troubles with docs/sources on Windows (docs for most libs reported to be at $topdir/../ which doesn't work on Windows) but let's tackle one thing at a time.
* using stack

### DIFF
--- a/src/Development/IDE/Spans/Documentation.hs
+++ b/src/Development/IDE/Spans/Documentation.hs
@@ -171,7 +171,7 @@ lookupSrcHtmlForModule =
 lookupHtmlForModule :: (FilePath -> FilePath -> FilePath) -> DynFlags -> Module -> IO (Maybe FilePath)
 lookupHtmlForModule mkDocPath df m = do
   -- try all directories
-  let mfs = fmap (concat . fmap go) (lookupHtmls df ui)
+  let mfs = fmap (concatMap go) (lookupHtmls df ui)
   htmls <- filterM doesFileExist (concat . maybeToList $ mfs)
   return $ listToMaybe htmls
   where
--- a/src/Development/IDE/Spans/Documentation.hs
+++ b/src/Development/IDE/Spans/Documentation.hs
@@ -170,15 +170,20 @@ lookupSrcHtmlForModule =
 
 lookupHtmlForModule :: (FilePath -> FilePath -> FilePath) -> DynFlags -> Module -> IO (Maybe FilePath)
 lookupHtmlForModule mkDocPath df m = do
-  let mfs = go <$> (listToMaybe =<< lookupHtmls df ui)
+  -- try all directories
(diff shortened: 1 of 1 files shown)


pepeiborra:
Thanks for your contribution! I note that this feature lacks a test suite, and therefore it's very hard to tell whether the changes you propose break anything, in any of the supported platforms and build tools.

Would you be able to contribute such a test suite? Ideally it would cover projects using both cabal and stack, and our CI system will take care of running both in Windows and LInux

wz1000:
This approach is not correct, strictly speaking. Instead of stripping out the last part of the module name (which has no relation with the module that actually exports the documentation for the module), a better idea would be to look in the `tcg_rdr_env` to check which import actually brought the name into scope and then looking up the haddocks for that module.

bubba (approved):

bubba:
Should be fine now as a plaster, but it would be nice to do it the proper way at some point as well as what @wz1000 said
@pepeiborra for the tests I think there's some functional tests to make sure that the Documentation/Source things are generated but they don't actually check for the correct URL:
https://github.com/haskell/ghcide/blob/0d7cae984692bb088ac0da1ea4b1ac1828a4f129/test/exe/Main.hs#L2171-L2177

DunetsNM:
> This approach is not correct, strictly speaking. Instead of stripping out the last part of the module name (which has no relation with the module that actually exports the documentation for the symbol), a better idea would be to look in the `tcg_rdr_env` to check which import actually brought the name into scope and then looking up the haddocks for that module.

@wz1000  This is probably neither the ultimate solution. The "NormalizedFilePath" from the original screenshot is imported from local module "Development.IDE.Types.Location" - not module from the package for which haddock file exists (confirmed it by looking up the import decl in `GlobalRdrEnv` : `mi_globals` on the containing `ModIface', then located the `GlobalRdrElt' by `Name`). 
Even if we go down recursively to the package module it also doesn't guarantee that imported module from the package is the one that has corresponding Haddock file (I don't think so).

One counter-example is `import Control.Lens` which brings a lot of stuff into scope, but corresponding haddock file (doc/lens-4.17.1/Control-Lens.html) is just a short Table of Contents to sub-module htmls. 

Overall, trying to replicate Haddock's logic feels wrong, it might be better to start from the opposite end e.g. parse doc-index.json to pinpoint exact location and anchor (unless there's any simpler api to parse Haddock index available - I don't know any)

PR #767, opened 2020-09-05
Fully asynchronous request handling

pepeiborra:
#727 made the Shake queue able to concurrently handle more than one rule, but ghcide still handled LSP requests sequentially. This change makes the LSP handler multi-threaded, so now ghcide no longer blocks while e.g. computing an expensive code lens.

This change also adds a custom request protocol for testing, which can be generally useful. For instance, tests can now query the location of the interface files cache dir which will be useful for testing #760

Reviewers - things to watch for:

- New space leaks
- New race conditions

To do:

- [ ] run benchmarks

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -81,7 +81,8 @@ main = do
     dir <- IO.getCurrentDirectory
     command <- makeLspCommandId "typesignature.add"
 
-    let plugins = Completions.plugin <> CodeAction.plugin <> Test.plugin
+    let plugins = Completions.plugin <> CodeAction.plugin
+            <> if argsTesting then Test.plugin else mempty
         onInitialConfiguration :: InitializeRequest -> Either T.Text LspConfig
         onInitialConfiguration x = case x ^. params . initializationOptions of
           Nothing -> Right defaultLspConfig
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -30,6 +30,7 @@ import Development.IDE.Types.Logger
 import Development.IDE.Plugin
 import Development.IDE.Plugin.Completions as Completions
 import Development.IDE.Plugin.CodeAction as CodeAction
+import Development.IDE.Plugin.Test as Test
 import Development.IDE.Session
 import qualified Language.Haskell.LSP.Core as LSP
 import Language.Haskell.LSP.Messages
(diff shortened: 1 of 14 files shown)


wz1000:
Looks good to me. Now let's see how many tests break because of the ordering of messages...

wz1000 (approved):

exe/Main.hs
```diff
@@ -30,6 +30,7 @@ import Development.IDE.Types.Logger
 import Development.IDE.Plugin
 import Development.IDE.Plugin.Completions as Completions
 import Development.IDE.Plugin.CodeAction as CodeAction
+import Development.IDE.Plugin.Test as Test
```
wz1000, line 33: Wonder if we can move the plugin to the test directory, and use a cabal flag to enable it only for tests.

wz1000 (approved):

exe/Main.hs
pepeiborra: We don't build ghcide in the tests stanza, it's only built once in the main stanza. But I've changed Main to "link" the plugin only when `--testing` is given

pepeiborra:
The test suite is failing because the new `CustomHandler` responds to a message sent by `testSessionWait` in a way that lap-test cannot parse. 

The message that causes the issue: 

https://github.com/haskell/ghcide/blob/597a0c5d0edf9d1ad4b80184080597ec8d4cf255/test/src/Development/IDE/Test.hs#L61

I've raised this up in the lsp-test issue tracker https://github.com/bubba/lsp-test/issues/73

wz1000:
Maybe we can merge after marking the test as known broken.

pepeiborra:
It's all the tests that use `testSessionWait`

pepeiborra:
Early benchmarks not looking good - edit performance is much slower. There is increased memory usage but that seems to be a regression upstream

image

wz1000:
What is `master`, `upstream`, `upstream1` and `upstream2`?

pepeiborra:
> What is `master`, `upstream`, `upstream1` and `upstream2`?

I'll publish the whole benchmark artifacts once they are ready. In that screenshot:

- Master points to v0.3.0 in my repo 
- Upstream/master is Haskell/ghcide:master
- Upstream1 is upstream~1
- etc

pepeiborra:
After re-running the benchmarks in an isolated non-throttling CPU it looks like this branch does not introduce any time or space regression. Will upload full results shortly.

I have opened #771 to address the pre-existing space leak.

pepeiborra:
The tests are now passing after upgrading to a fixed lsp-test version.

The artefacts from the `ghcide_bench_linux_stack` pipeline show no change in memory usage. There's also no significant change in time usage, but this not a reliable measurement since the benchmark runs in a shared environment.

Local benchmarks do show a 10% time decrease in the edit experiment.

image

PR #778, opened 2020-09-09
Fix obsolete hie.yaml.cbl and hie.yaml.stack

DunetsNM:
stack and backup cabal hie files lagged behind


DunetsNM:
Also can entirely delete and gitgnore the hie.yaml and leave it up to developer, like it is done in haskell-language-server

jneira (approved):
LGTM but not sure about ignoring hie.yaml

pepeiborra:
Deleting `hie.yaml` and adding it to `.gitignore` sounds like a good idea to me

fendor:
IIRC, @ndmitchell was in favour of a default `hie.yaml` to make it easier to start hacking on a project.

ndmitchell:
Ideally whatever implicit or explicit thing we do should enable people to start hacking our projects with a good IDE experience. I want to enable that for my personal projects (eg HLint) but it's we can't do it for this project, it's a bit of a failing and we are roughly recommending people don't bother.

DunetsNM:
OK I reverted the deletion. Won't harm to fix outdated backup configs anyway

PR #783, opened 2020-09-12
Add Haddocks and exports for use* combinators

pepeiborra:


pepeiborra:
`iface-error-test-2` has become extremely flaky and is blocking this PR too. I am disabling it in my other PR but we should figure out when it started and why.

wz1000:
I was able to restart individual jobs from the github actions UI.

wz1000 (approved):

PR #784, opened 2020-09-12
Add GetHieAsts rule, Replace SpanInfo, add support for DocumentHighlight and scope-aware completions for local variables

### LINKED ISSUE #663: Upstream things from hls-2
The hls-2 branch is `ghcide` head plus some changes intended to be upstreamed. As per https://github.com/haskell/haskell-language-server/issues/173, it seems once they are upstreamed hls can use ghcide master. These changes aren't too major, they mainly consist of

- [x] This PR: https://github.com/digital-asset/ghcide/pull/624
- [x] A backport of `.hie` files to ghc 8.6
- [ ] A rewrite of `definition`/`typedefinition`/`hover` using `.hie` files
- [ ] `documentSymbols` using `.hie` files
- [x] A module graph
- [x] Propagating changes to parent modules on save using the module graph
- [x] @bubba's patch to add hyperlinks to hover
- [ ] Initial support for telemetry thanks to @mpardalos

The `hiedb` changes are not included because I don't think they are ready yet.

Taken from a list by @wz1000 in https://github.com/haskell/haskell-language-server/issues/173#issuecomment-647933484

wz1000:
This is all the remaining changes from the `hls-3` branch.

Drops support for GHC 8.4

Fixes #663 

Changes:
- Remove the `GetSpanInfo` rule, depend entirely on `GetHieAst` for hover, go to definition etc.
 
New features:
- Document Highlight for highlighting references of a Name in the same file
- Completion for local variables in scope via `GetBindings`

Missing features:
- ~~Go to definition for module names~~
- ~~Kind info on hover~~
- Relevant constraints on hover

TODOs:

- [x] Add tests for new features
- [x] fix as many existing tests as possible
- [x] Re-add as many missing features as possible, using something like the `DocMap` mechanism
- [x] Benchmark performance differences

Related GHC MRs for `.hie` file changes:

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4037

https://gitlab.haskell.org/ghc/ghc/-/merge_requests/4068

One has already been merged into the ghc-9.0 branch, the other will be merged soon hopefully

Old description:

This is the slight modification of what already lives in the `hls-3` and `hiedb-3` branches, necessary for https://github.com/haskell/haskell-language-server/pull/391/

We make some modifications to be able to save the `HieAst` before compressing the types. We compress the types on demand when we actually need to write the file to disk. This should improve `hls` performance somewhat as we get rid of the overhead of type compression (which can be expensive) in most cases.


isovector:
Marvelous. Thank you so much!

pepeiborra:
Will other clients like Stan be able to use the hie files produced by ghcide? If not, maybe we should change the file extension

pepeiborra:
Any changes worth upstreaming to ghc?

src/Development/IDE/Core/Rules.hs
```diff
@@ -570,14 +585,20 @@ typeCheckRuleDefinition hsc pm isFoi source = do
         case isFoi of
           IsFOI Modified -> return (diags, Just tcm)
```
pepeiborra, line 586: how come we don't compute HIE information for files of interest?

src/Development/IDE/Core/Rules.hs
```diff
@@ -504,6 +504,21 @@ getDependenciesRule =
         let mbFingerprints = map (fingerprintString . fromNormalizedFilePath) allFiles <$ optShakeFiles opts
         return (fingerprintToBS . fingerprintFingerprints <$> mbFingerprints, ([], transitiveDeps depInfo file))
 
+getHieAstsRule :: Rules ()
+getHieAstsRule =
+    define $ \GetHieAst f -> do
```
pepeiborra, line 509: How come this rule isn't used anywhere?

pepeiborra (commented):
This PR seems a bit incomplete. Is there any reason for that?

wz1000:
> Will other clients like Stan be able to use the hie files produced by ghcide? If not, maybe we should change the file extension

Yes, the hie files are generated in precisely the same way.

> Any changes worth upstreaming to ghc?

Yeah, we need `enrichHie` exported. I will make a MR today. Hopefully it can get into 9.0

src/Development/IDE/Core/Rules.hs
wz1000: We do, in the `GetHieAst` rule. `Typecheck` only computes HIE information if it is writing out the ifaces. If we aren't doing that, typecheck will not compute it. `GetHieAst` depends on `Typecheck`, and will reuse the information if `Typecheck` already computed it, otherwise it will compute it itself.

src/Development/IDE/Core/Rules.hs
wz1000: It's meant to be used in `hls` code for now. 

I have code in the `hls` branch that replaces all usages of `GetSpanInfo` with this. However, since we can't generate HIE information for 8.4, so I've not included that in this patch.

One thing we can do is use it to do is to produce scope-aware local completions. I believe @isovector is working on generating one crucial requirement for this as part of his tactics PR(the interval map telling us which names are in scope at a point). The other part is #785, after which it should be just a matter of tying the two things together. We can wait for this to land before merging this patch.

pepeiborra: I'm happy to drop 8.4 compatibility in the next ghcide release.

pepeiborra: But that said, why not change `getDefinition` and `getTypeDefinition` to use `GetHieAsts` in this PR?

pepeiborra:
I just updated the benchmarks to measure allocations and delayed actions

src/Development/IDE/Spans/LocalBindings.hs
```diff
@@ -0,0 +1,73 @@
...
+-- | Given a 'Bindings' get every identifier in scope at the given
+-- 'RealSrcSpan',
+getLocalScope :: Bindings -> RealSrcSpan -> [(Name, Maybe Type)]
+getLocalScope bs rss
+  = nameEnvElts
+  $ foldMap snd
+  $ IM.dominators (realSrcSpanToInterval rss)
+  $ getBindings bs
+
+getFuzzyScope :: Bindings -> Position -> Position -> [(Name, Maybe Type)]
```
pepeiborra, line 68: doc comments

src/Development/IDE/Spans/LocalBindings.hs
```diff
@@ -0,0 +1,73 @@
...
+realSrcSpanToInterval rss =
+  Interval
+    (realSrcLocToPosition $ realSrcSpanStart rss)
+    (realSrcLocToPosition $ realSrcSpanEnd   rss)
+
+------------------------------------------------------------------------------
+-- | Compute which identifiers are in scope at every point in the AST. Use
+-- 'getLocalScope' to find the results.
+bindings :: RefMap -> Bindings
+bindings refmap = Bindings $ foldr (uncurry IM.insert) mempty  $ do
```
pepeiborra, line 33: This needs to go through the full list of name envs before it returns anything, right? If so, maybe it should be a `foldl'`

pepeiborra (approved):
There's a lot of changes I do not understand, but they've been in HLS since 0.2.0 so I am happy to merge them provided the tests pass and the new benchmarks don't show any nasty regressions. 

progress

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -392,19 +396,7 @@ getCompletions ideOpts cc pm pmapping prefixInfo caps withSnippets = do
           to                             'foo :: Int -> String ->    '
                                                               ^
       -}
-      pos =
```
wz1000, line 395: We don't need this stuff anymore because of fuzzy position mapping. I had to wade through multiple git commits across repositories to find @bubba as the source of this code. If I understand correctly, it existed so that we could detect type contexts even if the module didn't parse.

Suppose the cursor is at the `|`

```haskell
foo :: Integer -> Integer -> |
```

So the idea was to strip out the trailing "-> ", and then look for the "context" of the position (whether it was in a value/type/import etc.) at the position (`newPositionToOld` would just return `Nothing`).

Now, `fromDelta` will return the original range that was inserted to/replaced, so we can use it instead.

wz1000:
I'm a bit busy this week, will hopefully finish this up on the weekend.

Till then, here's a gif demonstrating local completions(you can try it out by running `haskell-language-server`):

wz1000:
I am leaning towards not adding back the "relevant constraints on hover" feature.

1. It is not complete, it misses many constraints (for example, constraints introduced by GADT pattern matching etc.)
2. It will require a generic `SYB` style traversal of the AST, which is not needed for anything else anymore thanks to this PR
3. GHC 9.0 will allow writing this feature properly with the changes from https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1286
4. (the weakest reason) this feature hasn't existed in HLS for a long time and nobody has seemed to miss it.

/cc @pepeiborra @ndmitchell thoughts?

wz1000:
I have fixed as many tests as I plan to. The remaining failures are genuine behavior divergences that I don't think are worth fixing. Please review them once CI completes.

The only TODO remaining is to add tests for the new features.

pepeiborra:
I am fine with dropping the "constraints on hover" feature, it was  bit erratic for me and only seemed to work on declaration lhs's. Is there any chance of back porting #1286 to the ghcide fork of Hie files?

Ignoring the tests related to that feature, there are only 5 failing tests, all related to hover. Good job! 

Please share some benchmarks if you have the chance too.

wz1000:
Here are the interesting benchmark results:

`hover after edit` seems to be down:



`completions after edit` is up, but that is to be expected since we are doing more work, calculating local completions:



There is some funkyness with the allocations for document symbols I can't explain:



The rest seem to be unchanged, except for the shape of the allocation graph: http://78.47.113.11/benchmarks/

pepeiborra:
Look at the "Delayed work" column in the `results.csv` file:
- "Hover after edit" is almost twice as efficient in delayed time. 50% more efficient in user time, 50% less allocations
- "Document symbols after edit" is almost twice as efficient in delayed time
- "Completions after edit" is almost twice as inefficient in delayed time and allocates 19% more

wz1000:
What exactly is delayed time?

pepeiborra:
> What exactly is delayed time?

The time spent waiting for all the delayed actions in the Shake queue to complete

wz1000:
The regression in completion performance is expected since we now depend on `Typecheck`, not just `GetParsedModule`. Stale information should hide most of the difference in actual use I think.

wz1000:
I have no idea why there is a change for `DocumentSymbols`.

wz1000:
Added tests! This is done except for that one `Unexpected Pass` on 8.10.

wz1000:
I don't have the energy to figure out why the test is passing on 8.10 but not before that. I diffed the `HieAst` generation code, but could find no obvious culprits. CPPing the error away. With any luck this is good to go now.

wz1000:
Any opinions on whether I should remove 8.4 specific CPP in this PR, or a different one?

pepeiborra:
I would say merge this when ready, clean up 8.4 CPP in another PR

wz1000:
OK, I think this is ready, we can merge if you are happy.

PR #785, opened 2020-09-12
Extend position mapping with fuzzy ranges

wz1000:
We sometimes need to know the original range that a position mapped to, when a simple failure is not helpful.
For example, when the user is typing inside a do block, we would like to use position mapping to tell us that the current cursor position is within the span of the do block, where the span is known from the old version of the file for which we have compiler artifacts. This can be helpful for things like generating completions based on local variables.

### DIFF
--- a/src/Development/IDE/Core/PositionMapping.hs
+++ b/src/Development/IDE/Core/PositionMapping.hs
@@ -2,6 +2,10 @@
 -- SPDX-License-Identifier: Apache-2.0
 module Development.IDE.Core.PositionMapping
   ( PositionMapping(..)
+  , PositionResult(..)
+  , lowerRange
+  , upperRange
+  , positionResultToMaybe
   , fromCurrentPosition
   , toCurrentPosition
   , PositionDelta(..)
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Core/PositionMapping.hs
```diff
@@ -21,17 +25,45 @@ import qualified Data.Text as T
 import Language.Haskell.LSP.Types
 import Data.List
 
+data PositionResult a = PositionRange { unsafeLowerRange :: a, unsafeUpperRange :: a } | PositionExact a
```
pepeiborra, line 28: Add bangs to fields

pepeiborra (approved):
Very nice, just add the missing bangs and good to merge

wz1000:
I'm a bit concerned about adding bangs to the datatype definition, I must admit. I think it changes the asymptotics of 'n' binds in sequence from linear to exponential. Let me confirm by writing a synthetic benchmark.

wz1000:
Yes, its as I suspected.

Here is the benchmark I'm using

```haskell
f :: Int -> PositionResult Int
f i = PositionRange (i-1) (i+1)

g :: Int -> Int -> PositionResult Int
g 0 x = pure x
g i x = f x >>= g (i-1)

main :: IO ()
main = do
  print $ g 100 0
```

with

```haskell
data PositionResult a = PositionRange { unsafeLowerRange :: !a, unsafeUpperRange :: !a } | PositionExact !a
```

This doesn't terminate any time soon.

It terminates pretty much instantly with the following:
```haskell
data PositionResult a = PositionRange { unsafeLowerRange :: a, unsafeUpperRange :: a } | PositionExact !a
```

It also terminates quickly if we add bangs to the Monad instance:
```haskell
instance Monad PositionResult where
  (PositionExact a) >>= f = f a
  (PositionRange !lower !upper) >>= f = PositionRange lower' upper'
    where
      lower' = lowerRange $ f lower
      upper' = upperRange $ f upper
```

However, adding bangs to `lower'` and `upper'` makes it diverge again

pepeiborra:
Wow, good catch!

PR #790, opened 2020-09-13
Disable optimisation in tests

pepeiborra:
Ideally we would do this with a Cabal flag, but I don't think it is possible
to disable optimisation only for the tests stanza

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -355,7 +355,7 @@ test-suite ghcide-tests
         text
     hs-source-dirs: test/cabal test/exe test/src bench/lib
     include-dirs: include
-    ghc-options: -threaded -Wall -Wno-name-shadowing
+    ghc-options: -Wall -Wno-name-shadowing -O0
     main-is: Main.hs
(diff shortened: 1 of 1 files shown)


ghcide.cabal
```diff
@@ -355,7 +355,7 @@ test-suite ghcide-tests
         text
     hs-source-dirs: test/cabal test/exe test/src bench/lib
     include-dirs: include
-    ghc-options: -threaded -Wall -Wno-name-shadowing
```
ndmitchell, line 358: You are also removing `-threaded`. Do we not run the tests in parallel? Or at least have the option to do so? Or does this actually break our test suite.

pepeiborra: We don't run the tests in parallel, see:

https://github.com/haskell/ghcide/blob/114e184e3228afbf20c3a3ca57e2ebb8771cce1d/test/exe/Main.hs#L60-L66

ndmitchell (approved):

PR #794, opened 2020-09-13
Skip unnecessary packing of cache artifacts

pepeiborra:
Not needed when we had a successful cache hit, saves ~ 2 minutes

### DIFF
--- a/.azure/windows-stack.yml
+++ b/.azure/windows-stack.yml
@@ -56,6 +56,11 @@ jobs:
       fi
     displayName: 'stack build --only-dependencies'
   - bash: |
+      mkdir -p .azure-cache
+      tar -vczf .azure-cache/stack-root.tar.gz $(cygpath $STACK_ROOT)
+      tar -vczf .azure-cache/stack-work.tar.gz .stack-work
+    displayName: "Pack cache"
+  - bash: |
       if [ "$STACK_YAML" = "stack8101.yaml" ]; then
         stack test --no-run-tests --ghc-options="-Werror -fexternal-interpreter" --stack-yaml $STACK_YAML || stack test --no-run-tests --ghc-options="-Werror -fexternal-interpreter" --stack-yaml $STACK_YAML || stack test --no-run-tests --ghc-options="-Werror -fexternal-interpreter" --stack-yaml $STACK_YAML
       else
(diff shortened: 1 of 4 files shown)


ndmitchell:
But won't the cache degrade over time? So we get a cache hit, add some new dependencies, but those will never make it to the cache?

pepeiborra:
The cache key includes the stack and cabal descriptors, which carry enough signal to ensure the cache doesn't degrade.

ndmitchell (approved):
2 min less sounds great

.azure/windows-stack.yml
```diff
@@ -38,9 +38,11 @@ jobs:
   - bash: |
       curl -sSkL http://www.stackage.org/stack/windows-x86_64 -o /usr/bin/stack.zip
       unzip -o /usr/bin/stack.zip -d /usr/bin/
-      mkdir -p "$STACK_ROOT"
+      mkdir -p "$(cygpath $STACK_ROOT)"
```
jneira, line 41: @pepeiborra was `cygpath` needed? afaik `mkdir -p "C:\x"` works (at least in my local msys console)

```bash
atrey@FNTSY MINGW64 /d/dev
$ mkdir -p "D:\dev\test"

atrey@FNTSY MINGW64 /d/dev
$ ls test/
```

pepeiborra: I don't have a Windows environment, only added it for consistency

.azure/windows-stack.yml
```diff
@@ -21,13 +21,13 @@ jobs:
...
     inputs:
-      key: stack-cache-v2 | $(Agent.OS) | $(Build.SourcesDirectory)/$(STACK_YAML) | $(Build.SourcesDirectory)/ghcide.cabal
+      key: stack-cache-v3 | $(Agent.OS) | $(Build.SourcesDirectory)/$(STACK_YAML) | $(Build.SourcesDirectory)/ghcide.cabal
       path: .azure-cache
       cacheHitVar: CACHE_RESTORED
     displayName: "Cache stack artifacts"
   - bash: |
       mkdir -p $STACK_ROOT
-      tar -vxzf .azure-cache/stack-root.tar.gz -C /c
+      tar -vxzf .azure-cache/stack-root.tar.gz -C $(cygpath C:\\)
```
jneira, line 30: i think `/c` also works

.azure/windows-stack.yml
jneira: Fair enough but i am afraid that the existing `cygpath` was the exception (to be removed) :-/
I can remove them in a follow up if you prefer

jneira (approved):
LGTM

.azure/windows-stack.yml
pepeiborra: If you could fix the caching for the Windows CI I would be eternally grateful. Otherwise I might have to disable since the network connection to the stackage infrastructure (`casa.fpcomplete.com`) has become very unreliable and stack builds fail all the time

PR #795, opened 2020-09-13
Support parsedResultAction of GHC plugins

### LINKED ISSUE #262: Add support for record-dot-preprocessor
It would be very nice to support record dot syntax enabled by [record-dot-preprocessor](https://hackage.haskell.org/package/record-dot-preprocessor) (either preprocessor or plugin).

srid:
Support for GHC plugins that use the [`parsedResultAction`](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/extending_ghc.html#parsed-representation) field to modify the syntax tree, eg [record-dot-preprocessor](https://hackage.haskell.org/package/record-dot-preprocessor). Resolves #262

- [x] Implementation, and tests
- [x] Investigate and resolve the timeout in the added test case


ghcide.cabal
```diff
@@ -340,6 +340,7 @@ test-suite ghcide-tests
         process,
         QuickCheck,
         quickcheck-instances,
+        record-dot-preprocessor,
```
ndmitchell, line 343: Not sure this builds on all the platforms we test on

ndmitchell (commented):
Seems you added the tests, but not the change to invoke the plugin?

srid:
Test times out for me locally.

image

ghcide.cabal
srid: Is there a bulitin (or working) plugin that uses `parsedResultAction`?

srid:
Umm, looks like I might just have to close this and open a new PR targetting the `master` branch.

(I was testing it against HLS, so didn't use `master` at that time)

srid:
hlint fails silent? 

https://dev.azure.com/garyverhaegen/temp-ghcide/_build/results?buildId=76&view=logs&j=0b7da2c4-8254-5192-f65f-abc5eca10828&t=b575db2a-71ce-5d51-eeee-a87d1677df6c&l=33

pepeiborra:
> hlint fails silently?
> 
> https://dev.azure.com/garyverhaegen/temp-ghcide/_build/results?buildId=76&view=logs&j=0b7da2c4-8254-5192-f65f-abc5eca10828&t=b575db2a-71ce-5d51-eeee-a87d1677df6c&l=33
```
src/Development/IDE/Core/Preprocessor.hs:4:1-20: Warning: Avoid restricted extensions
Found:
  {-# LANGUAGE CPP #-}
Note: may break the code

1 hint
```
Please move all the CPP to `D.I.GHC.Compat`

ghcide.cabal
ndmitchell: I think this might work actually because we are ditching 8.4 support. #803

src/Development/IDE/Core/Compile.hs
```diff
@@ -582,7 +583,11 @@ parseFileContents customPreprocessor dflags comp_pkgs filename modTime contents
                        , pm_annotations = hpm_annotations
                       }
                    warnings = diagFromErrMsgs "parser" dflags warns
-               pure (warnings ++ preproc_warnings, pm)
+                   applyPluginAction p opts = parsedResultAction p opts ms
+               hspm <- liftIO $ runHsc env $ withPlugins dflags applyPluginAction 
+                 (HsParsedModule parsed' [] hpm_annotations)
+               let pm' = pm { pm_parsed_source = hpm_module hspm }
```
ndmitchell, line 589: This looks odd. Should we really be using hpm_module in other places?

stack.yaml
```diff
@@ -2,6 +2,7 @@ resolver: nightly-2019-09-21
 packages:
 - .
 extra-deps:
+- aeson-1.4.6.0
```
ndmitchell, line 5: Why?

src/Development/IDE/GHC/Compat.hs
```diff
@@ -467,3 +470,12 @@ wopt_unset_fatal :: DynFlags -> WarningFlag -> DynFlags
 wopt_unset_fatal dfs f
     = dfs { fatalWarningFlags = EnumSet.delete f (fatalWarningFlags dfs) }
 #endif
+
+initializePlugins :: MonadIO m => HscEnv -> DynFlags -> m DynFlags
```
ndmitchell, line 474: Why not make this return IO, since that's all it requires.

ndmitchell (commented):
This looks pretty good. CI working would be great. Maybe that requires 8.4 support being removed, or conditionals, but nice work!

stack.yaml
srid: Oh, was cherry-picking from #801. It was going to be a temporary change in the PR. Gone now, due to merge from `master`

src/Development/IDE/Core/Compile.hs
srid: I had to convert back and forth between `HsParsedModule` and `ParsedModule`, because [`parsedResultAction`](https://hackage.haskell.org/package/ghc-8.10.1/docs/Plugins.html#v:parsedResultAction) deals with the former, whereas this function, `parseFileContents`, is dealing with the later.

This line is converting from the former (`hspm`) to the later (`pm'`), which is why we are pulling the parsed source from `hpm_source`. I'm not sure I understand why we would need to use it in other places?

Perhaps if I refactor it to be like this it would be more clear (transforming `parsed'` to `parsed''` where the plugin is applied)?

image

srid:
Just pushed a fix for the test timeout. Hopefully CI passes this time.

srid:
@ndmitchell About the 8.4 failures, which happens [during build](https://dev.azure.com/garyverhaegen/temp-ghcide/_build/results?buildId=99&view=logs&j=b0a35ed2-e50b-5e32-9491-924aca3e67a7&t=83bc6b67-ca1d-5a9e-5a6f-abb7118d168a&l=15), should I add conditional to the cabal file, or wait for #803 to be done in `master`?

srid:
All tests passed, except for `ghcide_stack_linux stack_810` timing out after 60m (which seems like known issue?).

src/Development/IDE/Core/Compile.hs
srid: All of this code got moved to `D.I.G.Compat`.

pepeiborra (approved):
Awesome stuff, thank you!

pepeiborra:
Hmm, could you run the benchmarks locally and share the results? You can use either `stack bench` or `cabal bench`, both should work. By default they compare your master and current branch, so make sure that your `master` branch points to HEAD or edit `bench/hist.yaml` to adjust.

srid:
@pepeiborra How do you run benchmarks on NixOS?

They fail with `stack bench` for me:
image

It looks like stack is discarding environment variables like PATH by using pure nix shell. 

Doing `stack --no-nix-pure bench` throws `In Nix shell but reExecL is False`. See https://github.com/commercialhaskell/stack/issues/5000

Is running ghcide benchmarks supported on Nix/NixOS?

pepeiborra:
Try telling the benchmark runner to use cabal  instead of stack by editing `bench/hist.yaml`

srid:
Plain cabal never worked in nix-shell of ghcide, which is why I resorted to using stack. With that change to `bench/hist.yaml`, running `stack --no-nix-pure bench` reports. I'll have to try this out from my dusty old macbook when I get home later today.

image

pepeiborra:
Cabal update ?

srid:
@pepeiborra https://github.com/srid/ghcide-pr795-bench

pepeiborra:
Looks good, thanks!

PR #801, opened 2020-09-13
Fix stack 8.6 build

pepeiborra:


jneira:
Mmm, the ci build for stack86.yaml was failing due to #763, but the ci build for the pr passed. And master works for me locally using the default stack.yaml just now.
It is due a second hackage revision on hie-bios which bumped up the lower bound over aeson : https://hackage.haskell.org/package/hie-bios-0.7.1/revisions/
I would put explicitly the revision in the stack yaml file (`- hie-bios-0.7.1@rev:2`), cause an already existing cache would bypass the revision. `stack` interacts badly with hackage revisions :worried: 
Thanks for fixing it.

pepeiborra:
> Mmm, the ci build for stack86.yaml was failing due to #763, but the ci build for the pr passed. And master works for me locally using the default stack.yaml just now.
> It is due a second hackage revision on hie-bios which bumped up the lower bound over aeson : https://hackage.haskell.org/package/hie-bios-0.7.1/revisions/
> I would put explicitly the revision in the stack yaml file (`- hie-bios-0.7.1@rev:2`), cause an already existing cache would bypass the revision. `stack` interacts badly with hackage revisions 😟
> Thanks for fixing it.

This doesn't seem like stack's fault, but rather the cache right? In any case I'm keen to merge this as soon as possible in order to unblock other PRs, so I'm going to leave it as is. Please send a separate PR after this lands if you have strong feelings about adding the revision to the stack descriptors.

jneira (approved):

jneira:
it seems the bench component is failing cause it is picking master to compare results and master cant be built?

PR #804, opened 2020-09-14
Add completion tests for records.

gdevanla:
Tests to demonstrate the current functionality of completions for records.


gdevanla:
Thanks. I can remove the `draft-status`. I had created this here to link it from the haskell/haskell-language-server#333

pepeiborra (approved):

PR #808, opened 2020-09-15
Enable windows ci

### LINKED ISSUE #805: Fix Windows CI caching and reenable Windows CI
If you could fix the caching for the Windows CI I would be eternally grateful. Otherwise I might have to disable since the network connection to the stackage infrastructure (`casa.fpcomplete.com`) has become very unreliable and stack builds fail all the time

_Originally posted by @pepeiborra in https://github.com/haskell/ghcide/pull/794#discussion_r488124365_

jneira:
* Closes #805

### DIFF
--- a/.azure/windows-stack.yml
+++ b/.azure/windows-stack.yml
@@ -47,12 +47,12 @@ jobs:
       stack install happy --stack-yaml $STACK_YAML
       stack install alex --stack-yaml $STACK_YAML
       stack install cabal-install --stack-yaml $STACK_YAML
-      # GHC 8.10 keeps crashing with various kinds of access violations and other
-      # errors so we retry 3 times.
+      # GHC 8.10.1 fails with ghc segfaults, using -fexternal-interpreter seems to make it working
+      # There are another transient errors like time outs downloading from stackage so we retry 3 times
       if [ "$STACK_YAML" = "stack8101.yaml" ]; then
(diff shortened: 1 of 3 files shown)


ndmitchell (approved):

pepeiborra (changes requested):
This doesn't fix the Azure cache issue

jneira:
It seems the workaround of packing the cache, copied from hie and needed due to azure issues with the cache task is no longer needed, so removing the pack/unpack is a win-win :smile: 
I think maybe we could remove it from linux jobs too, i'll do in a new pr.

.azure/windows-stack.yml
```diff
@@ -47,19 +46,14 @@ jobs:
       stack install happy --stack-yaml $STACK_YAML
       stack install alex --stack-yaml $STACK_YAML
       stack install cabal-install --stack-yaml $STACK_YAML
-      # GHC 8.10 keeps crashing with various kinds of access violations and other
-      # errors so we retry 3 times.
+      # GHC 8.10.1 fails with ghc segfaults, using -fexternal-interpreter seems to make it working
+      # There are another transient errors like time outs downloading from stackage so we retry 3 times
```
pepeiborra, line 50:
```suggestion
      # There are other transient errors like timeouts downloading from stackage so we retry 3 times
```

pepeiborra:
Wow, it really seems to work. Nice!

pepeiborra (approved):

PR #810, opened 2020-09-15
Mark files as modified on open

wz1000:
Initially I thought I would check if the file exists on disk, but then I realized that vscode violates our assumptions anyway, so we should mark it as `Modified` to be safe.


pepeiborra (approved):
Thanks for the quick fix!

PR #812, opened 2020-09-15
Remove pack/unpack cache from linux jobs

jneira:
* The workaround for some issues of the azure cache task is not longer needed so we can save some time and make the script simpler


pepeiborra (approved):
If this works it is very welcome

jneira:
One more run to check the use of the new cache

PR #814, opened 2020-09-16
Fix docs tooltip for base libraries on Windows

### LINKED ISSUE #769: (Windows, Stack) Documentation / Source links don't appear for base libraries, only for stackage
### Environment

Windows 10
stack 2.3.3
reproduced for GHC 8.6.5 and 8.10.1

built with haddock flag

### Symptoms

Links to docs and sources in tooltips appear just fine for libraries _from stackage_ (also not always atm, but it's a different story and there's a PR for it: https://github.com/haskell/ghcide/pull/766)
Stackage snapshots stored locally at `C:\sr\snapshots\`

aeson toJSON works fine:
image

But doc links don't appear for _base libraries_, stored at the stack ghc installation e.g. `c:\Users\KolA\AppData\Local\Programs\stack\x86_64-windows\ghc-8.10.1\doc\html\libraries`

no luck for Data.Text T.pack:
image

According to [the video in this post](https://mpickering.github.io/ide/posts/2020-07-10-ghc-libdir.html) it works fine on Mac (at least for T.pack), so I think it is Windows-specific.

### Cause

I fiddled with `Development.IDE.Spans.Documentation` and found out that `lookupHtmls` returns following directory for base/prelude stuff: 

for GHC 8.6.5 : `$topdir/../doc/html/libraries/base-4.12.0.0`
for GHC 8.10.1 : `$topdir/../doc/html/libraries/base-4.14.0.0`

assuming that $topdir is a synonym for location of ghc.exe (e.g. `c:\Users\KolA\AppData\Local\Programs\stack\x86_64-windows\ghc-8.10.1\bin\`) it should resolve documentation directory correctly however it doesn't.

As a matter of fact base docs just as well stored at snapshots directory `C:\sr\snapshots\` ! but with different subpath (no /html/libraries/ part - e.g. `c:\sr\snapshots\1b8f3c61\doc\base-4.14.0.0` )

### What to do?

I'm keen to fix it but not sure where to look for the root cause.   Should I look upstream in LSP ? in Stack ? Or in GHC itself ? Any hints appreciated.

DunetsNM:
This fixes #769 , see the issue thread for details


CHANGELOG.md
```diff
@@ -1,4 +1,26 @@
-### unreleased
+### 0.4.0 (2020-09-15)
```
ndmitchell, line 1: I don't think this should end up in this diff.

DunetsNM: I accidentally branched off latest / 0.4.0 prerelease , guess this PR won't get into 0.4 anyway so can just wait until Pepe's commit get into master

src/Development/IDE/Spans/Documentation.hs
```diff
@@ -172,7 +172,10 @@ lookupHtmlForModule :: (FilePath -> FilePath -> FilePath) -> DynFlags -> Module
 lookupHtmlForModule mkDocPath df m = do
   let mfs = go <$> (listToMaybe =<< lookupHtmls df ui)
   htmls <- filterM doesFileExist (concat . maybeToList $ mfs)
-  return $ listToMaybe htmls
+  -- canonicalize located html to remove /../ indirection which can break some clients
+  -- (vscode on Windows at least)
+  htmls' <- traverse canonicalizePath htmls
+  return $ listToMaybe htmls'
```
ndmitchell, line 178: This code applies doesFileExist/canonicalizePath to a sequence of items, then only takes the first. Yuk. That means we are paying for more IO operations that we need. Can we change the `filterM` to be a `findM`, then we have a `Maybe FIlePath`, you can still traverse it to canonicalise, but now we only canonicalize one element, and can skip the `listToMaybe`.

(I appreciate this wasn't your code to begin with, but a good thing to fix)

ndmitchell (commented):
Is there any way to write a test for this code? I imagine if we did a `doesFileExist` on the resulting path in the tooltip, before this diff it wouldn't have worked, after it will? I'm sensitive that this whole area feels a bit easy to break in future.

jneira:
very happy to see contributions for Windows, many thanks @DunetsNM

src/Development/IDE/Spans/Documentation.hs
DunetsNM: Did only the easy change so far, will look into testability a bit later
Thanks for the hint re findM, didn't know it (will take it as reminder to use Hoogle more often). I also had a thought about potentially wasteful IO on file exists/canonicalize , my naive intuition was that laziness will take care of it (which is not the case as it turned out)

pepeiborra (approved):

PR #815, opened 2020-09-16
Tag unused warning as such

serras:
This patch uses the new "unnecessary" or "deprecated" tags for diagnostics to show unused variables/functions/... dimmed in the editor, as discussed in https://github.com/haskell/ghcide/issues/165#issuecomment-692532054

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -136,7 +136,7 @@ typecheckModule (IdeDefer defer) hsc pm = do
             GHC.typecheckModule $ enableTopLevelWarnings
                                 $ enableUnnecessaryAndDeprecationWarnings
                                 $ demoteIfDefer pm{pm_mod_summary = tweak modSummary'}
-        let errorPipeline = unDefer . tagDiag . hideDiag dflags
+        let errorPipeline = unDefer . hideDiag dflags . tagDiag
             diags = map errorPipeline warnings
         tcm2 <- mkTcModuleResult tcm (any fst diags)
         return (map snd diags, tcm2)
@@ -252,9 +252,10 @@ upgradeWarningToError (nfp, sh, fd) =
 hideDiag :: DynFlags -> (WarnReason, FileDiagnostic) -> (WarnReason, FileDiagnostic)
 hideDiag originalFlags (Reason warning, (nfp, sh, fd))
   | not (wopt warning originalFlags)
-  = if warning `elem` unnecessaryDeprecationWarningFlags
-       then (Reason warning, (nfp, sh, fd{_severity = Just DsInfo}))
-       else (Reason warning, (nfp, HideDiag, fd))
+  = if null (_tags fd)
+       then (Reason warning, (nfp, HideDiag, fd))
+            -- keep the diagnostic if it has an associated tag
+       else (Reason warning, (nfp, sh, fd{_severity = Just DsInfo}))
 hideDiag _originalFlags t = t
 
 enableUnnecessaryAndDeprecationWarnings :: ParsedModule -> ParsedModule
@@ -262,7 +263,7 @@ enableUnnecessaryAndDeprecationWarnings =
   (update_pm_mod_summary . update_hspp_opts)
   (foldr (.) id [(`wopt_set` flag) | flag <- unnecessaryDeprecationWarningFlags])
 
--- warnings which lead to a diagnostic tag
+-- | Warnings which lead to a diagnostic tag
 unnecessaryDeprecationWarningFlags :: [WarningFlag]
 unnecessaryDeprecationWarningFlags
   = [ Opt_WarnUnusedTopBinds
@@ -281,6 +282,7 @@ unnecessaryDeprecationWarningFlags
     , Opt_WarnWarningsDeprecations
     ]
 
+-- | Add a unnecessary/deprecated tag to the required diagnostics.
 tagDiag :: (WarnReason, FileDiagnostic) -> (WarnReason, FileDiagnostic)
 tagDiag (Reason warning, (nfp, sh, fd))
   | Just tag <- requiresTag warning
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -269,7 +269,9 @@ tagDiag (Reason warning, (nfp, sh, fd))
 #if MIN_GHC_API_VERSION(8,10,0)
     requiresTag Opt_WarnUnusedRecordWildcards = Just DtUnnecessary
 #endif
+#if MIN_GHC_API_VERSION(8,6,0)
     requiresTag Opt_WarnInaccessibleCode      = Just DtUnnecessary
+#endif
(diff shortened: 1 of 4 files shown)


serras:
Mandatory screenshot (notice the grey color)

Captura de pantalla 2020-09-16 a las 20 45 06

ndmitchell:
Awesome! One question - I usually don't turn off unused variables, just unused top-level bindings. It would be great to still get the grey color. Could we always turn these warnings on, but suppress them from making it to the diagnostics if we turned them on rather than the user?

serras:
That sounds like a good idea, it shouldn't be more difficult than the `unDefer` bussiness in the same file.

serras:
It took a bit more than expected, mostly because some tests broke because of that. In fact, locally all the benchmarks fail for me, but I haven't been able to fix them.

src/Development/IDE/Core/Compile.hs
```diff
@@ -247,10 +250,55 @@ upgradeWarningToError (nfp, sh, fd) =
   warn2err = T.intercalate ": error:" . T.splitOn ": warning:"
 
 hideDiag :: DynFlags -> (WarnReason, FileDiagnostic) -> (WarnReason, FileDiagnostic)
-hideDiag originalFlags (Reason warning, (nfp, _sh, fd))
-  | not (wopt warning originalFlags) = (Reason warning, (nfp, HideDiag, fd))
+hideDiag originalFlags (Reason warning, (nfp, sh, fd))
+  | not (wopt warning originalFlags)
+  = if warning `elem` unnecessaryDeprecationWarningFlags
+       then (Reason warning, (nfp, sh, fd{_severity = Just DsInfo}))
+       else (Reason warning, (nfp, HideDiag, fd))
```
ndmitchell, line 257: I assume HideDiag means they don't appear at all? Could you just look at _tags ahd only Hide the ones with no tags? (I've no idea here, somewhat guessing)

src/Development/IDE/Core/Compile.hs
```diff
@@ -247,10 +250,55 @@ upgradeWarningToError (nfp, sh, fd) =
...
+#if MIN_GHC_API_VERSION(8,10,0)
+    , Opt_WarnUnusedRecordWildcards
+#endif
+#if MIN_GHC_API_VERSION(8,6,0)
+    , Opt_WarnInaccessibleCode
+#endif
+    , Opt_WarnWarningsDeprecations
+    ]
+
+tagDiag :: (WarnReason, FileDiagnostic) -> (WarnReason, FileDiagnostic)
```
ndmitchell, line 284: Could we get a comment on what this does and why? (From the code, its clear, but nice to be able to eyeball it without reading the code)

test/src/Development/IDE/Test.hs
```diff
@@ -76,7 +83,11 @@ expectNoMoreDiagnostics timeout = do
     ignoreOthers = void anyMessage >> handleMessages
 
 expectDiagnostics :: [(FilePath, [(DiagnosticSeverity, Cursor, T.Text)])] -> Session ()
-expectDiagnostics expected = do
+expectDiagnostics
+  = expectDiagnostics_ . map (\(fp, dd) -> (fp, map (\(ds, c, t) -> (ds, c, t, Nothing)) dd))
+
+expectDiagnostics_ :: [(FilePath, [(DiagnosticSeverity, Cursor, T.Text, Maybe DiagnosticTag)])] -> Session ()
+expectDiagnostics_ expected = do
```
ndmitchell, line 90: Ususally `foo_` would be the version of `foo` that does slightly less than `foo`, e.g. doesn't return some info. Could we call `expectDiagnostics_` something like `expectedDiagnosticsTags`?

test/exe/Main.hs
```diff
@@ -3065,6 +3083,8 @@ ifaceErrorTest = testCase "iface-error-test-1" $ withoutStackEnv $ runWithExtraF
     --   - P is being typechecked with the last successful artifacts for A.
     expectDiagnostics [("P.hs", [(DsWarning,(4,0), "Top-level binding")])
                       ,("P.hs", [(DsWarning,(6,0), "Top-level binding")])
+                      ,("P.hs", [(DsInfo,(4,0), "Defined but not used")])
```
ndmitchell, line 3086: The really important thing is that these diagnostics have the right tag associated with them. While they don't all need to be `expectDiagnostics_` (or whatever it ends up being called), a few more being that with the tag assertion would be good.

ndmitchell (approved):
Looks awesome!

src/Development/IDE/Core/Compile.hs
serras: I think that's a good idea: make those with tags "less important" but not hide them.

Implemented in the latest commit.

src/Development/IDE/Core/Compile.hs
serras: Added comment to the file:

> -- | Add a unnecessary/deprecated tag to the required diagnostics.

test/src/Development/IDE/Test.hs
serras: I've renamed it to `expectDiagnosticsWithTags`.

test/exe/Main.hs
serras: That makes sense. I've extended the checks in the latest commit.

ndmitchell:
I see some test failure:

```
FAIL (0.59s)
      test/src/Development/IDE/Test.hs:77:
      Got unexpected diagnostics for Uri {getUri = "file:///tmp/extra-dir-28584335661024/ModuleA.hs"} got List [Diagnostic {_range = Range {_start = Position {_line = 1, _character = 0}, _end = Position {_line = 1, _character = 29}}, _severity = Just DsInfo, _code = Nothing, _source = Just "typecheck", _message = "The import of \8216ModuleB\8217 is redundant\n  except perhaps to import instances from \8216ModuleB\8217\nTo import instances alone, use: import ModuleB()", _tags = Just (List [DtUnnecessary]), _relatedInformation = Nothing}]
```

I did have to fix a merge conflict to get it going again. So certainly possible that's my fault.

serras:
> I see some test failure:
> 
> ```
> FAIL (0.59s)
>       test/src/Development/IDE/Test.hs:77:
>       Got unexpected diagnostics for Uri {getUri = "file:///tmp/extra-dir-28584335661024/ModuleA.hs"} got List [Diagnostic {_range = Range {_start = Position {_line = 1, _character = 0}, _end = Position {_line = 1, _character = 29}}, _severity = Just DsInfo, _code = Nothing, _source = Just "typecheck", _message = "The import of \8216ModuleB\8217 is redundant\n  except perhaps to import instances from \8216ModuleB\8217\nTo import instances alone, use: import ModuleB()", _tags = Just (List [DtUnnecessary]), _relatedInformation = Nothing}]
> ```
> 
> I did have to fix a merge conflict to get it going again. So certainly possible that's my fault.

I've spent some time and fixed the test. Hopefully everything should be green in some minutes.

ndmitchell:
Awesome work - this is a great feature. Thanks for all your hard to work to write it and adjust so much of the test suite.

PR #816, opened 2020-09-17
Use optExtensions in Session loader

pepeiborra:
This is a minor correctness fix

### DIFF
--- a/session-loader/Development/IDE/Session.hs
+++ b/session-loader/Development/IDE/Session.hs
@@ -388,7 +388,11 @@ fromTargetId :: [FilePath]          -- ^ import paths
              -> IO [TargetDetails]
 -- For a target module we consider all the import paths
 fromTargetId is exts (TargetModule mod) env dep = do
-    let fps = [i </> moduleNameSlashes mod -<.> ext | ext <- exts, i <- is ]
+    let fps = [i </> moduleNameSlashes mod -<.> ext <> boot
+              | ext <- exts
+              , i <- is
+              , boot <- ["", "-boot"]
+              ]
     locs <- mapM (fmap toNormalizedFilePath' . canonicalizePath) fps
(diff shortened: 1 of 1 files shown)


session-loader/Development/IDE/Session.hs
```diff
@@ -380,18 +381,18 @@ data TargetDetails = TargetDetails
...
+             -> [String]            -- ^ extensions to consider
              -> TargetId
              -> IdeResult HscEnvEq
              -> DependencyInfo
              -> IO [TargetDetails]
 -- For a target module we consider all the import paths
-fromTargetId is (TargetModule mod) env dep = do
+fromTargetId is exts (TargetModule mod) env dep = do
     let fps = [i </> moduleNameSlashes mod -<.> ext | ext <- exts, i <- is ]
-        exts = ["hs", "hs-boot", "lhs"]
```
ndmitchell, line 390: Normally we set optExtensions to be `hs` and `lhs`. Before that, we were looking for `hs-boot` too. Is that behaviour deliberate? Was looking for `hs-boot` wrong? If looking for `hs-boot` wasn't wrong, why wasn't `lhs-boot` on the list?

pepeiborra: Yes, that's a good point. I think "hs-boot" needs to be added to the list

pepeiborra: I don't know what's the story with `lhs-boot`. https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/separate_compilation.html does not mention them so perhaps they are not a thinh

ndmitchell: Fair enough. I wish both boot files AND lhs files were not a thing, so the combination of them is a bit of a monster!

ndmitchell (approved):

PR #820, opened 2020-09-17
Preserve more information about targets

pepeiborra:
The problem we are trying to solve is accurately decide what modules belong to a cradle when there are multiple projects sharing the same source folder, and project membership is driven by a Bios cradle. 

For instance, say we have modules A,B,C in the same folder, and A,B should use the same set of flags whereas C should use a different set of flags. For A or B, the Bios cradle will return the set of flags F1 and the targets {A,B}. For C, the Bios cradle will return the set of flags F2 and the set of targets {C}. 

While editing A, when finding an `import C`, ghcide must not allow ghc to load `C`, and instead must locate it in a package. 

The improvement in this PR is to preserve more information from the set of targets returned by the cradle. Targets can be either file paths or modules. Previously we would only store modules, inferring possible module names from file paths targets which was lossy and unnecessary. The change here is to preserve either module names or file paths.

### DIFF
--- a/session-loader/Development/IDE/Session.hs
+++ b/session-loader/Development/IDE/Session.hs
@@ -34,7 +34,8 @@ import Data.Version
 import Development.IDE.Core.OfInterest
 import Development.IDE.Core.Shake
 import Development.IDE.Core.RuleTypes
-import Development.IDE.GHC.Compat
+import Development.IDE.GHC.Compat hiding (Target, TargetModule, TargetFile)
+import qualified Development.IDE.GHC.Compat as GHC
 import Development.IDE.GHC.Util
 import Development.IDE.Session.VersionCheck
 import Development.IDE.Types.Diagnostics
@@ -59,13 +60,12 @@ import System.IO
 
 import GHCi
 import DynFlags
-import HscTypes
+import HscTypes (ic_dflags, hsc_IC, hsc_dflags, hsc_NC)
 import Linker
 import Module
 import NameCache
 import Packages
 import Control.Exception (evaluate)
-import Data.Char
 
 -- | Given a root directory, return a Shake 'Action' which setups an
 -- 'IdeGhcSession' given a file.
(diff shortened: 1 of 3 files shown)


wz1000:
Why not use `GHC.TargetId`?

pepeiborra:
> Why not use `GHC.TargetId`?
> […](#)

I wanted to, but the phase argument is redundant

wz1000:
Can't we ignore the phase argument/set it to Nothing?

pepeiborra:
> Can't we ignore the phase argument/set it to Nothing?

We could. Why does it matter?

wz1000:
Doesn't really, just seems a bit strange to redefine a type that GHC already provides.

wz1000 (approved):

session-loader/Development/IDE/Session.hs
```diff
@@ -448,7 +437,8 @@ newComponentCache logger exts cradlePath hsc_env uids ci = do
     -- the component, in which case things will be horribly broken anyway.
     -- Otherwise, we will immediately attempt to reload this module which
     -- causes an infinite loop and high CPU usage.
-    let special_target = TargetDetails (mkModuleName "special") targetEnv targetDepends [componentFP ci]
+    let special_target = TargetDetails (TargetModule m) targetEnv targetDepends [componentFP ci]
```
wz1000, line 440: This seems strange, though I suppose it predates this PR. Can't we use `TargetFile` now?

pepeiborra: I'm not sure what's supposed to be the special target. In the original version of the code, it was simply the string "special".

wz1000: I think it might be `_cfp` from `session`, which might need to be passed into `newComponentCache` as an extra argument.

/cc @fendor

pepeiborra: Yeah, that makes sense. Fixed

PR #823, opened 2020-09-19
Enhance benchmarks & bug fixes

### LINKED ISSUE #703: Collect progressEnd timings in benchmarks
The current suite of benchmark experiments has become obsolete after the change to use delayed data in #624 

To remediate this, @wz1000 and I propose that the benchmark runner is extended to collect two timings per experiment:

1. The experiment conclusion, which is generally a user visible response (e.g. completions) and involves delayed data
2. The progressEnd notification, which includes the time taken to refresh the actual data (e.g. recompute the completions)

So, keep the current experiments, but collect more metrics.

pepeiborra:
Enhance the benchmark suite to measure user time, delayed work time and allocations.
Fixes bugs I found in the process.

Fixes #703 and is also relevant for #242

### DIFF
--- a/bench/lib/Experiments.hs
+++ b/bench/lib/Experiments.hs
@@ -239,7 +239,7 @@ runBenchmarks allBenchmarks = do
                 in (b,) <$> runBench run b
 
   -- output raw data as CSV
-  let headers = ["name", "success", "samples", "startup", "setup", "experiment", "maxResidency"]
+  let headers = ["name", "success", "samples", "startup", "setup", "experiment", "maxResidency", "allocatedBytes"]
       rows =
         [ [ name,
             show success,
@@ -247,7 +247,8 @@ runBenchmarks allBenchmarks = do
             show startup,
             show runSetup',
             show runExperiment,
-            showMB maxResidency
+            show maxResidency,
+            show allocations
           ]
           | (Bench {name, samples}, BenchRun {..}) <- results,
             let runSetup' = if runSetup < 0.01 then 0 else runSetup
@@ -266,7 +267,8 @@ runBenchmarks allBenchmarks = do
             showDuration startup,
             showDuration runSetup',
             showDuration runExperiment,
-            showMB maxResidency
+            showMB maxResidency,
+            showMB allocations
           ]
           | (Bench {name, samples}, BenchRun {..}) <- results,
             let runSetup' = if runSetup < 0.01 then 0 else runSetup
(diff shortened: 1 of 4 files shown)


bench/lib/Experiments.hs
```diff
@@ -328,27 +350,35 @@ runBench runSess Bench {..} = handleAny (\e -> print e >> return badRun)
       changeDoc doc [hygienicEdit]
       waitForProgressDone
 
-
     liftIO $ output $ "Running " <> name <> " benchmark"
     (runSetup, userState) <- duration $ benchSetup doc
-    let loop 0 = return True
-        loop n = do
+    let loop userWaits delayedWork 0 = return $ Just (userWaits, delayedWork)
```
ndmitchell, line 355: I love a good space leak :) - from a cursor read, it looks like the first two variables will be space leaks, but can't be sure - best to just shove a ! in front of all three.

bench/lib/Experiments.hs
```diff
@@ -401,11 +431,14 @@ setup = do
...
-  case find ("maximum residency" `isInfixOf`) (reverse $ lines input) of
-    Just l -> read $ filter isDigit $ head (words l)
-    Nothing -> -1
+parseMaxResidencyAndAllocations :: String -> (Int, Int)
+parseMaxResidencyAndAllocations input =
+    (f "maximum residency", f "bytes allocated in the heap")
+  where
+    inps = reverse $ lines input
+    f label = case find (label `isInfixOf`) inps of
+        Just l -> read $ filter isDigit $ head (words l)
```
ndmitchell, line 440: Fixing bracket and $ is a bit odd.

```suggestion
        Just l -> read $ filter isDigit $ head $ words l
```

Or you might want to use `word1` from Extra.

session-loader/Development/IDE/Session.hs
```diff
@@ -245,11 +245,21 @@ loadSession dir = do
...
-
-    let consultCradle :: Maybe FilePath -> FilePath -> IO ([NormalizedFilePath], (IdeResult HscEnvEq, [FilePath]))
+          -- Typecheck all files in the project on startup
+          unless (null cs || not checkProject) $ do
+                cfps' <- liftIO $ filterM (IO.doesFileExist . fromNormalizedFilePath) (concatMap targetLocations cs)
+                void $ shakeEnqueue extras $ mkDelayedAction "InitialLoad" Debug $ void $ do
+                    mmt <- uses GetModificationTime cfps'
+                    let cs_exist = catMaybes (zipWith (<$) cfps' mmt)
+                    modIfaces <- uses GetModIface cs_exist
+                    -- update xports map
```
ndmitchell, line 255:
```suggestion
                    -- update exports map
```

ndmitchell (approved):

bench/lib/Experiments.hs
pepeiborra: good catch!

pepeiborra: Although hopefully the GHC strictness analyser would figure that out as well. I wonder if that would make a good hover provider.

PR #824, opened 2020-09-20
Add tests demonstrating diagnostic errors on missing record fields

gdevanla:
Adds tests to demonstrate that diagnostic errors are produced on missing errors fields. Addresses https://github.com/haskell/haskell-language-server/issues/333

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -232,6 +232,46 @@ diagnosticTests = testGroup "diagnostics"
           , [(DsError, (2, 8), "Found hole: _ :: Int -> String")]
           )
         ]
+  , testSessionWait "missing_fields" $ do
+      let content = T.unlines
+            [ "data Person = Person {_personName:: String, _personAge:: Int}"
+            , "foo :: Person"
+            , "foo = Person {_personName=\"test\" }"
+            ]
+      _ <- createDoc "Testing.hs" "haskell" content
+      expectDiagnostics
+        [ ( "Testing.hs"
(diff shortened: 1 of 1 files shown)


pepeiborra:
I'm not sure what purpose is served by these tests. There are already general tests for diagnostics, and we certainly don't want to test every diagnostic produced by GHC.

ndmitchell:
My guess is that records and record related errors are going to get more complicated over time. We are likely to take such errors and turn them into completions, and when RecordDotSyntax lands, that will be even more true. While we shouldn't have tests for most of GHC errors, I can see the benefit for taking a bit more time with records - especially if we parse such errors as it will give us an early/easy signal when their format changes.

Is that convincing @pepeiborra ? Is that your motivation @gdevanla ?

pepeiborra:
We don't parse record diagnostics at the moment, so I remain unconvinced.

gdevanla:
I have to write these tests as I have been exploring ways to provide completions for records as part of the haskell/haskell-language-server#333 issue.  

I thought this would be a first step to make sure that I get this information from the diagnostic's step.  For the actual issues at hand, I have been exploring two options.

1.  See if I can parse the diagnotistic information, to just provide a `code snippet` of just the missing fields.
2.  Use the completion information,  investigate the AST provided by GHC and then provide intelligent convictions.  

I am still not too familiar on what either approaches would entail. But, currently, I am on that path of exploring those options..

Any pointers, to one of the 2 options will be greatly appreciated.

That said, if these tests do help in that path, I am happy to close this issue.

Thanks!

Issue #421, opened 2020-02-11
Memory usage


adamConnerSax:
I know versions of this have been reported before but I thought this might be a useful data point and maybe a way for me to get things to try...
I'm using ghcide-0.1.0.0, with ghc-8.6.5, from emacs (with recent versions of hie-bios and haskell-lsp) on Mac OSX 10.15.3. 
It works fine but memory usage (of ghcide) starts at about 5GB, grows fast and eventually (an hour of editing a few files in the project) makes things too slow.  So I kill it and then emacs asks me to restart and it all works fine for a while.  
One observation: I was using ghcide 0.0.6 and the same thing would happen but much more slowly.  I'd maybe restart once a day instead of once an hour.
Anything I can try to improve this or help debug it?

mpickering:
@adamConnerSax You want to save your bug report until the work on #355 is resolved. Perhaps you can try out that branch?

adamConnerSax:
Does it work with 8.6.5?  Then I'd be happy to.  But I am stuck with a dependency that I can't use with 8.8.

mpickering:
There is a branch which works with 8.6.5 which I am planning to work on a bit more tonight.

adamConnerSax:
If you point me to a commit/branch, I'm happy to try!

mpickering:
https://github.com/mpickering/ghcide/tree/wip/interface-files-hie-8.6.5

adamConnerSax:
That's working!  And holding at about 2.7GB.  I can't always tell if all the features are working...sometimes I get the little window in the upper left with definitions and sometimes I don't, but that was true before as well, I think.

mpickering:
@adamConnerSax In the current design no results of shake rules are ever garbage collected, so over the course of run then it will leak memory. The branch should definitely use less memory overall though, especially on subsequent runs once the interface files have been created.

LukaHorvat:
@mpickering Is that branch merged somewhere now or should I still use it if I want to test it?

pepeiborra:
Yes it is, merged to master

mpardalos:
We've spoken with @mpickering about working on this as part of my GSoC project. We would need to add some kind of garbage collection scheme for the results of the Shake rules.

Would something simple like removing all entries relating to a file when it is closed work? 
Also, would performance regressions relating to this kind of change show up in the benchmarks?

jneira:
@mpardalos hi! anything of interest about this issue? could it be closed?

[...]

PR #826, opened 2020-09-22
Memory info and more details in opentelemetry

mpardalos:
This patch adds the following information to opentelemetry traces:

* Number of entries in the "state" hashmap in ShakeExtras (Metric is called "values map size_bytes")
* Memory occupancy in that map. (This is currently very slow, and so sizes are only reported about once/minute)
* Trace of *all* LSP requests and notifications (as opposed to just those defined in HoverDefinition.hs, as before)

Other changes:
* Put opentelemetry tracing behind `--ot-profiling` flag
* Add `--ot-profiling` flag to benchmark to run ghcide with `--ot-profiling`
* Improve recording of heap_live_bytes in opentelemetry trace by running performGC every 0.5s (only with `--ot-profiling`)

### DIFF
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -445,6 +445,9 @@ shakeOpen getLspId eventer withProgress withIndefiniteProgress logger debouncer
     initSession <- newSession shakeExtras shakeDb []
     shakeSession <- newMVar initSession
     let ideState = IdeState{..}
+
+    startTelemetry "Values map" (state shakeExtras)
+
     return ideState
     where
         -- The progress thread is a state machine with two states:
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -128,6 +128,12 @@ import UniqSupply
 import PrelInfo
 import Data.Int (Int64)
 import qualified Data.HashSet as HSet
+import OpenTelemetry.Eventlog (observe, mkValueObserver, withSpan, setTag, withSpan_)
+import System.Mem (performGC)
+import HeapSize (recursiveSizeNoGC)
+import Data.Functor ((<&>))
+import Data.Function ((&))
+import Foreign (Storable(sizeOf))
 
 -- information we stash inside the shakeExtra field
 data ShakeExtras = ShakeExtras
@@ -1175,3 +1181,35 @@ updatePositionMapping IdeState{shakeExtras = ShakeExtras{positionMapping}} Versi
         pure $! HMap.insert uri updatedMapping allMappings
   where
     shared_change = mkDelta changes
+
+startTelemetry :: String -> Var Values -> IdeOTProfiling -> IO ()
+startTelemetry _ _ (IdeOTProfiling False) = return ()
+startTelemetry name valuesRef (IdeOTProfiling True) = do
+  mapBytesInstrument <- mkValueObserver (BS.pack name <> " size_bytes")
+
+  mapCountInstrument <- mkValueObserver (BS.pack name <> " count")
+  _ <- regularly 10000 $ -- 100 times/s
+    withSpan_ "Measure length" $
+      readVar valuesRef
+      >>= observe mapCountInstrument . length
+
+  _ <- regularly 500000 $
+    withSpan_ "Measure Memory" $ do
+      performGC
+      values <- readVar valuesRef
+      let groupedValues =
+              HMap.toList values
+              <&> (\((f, k), v) -> (k, [(f, v)]))
+              & HMap.fromListWith (++)
+      valuesSize <- sequence $ HMap.mapWithKey (\k v -> withSpan ("Measure " <> (BS.pack $ show k)) $ \sp -> do
+            { byteSize <- (sizeOf (undefined :: Word) *) <$> recursiveSizeNoGC v
+            ; setTag sp "size" (BS.pack $ (show byteSize ++ " bytes"))
+            ; return byteSize
+            })
+            groupedValues
+      observe mapBytesInstrument (sum valuesSize)
+  return ()
+
+  where
+    regularly :: Int -> IO () -> IO (Async ())
+    regularly delay act = async $ forever (act >> threadDelay delay)
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -1193,7 +1193,7 @@ startTelemetry shakeExtras = do
         mapBytesInstrument <- mkValueObserver ("value map size_bytes")
         mapCountInstrument <- mkValueObserver ("values map count")
 
-        _ <- regularly 10000 $ -- 100 times/s
+        _ <- regularly 50000 $ -- 100 times/s
             withSpan_ "Measure length" $
             readVar stateRef
             >>= observe mapCountInstrument . length
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -118,7 +118,9 @@ import Control.Monad.IO.Class
 import Control.Monad.Reader
 import Control.Monad.Trans.Maybe
 import Data.Traversable
+import Text.Printf
 import Data.Hashable
+import Development.IDE.Core.Tracing
 
 import Data.IORef
 import NameCache
@@ -849,7 +851,7 @@ defineEarlyCutoff
     :: IdeRule k v
     => (k -> NormalizedFilePath -> Action (Maybe BS.ByteString, IdeResult v))
     -> Rules ()
-defineEarlyCutoff op = addBuiltinRule noLint noIdentity $ \(Q (key, file)) (old :: Maybe BS.ByteString) mode -> do
+defineEarlyCutoff op = addBuiltinRule noLint noIdentity $ \(Q (key, file)) (old :: Maybe BS.ByteString) mode -> otTracedAction key file $ do
     extras@ShakeExtras{state, inProgress} <- getShakeExtras
     -- don't do progress for GetFileExists, as there are lots of non-nodes for just that one key
     (if show key == "GetFileExists" then id else withProgressVar inProgress file) $ do
--- a/src/Development/IDE/Core/Shake.hs
+++ b/src/Development/IDE/Core/Shake.hs
@@ -452,7 +452,7 @@ shakeOpen getLspId eventer withProgress withIndefiniteProgress logger debouncer
     shakeSession <- newMVar initSession
     let ideState = IdeState{..}
 
-    startTelemetry "Values map" (state shakeExtras) =<< (optOTProfiling <$> getIdeOptionsIO shakeExtras)
+    startTelemetry shakeExtras
 
     return ideState
     where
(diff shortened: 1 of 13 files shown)


wz1000:
Did you try this out on a ghcide session? Any memorable findings? Which actions/requests take the most time, which ones leak memory, which ones cause memory consumption to spike, etc?

wz1000:
It would be good to have some documentation on how to see/use/interpret these traces.

src/Development/IDE/Core/Tracing.hs
```diff
@@ -0,0 +1,77 @@
...
+        return sp
+    )
+    endSpan
+    (const act)
+
+startTelemetry :: Show k => String -> Var (HMap.HashMap k v) -> IdeOTProfiling -> IO ()
+startTelemetry _ _ (IdeOTProfiling False) = return ()
+startTelemetry name valuesRef (IdeOTProfiling True) = do
+  mapBytesInstrument <- mkValueObserver (BS.pack name <> " size_bytes")
+  mapCountInstrument <- mkValueObserver (BS.pack name <> " count")
```
pepeiborra, line 62: In addition to these global counters, it would be very useful to have per-rule counters, in order to show which rules are expensive (in items and in memory)

pepeiborra (approved):
The changes look reasonable to me, if a bit unfinished.

For instance, the benchmark suite has been extended with an option to collect telemetry details, like the length and size of the state HashMap. But nothing is being done with this data! 
- It should be added to the CSV outputs of an experiment
- It could be graphed by the `bench/hist` build system so that we can easily spot leaks

src/Development/IDE/Core/Tracing.hs
mpardalos: I agree with all the suggested changes, with one caveat: I don't think it's feasible to measure per-rule allocations with the current method, as it would be too slow, even for benchmarking use.

mpardalos:
@wz1000 
> It would be good to have some documentation on how to see/use/interpret these traces.

Should I add that in the README?

src/Development/IDE/Core/Tracing.hs
pepeiborra: Would a measure of per-rule allocations and deriving the total from that be fast enough?

mpardalos: I'm going to try implementing this and see the results. My guess is it would actually be quite a bit faster because now you are counting items only when they are inserted into the map. However now the metric becomes one of total allocations instead of live memory since you miss deletions. Would there be some way "intercepting" those so you can count the size of deleted elements?

wz1000:
> Should I add that in the README?

The `docs/` folder seems like the right place for this. You can also link there from the benchmark README.

mpardalos:
@wz1000 
> Did you try this out on a ghcide session? Any memorable findings? Which actions/requests take the most time, which ones leak memory, which ones cause memory consumption to spike, etc?

I can't say I've found anything directly interesting as this stands. Most requests seem basically instant in  my (admittedly limited) testing on the ghcide codebase. As for actions, `TypeCheck` seems to usually take long (as expected), and I have also seen `ReportImportCycles` stand out at ~1s.

Memory consumption is, as expected due to #421, constantly rising. I am hoping this will help in addressing that, however I need to find a way of getting more detailed readings to do that. Hopefully the change @pepeiborra suggested (https://github.com/haskell/ghcide/pull/826#discussion_r494643451) helps.

src/Development/IDE/Core/Tracing.hs
pepeiborra: Hmm, I don't think it's going to be easy (or possible) to observe the deletions. 

Maybe what we want is to keep the current behaviour, but instead of getting the recursive size of the whole values map, break it down by key type and get the recursive size of each group

mpardalos: So, as a test, I tried measuring allocations when objects get added to the `state` map, (https://github.com/mpardalos/ghcide/compare/opentelemetry-memory...detailed-memory-test), and it turns out I was wrong. Running a small "hover" benchmark (see code), took about 3 hours, which I think qualifies as unusable for now. 

> Maybe what we want is to keep the current behaviour, but instead of getting the recursive size of the whole values map, break it down by key type and get the recursive size of each group

That is probably much more feasible. I will give that a try as well and report back

wz1000: Why is it so slow?

pepeiborra:
@mpardalos  any updates on this?

mpardalos:
> @mpardalos any updates on this?

Sorry for the radio silence, starting final year of uni has been a bit busy. I haven't worked on this since we last spoke, but I'll try to get it ready to merge this week.

See my comment on the code review for what I think is left to do.

src/Development/IDE/LSP/LanguageServer.hs
```diff
@@ -79,14 +80,16 @@ runLanguageServer options userHandlers onInitialConfig onConfigChange getIdeStat
     -- The set of requests that have been cancelled and are also in pendingRequests
     cancelledRequests <- newTVarIO Set.empty
 
-    let withResponse wrap f = Just $ \r@RequestMessage{_id} -> do
+    let withResponse wrap f = Just $ \r@RequestMessage{_id, _method} -> otTracedHandler "Request" (show _method) $ do
             atomically $ modifyTVar pendingRequests (Set.insert _id)
             writeChan clientMsgChan $ Response r wrap f
-    let withNotification old f = Just $ \r -> writeChan clientMsgChan $ Notification r (\lsp ide x -> f lsp ide x >> whenJust old ($ r))
-    let withResponseAndRequest wrap wrapNewReq f = Just $ \r@RequestMessage{_id} -> do
+    let withNotification old f = Just $ \r@NotificationMessage{_method} -> otTracedHandler "Notification" (show _method) $
```
wz1000, line 86: Are these traces correct? The only thing they will be measuring the time required to forward the request/notification on the channel, not the time we actually spend computing.

src/Development/IDE/Core/Tracing.hs
mpardalos: I tried analyzing which entries in the map took the longest and it seems like the worst offenders are GhcSessionIO and GhcSession entries, each of which can take about 10s to measure, I think because their representation is very nested (each can be >1000000 closures). If there is a way to estimate the size of these entries instead of having to traverse them it could improve performance a lot. 

If there isn't a way to handle this then the only way to improve performance would be to optimise heapsize further.

src/Development/IDE/LSP/LanguageServer.hs
mpardalos: You are completely right. Should I be adding that further below in `handleInit`?

src/Development/IDE/Core/Tracing.hs
pepeiborra: Have you tried the group approach below? I think that's probably what we want to measure and not the individual insertions, since I suspect that when inserting a new item in the map part of its cost is shared with other entries in the map. 

> Maybe what we want is to keep the current behaviour, but instead of getting the recursive size of the whole values map, break it down by key type and get the recursive size of each group

mpardalos:
I made it report the size for each key separately (lumping together all the entries for a given key which refer to different files). This could be pretty useful although it is still slow.

Here's a screenshot of what the output might look like in tracy.

image

"values map size_bytes" is the total size of the values map, and each key gets its own graph below that.

As I mentioned before, the worst offenders for the slowdown are GhcSession and GhcSessionIO, which are also the biggest entries. I will try skipping them first (to see how much of a difference that makes) and then try putting their measurement on a separate thread so that every other entry can at least have some more detailed data.

wz1000:
@mpardalos great work, that looks awesome!

Here are some sets of rules among which a lot of sharing might be going on: `[GhcSession, GhcSessionDeps,  GhcSessionIO], [GetModSummaryWithoutTimestamps, GetModSummary], [GetModIfaceFromDisk,GetModIface, GetModIfaceWithoutLinkables], [GetParsedModule,TypeCheck]`

It would be awesome if we could group together the entries for such lists too.

Also, can you upload the full trace that you computed somewhere? I'd be very interested in looking at it.

wz1000:
What does the `gc` graph show?

wz1000:
As a way to workaround any slowness, could it be possible to take periodic heap snapshots and dump them to disk (which should be fast), and load them later to all the number crunching?

/cc @mpickering

mpardalos:
> What does the `gc` graph show?

This is part of OpenTelemetry itself. It counts total GC time.

> As a way to workaround any slowness, could it be possible to take periodic heap snapshots and dump them to disk (which should be fast), and load them later to all the number crunching?

If that's possible it could be very very interesting

> Also, can you upload the full trace that you computed somewhere? I'd be very interested in looking at it. 

Here it is. It's basically just me jumping around in ghcide's source, opening/closing files, adding syntax errors, etc. It is still slow but I think that skipping those `GhcSession(IO)?` entries brings it into the realm of usable.

[ghcide.eventlog.gz](https://github.com/haskell/ghcide/files/5403255/ghcide.eventlog.gz)

wz1000:
Some more things to measure: 

- number of entries/files for each key
- Number of stale vs current entries

pepeiborra:
This is starting to look great. Should we focus on getting a minimum useful kernel checked in so that we can iterate over it?

wz1000:
Here is one hack that might possibly work to emulate the 'snapshot' idea if you have enough swap:

```
sample :: UTCTime -> IO ProcessId
sample time = forkProcess $ do
  raiseSignal sigSTOP
  -- Do profiling
  profileResults <- ...
  dumpResultsToFile ("sample-"++show time) profileResults
```

Then we need to collect all of these `ProcessId`s (one for each sample), and before exiting `ghcide`, raise `SIGCONT` on each of these one by one and waiting for the process to exit.

However, you do have to be careful about not using any `Handle`s/`MVars`/locks from the original process when you are computing results or dumping them out. I'm pretty sure it would crash if you try to write them to the eventlog, so you need to dump it to a newly opened file.

pepeiborra:
It's starting to look like this PR is abandoned. 
@mpardalos can we get a minimum useful change merged?

mpardalos:
As it stands, would you consider this ready to merge? Functionality is there, everything is just quite slow.

pepeiborra:
As long as it is behind a flag, it's mergeable

pepeiborra:
@mpardalos I want to have a play with this, but I don't know how do I hook the event trace to Tracy. Could you add some docs or links?

wz1000:
I believe you want to use this executable: https://github.com/ethercrow/opentelemetry-haskell/blob/master/opentelemetry-extra/exe/eventlog-to-tracy/Main.hs

mpardalos:
I added a doc for how to use this functionality. 
I included how to actually build the tracy viewer because the instructions are buried in the pdf manual for tracy.

pepeiborra:
That leads to an error, what's next?

```
[nix-shell:~/scratch/ghcide]$ eventlog-to-tracy ghcide.eventlog
Converting ghcide.eventlog to ghcide.eventlog.trace.json...
dyld: Library not loaded: libcapstone.4.dylib
  Referenced from: /nix/store/ks7805136p3f308p6i2jwygv5nd0jbmb-tracy-0.7.1/bin/import-chrome
  Reason: image not found
eventlog-to-tracy: callProcess: import-chrome "ghcide.eventlog.trace.json" "ghcide.eventlog.tracy" (exit -6): failed
```

I am using the following branch: http://github.com/pepeiborra/ghcide/tree/opentelemetry

mpardalos:
Was it maybe something unclear from the instructions? Might be a good idea to specify.

src/Development/IDE/Core/Shake.hs
```diff
@@ -1170,3 +1181,51 @@ updatePositionMapping IdeState{shakeExtras = ShakeExtras{positionMapping}} Versi
...
+            withSpan_ "Measure Memory" $ do
+            performGC
+            values <- readVar stateRef
+            let groupedValues =
+                    HMap.toList values
+                    <&> (\((f, k), v) -> (k, [(f, v)]))
+                    & HMap.fromListWith (++)
+            valuesSize <- sequence $ HMap.mapWithKey (\k v -> withSpan ("Measure " <> (BS.pack $ show k)) $ \sp -> do
+                    { instrument <- instrumentFor k
+                    ; byteSize <- (sizeOf (undefined :: Word) *) <$> recursiveSizeNoGC v
```
pepeiborra, line 1211: This call to `recursiveSizeNoGC` leads to a segfault if a GC arises while we are counting, which is quite easy to reproduce.

Given the way `recursiveSizeNoGC` works this is impossible to avoid (the act of measuring the size itself allocates memory) but it could be minimised by stopping the world while measuring.

Did you consider this and/or other workarounds @mpardalos ?

pepeiborra: Another, much cheaper way to minimise the segfaults in `recursiveSizeNoGC` is to use `System.Mem.Weak` to observe GC runs, and abort the count when this happens.

@mpardalos are you accepting PRs for heapsize?

mpardalos: Yeah absolutely

pepeiborra: I have tested the `Weak` workaround and bingo, no more segfaults. So I'm going to:

* Send you a heapsize PR
* Ask you to upload a new heapsize version to Hackage 
* Commandeer this diff (I have made plenty of changes)
* Get this merged

pepeiborra: https://github.com/mpardalos/Heapsize/pull/1

pepeiborra: https://github.com/mpardalos/ghcide/pull/1

pepeiborra: I spoke too soon, segfaults can still arise if GC strikes in between the Weak check and the address dereferencing. It would be much safer if this code was written in C and accessed through an unsafe foreign import, because that guarantees that a GC cannot strike.

I also investigated a solution for the sharing problems, providing an Applicative interface so that it's possible to do multiple heapsize measurements with the same set of seen pointers. Again, the GC gets in the way - making this approach not viable sadly. One would need to rewrite the whole thing in C.

pepeiborra:
I managed to get full measurements with sharingfor a small project by setting a very large nursery (`-A4G`) and grouping key types in chunks instead of trying to measure the whole values dictionary in one go, as well as optimising the measurement code a bit to allocate less. 

However, I still observe crashes due to SIGILL (Illegal Instruction). My attempts to get a backtrace have been unsuccessful, Mac OS  got in the way. I'll try to replicate in Linux tomorrow.

Screenshot below where you can observe the effects of sharing: Typecheck, GetModIface, GetModSummary, and others have a cost of 1 because everything? is reachable from the GhcSession, apparently. Sadly not very helpful. The only exception is `GetParsedModule`, but that's accidental: it wasn't in the sharing group when I took the measurement. Once added to the sharing group, it also has cost 1.

image
image
image

wz1000:
Why does `GetHieAst` have a size of 1? I'm pretty sure its not being shared by anything else.

pepeiborra:
Superseded by #922

pepeiborra:
> Why does `GetHieAst` have a size of 1? I'm pretty sure its not being shared by anything else.

This had a simple explanation: Shake extras is reachable from the `GhcSessionIO` closure.

wz1000:
> This had a simple explanation: Shake extras is reachable from the GhcSessionIO closure.

Does that mean everything in the Shake store is reported as of `GhcSessionIO`?

pepeiborra:
> > This had a simple explanation: Shake extras is reachable from the GhcSessionIO closure.
> 
> 
> 
> Does that mean everything in the Shake store is reported as part of `GhcSessionIO`?

Yes, I would expect so

PR #827, opened 2020-09-23
Use hie-implicit-cradle-0.2.0.1

jneira:
To make progress fixing https://github.com/haskell/haskell-language-server/issues/417


ndmitchell (approved):

jneira:
@wz1000 what would be the best approach to get this bug fix in hls-3, wait for #784?

wz1000:
You can do that, or if you are impatient, you can cherry pick to hls-3.3 on `haskell/ghcide`(I think) and then bump the submodule.

PR #831, opened 2020-09-24
FileExists: set one watcher instead of thousands

### LINKED ISSUE #776: `didChangedWatchedFiles` notifications lock up emacs on startup
I observed this in HLS 0.3.0.0, but I'm pretty sure this is a ghcide issue, since this seems to be where we send the notifications.

When I start up a session in my (not ridiculously large) project, we send ~3k `didChangedWatchedFiles` notifications to the client. This locks up emacs completely until they are all processed, which it seems to do slowly, at a rate of ~10 responses a second (by tailing `hie.log`).

This is a pretty bad user experience. Clearly some of the fault is emacs' or `lsp-mode`'s: 3k files to watch is really not that much, and I don't know how it's dealing with them so badly. Maybe we can do something upstream.

However, maybe we could do something to stress it out less. Some possibilities:
- If the problem is large numbers of *messages*, we could debounce the notification queue and accumulate multiple watchers into a single `didChangeWatchedFiles` notification (it supports sending multiple watchers per notification). I suspect this would reduce the number of messages a *lot*.
- If the problem is a large number of watchers, we might be able to use glob patterns to reduce the number of watchers a lot.
    - At least we seem to watch `Foo.hs` and `Foo.lhs` separately, which could be merged.
    - Doing `<src>/**/*.hs` for all the component's source folders might be an over-estimate, but it might be worth it if it's cheaper.

michaelpj:
This is broken. It sends an illegal message before we've sent the initialize response. I got this far because (amusingly) it actually works in emacs, but the test suite complains.

I'm not sure quite how to fix this. @pepeiborra suggested sending the message in `fileExistsRules` as I'm doing, but this appears to happen too early.

I can think of a few gross ways to fix this (actually do the registration the first time `GetFileExists` runs, add some extra state to determine whether we've done it), but I thought I'd put it up to solicit ideas.

------

This prevents us from sending thousands of notifications to the client
on startup, which can lock up some clients like emacs. Instead we send
precisely one.

This has some consequences for the behaviour of the fast file existence
lookup, which I've noted in the code, alongside a description of how it
works (I spent a while figuring it out, I thought I might as well write
it down).

Fixes #776.

### DIFF
--- a/src/Development/IDE/Core/FileExists.hs
+++ b/src/Development/IDE/Core/FileExists.hs
@@ -11,30 +11,60 @@ where
 import           Control.Concurrent.Extra
 import           Control.Exception
 import           Control.Monad.Extra
-import qualified Data.Aeson                    as A
 import           Data.Binary
 import qualified Data.ByteString               as BS
 import Data.HashMap.Strict (HashMap)
 import qualified Data.HashMap.Strict as HashMap
 import           Data.Maybe
-import qualified Data.Text                     as T
 import           Development.IDE.Core.FileStore
 import           Development.IDE.Core.IdeConfiguration
 import           Development.IDE.Core.Shake
 import           Development.IDE.Types.Location
-import           Development.IDE.Types.Logger
 import           Development.Shake
 import           Development.Shake.Classes
 import           GHC.Generics
-import           Language.Haskell.LSP.Messages
-import           Language.Haskell.LSP.Types
 import           Language.Haskell.LSP.Types.Capabilities
 import qualified System.Directory as Dir
 
--- | A map for tracking the file existence
+{- Note [File existence cache and LSP file watchers]
+Some LSP servers provide the ability to register file watches with the client, which will then notify
+us of file changes. Some clients can do this more efficiently than us, or generally it's a tricky
+problem
+
+Here we use this to maintain a quick lookup cache of file existence. How this works is:
+- On startup, if the client supports it we ask it to watch some files (see below).
+- When those files are created or deleted (we can also see change events, but we don't
+care since we're only caching existence here) we get a notification from the client.
+- The notification handler calls 'modifyFileExists' to update our cache.
+
+This means that the cache will only ever work for the files we have set up a watcher for.
+So we pick the set that we mostly care about and which are likely to change existence
+most often: the source files of the project (as determined by the source extensions
+we're configured to care about).
+
+For all other files we fall back to the slow path.
+
+There are a few failure modes to think about:
+
+1. The client doesn't send us the notifications we asked for.
+
+There's not much we can do in this case: the whole point is to rely on the client so
+we don't do the checking ourselves. If the client lets us down, we will just be wrong.
+
+2. Races between registering watchers, getting notifications, and file changes.
+
+If a file changes status between us asking for notifications and the client actually
+setting up the notifications, we might not get told about it. But this is a relatively
+small race window around startup, so we just don't worry about it.
+-}
+
+-- See Note [File existence cache and LSP file watchers]
+-- | A map for tracking the file existence.
+-- If a path maps to 'True' then it exists; if it maps to 'False' then it doesn't exist'; and
+-- if it's not in the map then we don't know.
 type FileExistsMap = (HashMap NormalizedFilePath Bool)
 
--- | A wrapper around a mutable 'FileExistsMap'
+-- | A wrapper around a mutable 'FileExistsState'
 newtype FileExistsMapVar = FileExistsMapVar (Var FileExistsMap)
 
 instance IsIdeGlobal FileExistsMapVar
@@ -45,22 +75,16 @@ getFileExistsMapUntracked = do
   FileExistsMapVar v <- getIdeGlobalAction
   liftIO $ readVar v
 
--- | Modify the global store of file exists
-modifyFileExistsAction :: (FileExistsMap -> IO FileExistsMap) -> Action ()
-modifyFileExistsAction f = do
-  FileExistsMapVar var <- getIdeGlobalAction
-  liftIO $ modifyVar_ var f
-
--- | Modify the global store of file exists
+-- | Modify the global store of file exists.
 modifyFileExists :: IdeState -> [(NormalizedFilePath, Bool)] -> IO ()
 modifyFileExists state changes = do
   FileExistsMapVar var <- getIdeGlobalState state
   changesMap           <- evaluate $ HashMap.fromList changes
-
   -- Masked to ensure that the previous values are flushed together with the map update
   mask $ \_ -> do
     -- update the map
     modifyVar_ var $ evaluate . HashMap.union changesMap
+    -- See Note [Invalidating file existence results]
     -- flush previous values
     mapM_ (deleteValue state GetFileExists . fst) changes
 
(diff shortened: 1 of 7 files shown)


pepeiborra:
Yup, the spec is quite clear on this:

> In addition the server is not allowed to send any requests or notifications to the client until it has responded with an InitializeResult

To fix this, installing the watcher on the first call to `GetFileExists` is ok. A nicer way would be to change the `Rules ()` computation taken by `initialise` to `Rules (IO ())`, so that rules are allowed to return some delayed initialisation to be run on the `LSP.onStartup` handler.

https://github.com/haskell/ghcide/blob/master/src/Development/IDE/Core/Service.hs#L41-L44

https://github.com/haskell/ghcide/blob/master/src/Development/IDE/LSP/LanguageServer.hs#L135-L136

michaelpj:
> To fix this, installing the watcher on the first call to GetFileExists is ok.

I did this. It seems to be okay.

> A nicer way would be to change the Rules () computation taken by initialise to Rules (IO ()), so that rules are allowed to return some delayed initialisation to be run on the LSP.onStartup handler.

I agree this would be nicer, but it's not clear we can do this: the `Rules ()` eventually gets passed to a function from Shake, and we don't get the return value. Or at least, it seemed non-trivial to make this work.

That said, I've got quite lost trying to follow how the LSP initialization works. So perhaps there is a place we could slip in a hook for something to run post-initialize.

michaelpj:
Okay, now I have an issue where some of the tests are timing out. I'm somewhat mystified.

My initial guess was that this resulted from a TOCTOU issue with checking whether the watcher had been initialized and initializing it, which could potentially result in lots of notifications being sent. But I think I fixed that in the second commit, and things still time out. Any ideas for debugging?

pepeiborra:
There's one test failing:
```
    add missing module:                                                                                   FAIL
      Exception: Timed out waiting to receive a message from the server.
      Last message received:
      {
          "tag": "NotLogMessage",
          "contents": {
              "jsonrpc": "2.0",
              "params": {
                  "type": 4,
                  "message": "finish: kick (took 0.03s)"
              },
              "method": "window/logMessage"
          }
      }
```
Tests are parsers that recognise a specific sequence of lsp messages. A timeout means that ghcide didn't send the expected message back after waiting for long enough time. 

This test is checking that ghcide notices that a new module was created in the file system. You probably need to update the test expectations to work with your changes. 

By the way, I'm surprised that the tests below passed; they count the number of file watchers created for a file:
```
  watched files
    workspace files:                                                                                      OK (0.18s)
    non workspace file:                                                                                   OK (0.19s)
```

michaelpj:
I changed the expectations for the watched files indeed! I had a look at the other test, but it wasn't obvious what it was doing wrong (that it would fail when the others don't). I'll have another stab!

michaelpj:
I think I tracked down the problem: in the case where we didn't have cached information for a file, we would delegate to the VFS  lookup, but unlike `fileExistsSlow`, we wouldn't set `alwaysRerun`, so this result would not get invalidated.

In the old version, if we hit the slow case in `fileExistsFast`, we would register a listener, and so we would later get notifications from that, which would handle invalidation. Tricky.

I think the handling of invalidation here is maybe a bit more complicated than it needs to be. I might investigate just using `alwaysRerun` unconditionally, but I expect that could affect performance, so I'll do that in a follow-up PR.

pepeiborra:
The `fileExistsFast` rule is quite performance critical and a bit special - it doesn't need an `alwaysRerun` because `modifyFileExists` calls `deleteValue` to mutate the state used by `defineEarlyCutoff`. In other words, it bypasses the Shake machinery by abusing the implementation details of the rules defined by ghcide.

Why is this better than a cheap `alwaysRerun` with cutoff? Back in the day we didn't have a benchmark suite so it's entirely possible that it isn't any better.

pepeiborra (approved):

src/Development/IDE/Core/FileExists.hs
```diff
@@ -31,37 +32,78 @@ import           Language.Haskell.LSP.Types
...
+-- | A map for tracking the file existence.
+-- If a path maps to 'True' then it exists; if it maps to 'False' then it doesn't exist'; and
+-- if it's not in the map then we don't know.
 type FileExistsMap = (HashMap NormalizedFilePath Bool)
 
--- | A wrapper around a mutable 'FileExistsMap'
-newtype FileExistsMapVar = FileExistsMapVar (Var FileExistsMap)
+-- | The state of our file existence cache. A pair of a boolean indicating whether we have initialized the
+-- file watcher in the client yet, and a map tracking file existence.
+data FileExistsState = FileExistsState { fileExistsWatcherInitialized :: Bool, fileExistsMap :: FileExistsMap }
```
pepeiborra, line 62: Both fields should probably be strict

src/Development/IDE/Core/FileExists.hs
```diff
@@ -31,37 +32,78 @@ import           Language.Haskell.LSP.Types
...
-  FileExistsMapVar var <- getIdeGlobalState state
+  FileExistsStateVar var <- getIdeGlobalState state
   changesMap           <- evaluate $ HashMap.fromList changes
 
   -- Masked to ensure that the previous values are flushed together with the map update
   mask $ \_ -> do
     -- update the map
-    modifyVar_ var $ evaluate . HashMap.union changesMap
-    -- flush previous values
+    modifyVar_ var $ \st -> evaluate $ st{fileExistsMap=HashMap.union changesMap (fileExistsMap st)}
```
pepeiborra, line 105: This is why the fileExistsMap field needs to be strict

pepeiborra:
The CI benchmarks show a 23% performance regression - could you run them locally in a stable environment and verify the results? I suspect this is because now the fast path calls the slow path and that propagates the `alwaysRerun`

michaelpj:
Yep, I can reproduce the benchmarks being bad. It's not yet clear to me why that is, I'll do some more digging.

---

On the functional issue: my understanding is that the problem is the invalidation in the fast path. On this branch, if we have a file that's a workspace file but not covered by our watcher, we will use `fileExistsFast`, do a slow lookup, and then *cache that forever*, since we're not using `alwaysRerun`, and we will never get a client notification about changes to that file (since we're not watching it).

A simple fix would be to change the `isWorkspaceFile` guard for using `fileExistsFast` to also check if the file matches our watch pattern. That would restore the invariant that "x is handled with fileExistsFast <=> x is watched". But it would add some more checks.

However, it would be nice if we could use `alwaysRerun`, because the current logic seems racy even for files that are watched. Consider the following series of events:
- We check for the existence of X.hs
- This triggers us asking the client to watch all ".hs" files
- We do a slow lookup for X.hs, it doesn't exist.
- X.hs is created.
- The client installs the watcher we ask for.
- We check for X.hs, and get a cached answer that it doesn't exist... forever.

What I really want to say is that *if* we got the answer the slow way *then* we should recheck next time. But I don't know how to do that. `alwaysRerun` errs in the opposite direction of checking too often.

michaelpj:
The performance degrades even before the commit which uses `alwaysRerun` :/

wz1000:
Can't we register the watcher immediately after initialization?

michaelpj:
> Can't we register the watcher immediately after initialization?

I just came to this realization myself, in particular I found https://microsoft.github.io/language-server-protocol/specification#initialized, which seems to be exactly for this. I'm in the process of rewriting to use that, let's see if it works.

Not sure that will help with the performance, but it should be neater.

pepeiborra:
> Yep, I can reproduce the benchmarks being bad. It's not yet clear to me why that is, I'll do some more digging.
> 
> On the functional issue: my understanding is that the problem is the invalidation in the fast path. On this branch, if we have a file that's a workspace file but not covered by our watcher, we will use `fileExistsFast`, do a slow lookup, and then _cache that forever_, since we're not using `alwaysRerun`, and we will never get a client notification about changes to that file (since we're not watching it).
> A simple fix would be to change the `isWorkspaceFile` guard for using `fileExistsFast` to also check if the file matches our watch pattern. That would restore the invariant that "x is handled with fileExistsFast <=> x is watched". But it would add some more checks.

Yes. We were assuming that we never care about files not covered by the watcher. Unfortunately this is not the case for arbitrary cradle dependencies, like cabal descriptors and `stack.yaml` files; we'll need individual watchers for any file not covered by the glob pattern in the workspace watcher.
> 
> However, it would be nice if we could use `alwaysRerun`, because the current logic seems racy even for files that are watched. Consider the following series of events:
> 
> * We check for the existence of X.hs
> * This triggers us asking the client to watch all ".hs" files
> * We do a slow lookup for X.hs, it doesn't exist.
> * X.hs is created.
> * The client installs the watcher we ask for.
> * We check for X.hs, and get a cached answer that it doesn't exist... forever.
> 
> What I really want to say is that _if_ we got the answer the slow way _then_ we should recheck next time. But I don't know how to do that. `alwaysRerun` errs in the opposite direction of checking too often.

It is a bit racy. What's the cost of using `alwaysRerun`? Let's measure it and make an informed decision: if the cost is small do plug the gap, but if it's big the choice is less clear

pepeiborra:
I measured for the Cabal project using the branch https://github.com/pepeiborra/ghcide/tree/alwaysRerun-fileExists

The results are around 10% slower, so I think it's probably worth dropping the hack switching to `alwaysRerun`. My only concern is whether bigger projects like GHC will see a bigger impact, since the cost of `alwaysRerun` is linear in the number of possible import paths (the benchmark suite uses the Cabal codebase). I'll see about repeating the measurements with 10x the number of import source folders.

## Results
HEAD - Use `alwaysRerun`
upstream - Use the values hack instead

image
image

Row Labels | HEAD | upstream
-- | -- | --
code actions | 0.993193472 | 0.777636776
code actions after edit | 36.5799817 | 32.86563162
completions after edit | 26.90540779 | 22.91404066
documentSymbols after edit | 5.872847811 | 6.360484811
edit | 34.66888288 | 31.02675419
getDefinition | 1.381465592 | 1.322841643
hover | 0.99478149 | 1.07828978
hover after edit | 54.36070061 | 51.61647408
Grand Total | 161.7572613 | 147.9621536

pepeiborra:
With 10 additional `hs-import-dirs` entries in the Cabal file the impact is much bigger, see chart below. 

This scenario is not unrealistic in codebases like GHC, where (unless things have changed recently) there are multiple src import dirs, or ours at Facebook where dozens of small projects are loaded into ghcide, each one with its own import dir.

I think this justifies the current hack to avoid the Shake overhead.

image

michaelpj:
Okay. I've attempted a version that filters based on whether the file in question would match our watcher, and removes the `alwaysRerun`.

However, I get the same timeout issue with that branch, so I need to investigate a bit more to figure out a way to get correct behaviour.

michaelpj:
Well, turns out I was wrong about why the tests were failing, it was actually https://github.com/bubba/lsp-test/issues/77.

That was inadvertently exercising the failure mode of "the client doesn't send us change notifications even though we asked for them". This will result in us permanently caching the stale results, but this seems okay, since the whole point is to *rely* on the client for this and not do it ourselves, so there's not much we can do if the client lets us down. I've documented that, nonetheless.

That meant the tests passed, but I still needed to handle the case of workspace-but-not-watched files. I've implemented something like what I described above: we guard the fast path by a check whether the path matches the glob patterns that we are watching. This costs us a bit, but makes it more correct.

My benchmarks seem a bit inconsistent, particularly for "hover", but the general pattern seems to be that:
- My version without the glob checking (first commit) is faster than master.
- My version *with* glob checking  (second commit) is slower than the previous version, but still faster than master.

I'd appreciate a check on this, since I'm not super-confident that I'm getting the right results here.

michaelpj:
Otherwise, I think this is good to go!

michaelpj:
Ugh, the Azure benchmarks show more of a regression on this PR. They look completely different to when I run them locally :roll_eyes:

pepeiborra:
The Azure benchmarks are not reliable, because they run in a shared environment. The only reliable metric in that setting is the total allocations

pepeiborra:
Performance is fine. Total time measurements for your branch (HEAD~2 is upstream): 

image

pepeiborra (approved):
This is an awesome contribution, thanks.

michaelpj:
Thanks for helping me through the thorns :) And I'm super glad we have a benchmark suite!

mrBliss:
Thanks @michaelpj!

PR #833, opened 2020-09-26
Enable test suite to run in parallel

pepeiborra:
To run the test suite in parallel with Cabal:

> cabal test --test-options="+RTS -N"

Locally, this runs in 58s in a Xeon with 56 logical cores

Importantly, this does not change CI
(unless stack passes +RTS -N secretly)

### DIFF
--- a/test/exe/Main.hs
+++ b/test/exe/Main.hs
@@ -2106,11 +2106,12 @@ findDefinitionAndHoverTests = let
   tst (get, check) pos targetRange title = testSessionWithExtraFiles "hover" title $ \dir -> do
 
     -- Dirty the cache to check that definitions work even in the presence of iface files
-    let fooPath = dir </> "Foo.hs"
-    fooSource <- liftIO $ readFileUtf8 fooPath
-    fooDoc <- createDoc fooPath "haskell" fooSource
-    _ <- getHover fooDoc $ Position 4 3
-    closeDoc fooDoc
+    liftIO $ runInDir dir $ do
+      let fooPath = dir </> "Foo.hs"
+      fooSource <- liftIO $ readFileUtf8 fooPath
+      fooDoc <- createDoc fooPath "haskell" fooSource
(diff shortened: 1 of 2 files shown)


test/exe/Main.hs
```diff
@@ -2107,14 +2106,14 @@ findDefinitionAndHoverTests = let
   tst (get, check) pos targetRange title = testSessionWithExtraFiles "hover" title $ \dir -> do
 
     -- Dirty the cache to check that definitions work even in the presence of iface files
-    liftIO $ runInDir dir $ do
```
wz1000, line 2110: The point of this was to ensure that we have a different ghcide session, so that the shake graph is created anew using iface files.

test/exe/Main.hs
```diff
@@ -2186,7 +2185,7 @@ findDefinitionAndHoverTests = let
   aaaL14 = Position 18 20  ;  aaa    = [mkR  11  0   11  3]
   dcL7   = Position 11 11  ;  tcDC   = [mkR   7 23    9 16]
   dcL12  = Position 16 11  ;
-  xtcL5  = Position  9 11  ;  xtc    = [ExpectExternFail,   ExpectHoverText ["Int", "Defined in ‘GHC.Types’"]]
+  xtcL5  = Position  9 11  ;  xtc    = [ExpectExternFail,   ExpectHoverText ["Int", "Defined in 'GHC.Types'"]]
```
wz1000, line 2188: Why this change?

pepeiborra: `putStrLn` was crashing when trying to print the Unicode quotes (in test failures).

test/exe/Main.hs
pepeiborra: I thought it was a leftover. Reverted

wz1000 (approved):

wz1000:
Nobody warned me about memory usage...
This ate through 12G of mem and 16G of swap like it was nothing on 12 logical core machine.

Issue #803, opened 2020-09-14
Drop GHC 8.4.x support


pepeiborra:
I am thinking that we can drop 8.4.x as soon as we upload a new version to Hackage.
If anyone has concerns about this, please discuss them here

pepeiborra:
We are keen on dropping since HIE files cannot be easily back ported to 8.4.x and we cannot afford to maintain two code paths for typechecking, hover and gotodef. This is blocking the merge of changes from the hls-3 branch.

ndmitchell:
Fine by me - I have no preferences or restrictions on what you drop.

cocreature:
Fine with me, I never needed 8.4 support personally or at work. @ndmitchell was originally advocating for adding it but that was a while back.

jneira:
Support definitely droped with #834

PR #834, opened 2020-09-27
Remove 8.4 CPP

wz1000:
Do we still want to support the `ghc-lib` code paths?


pepeiborra (approved):
Nice clean up!

jneira:
There exists still the ghc84.yaml but afaiu the project is not buildable with ghc-8.4.4 after this and theprevious pr.
It would close #803.

PR #837, opened 2020-09-27
Remove duplicated completions

### LINKED ISSUE #472: Completions are duplicated if imports are
image

Both these point at GHC.Util. Would be great to avoid generating them first, but a simple `nubOrd` on showing the completions might be worthwhile too. Root cause is that I am importing the module GHC.Util twice.

botal9:
This should fix #472 

After filtering completions by the texted prefix, we'll pass only different ones.


jneira (approved):
Thanks for the fix, bonus for including tests!

ndmitchell:
Very nice work! Thanks so much.

PR #838, opened 2020-09-27
Allow to easily customise the example used for benchmarks

pepeiborra:


wz1000 (approved):

wz1000:
Cabal file needs to be updated.

wz1000:
It would be nice if we could configure the Positions (hygienicP, identifierP) too.

pepeiborra:
> It would be nice if we could configure the Positions (hygienicP, identifierP) too.

Good point, will do. Ideally those positions would be automatically created by the benchmark suite

PR #839, opened 2020-09-28
Fix code action for adding missing constraints to type signatures

jhrcek:
I'm trying to fix [this issue](https://github.com/haskell/haskell-language-server/issues/360), which is quite relevant for our codebase at work: we have many type signatures like this, for which the code action `"Add CONSTRAINT to the context of the type signature for FUNCTION"` doesn't work.

```haskell
someFunction
    :: ( WithDb env m
       , WithError m
       , Has PieceOfConfig env
       )
    => Id User
    -> Id ProjectInfo
    -> m ProjectInfo
someFunction userId projectId = do
...
```

I added 2 failing tests illustrating both issues. The current implementation seems quite brittle (based on parsing stuff with regexes), I'm trying to figure out how to make it more robust..

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -781,14 +781,13 @@ suggestFunctionConstraint ParsedModule{pm_parsed_source = L _ HsModule{hsmodDecl
               | L _ (SigD _ (TypeSig _ identifiers (HsWC _ (HsIB _ locatedType)))) <- hsmodDecls
               , any (`isSameName` T.unpack typeSignatureName) $ fmap unLoc identifiers
               ]
-          srcSpanToRange $ case locatedType of
-            -- The type signature has explicit Context
-            L _ (HsQualTy _ (L contextSrcSpan _ ) _) -> contextSrcSpan
-            -- No explicit context, return SrcSpan at the start of type sig where we can write the Context
-            L typeSigSrcSpan _ ->
-                let start = srcSpanStart typeSigSrcSpan
-                in mkSrcSpan start start
-    
+          srcSpanToRange $ case splitLHsQualTy locatedType of
+            (L contextSrcSpan _ , _) ->
+              if isGoodSrcSpan contextSrcSpan
+                then contextSrcSpan -- The type signature has explicit context
+                else -- No explicit context, return SrcSpan at the start of type sig where we can write context
+                     let start = srcSpanStart $ getLoc locatedType in mkSrcSpan start start
+
       isSameName :: IdP GhcPs -> String -> Bool
       isSameName x name = showSDocUnsafe (ppr x) == name
 
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -740,6 +740,18 @@ findTypeSignatureName t = matchRegexUnifySpaces t "([^ ]+) :: " <&> head
 
 findTypeSignatureLine :: T.Text -> T.Text -> Int
 findTypeSignatureLine contents typeSignatureName =
+  -- TODO -- issue #1: this code makes a false assumption that type signature always has the form  
+  -- `functionName ::`
+  -- But when type signature looks as follows, the splitOn doesn't split anything
+  --  => length . lines counts # of lines in the whole file
+  --  => the signature is added at the end of file
+  -- ```
+  -- functionName
+  --    :: ( Monad m
+  --       , WithDb env m
+  --       )
+  --    => m ()
+  -- ```
   T.splitOn (typeSignatureName <> " :: ") contents & head & T.lines & length
 
 -- | Suggests a constraint for a type signature for which a constraint is missing.
@@ -774,6 +786,17 @@ suggestFunctionConstraint contents Diagnostic{..} missingConstraint
         typeSignatureFirstChar = T.length $ typeSignatureName <> " :: "
         startOfConstraint = Position typeSignatureLine typeSignatureFirstChar
         endOfConstraint = Position typeSignatureLine $
+        -- Issue #2: The mExistingConstraints is based on type signature from GHC error message,
+        -- whose type signature is normalized. If we base the Range of edit on the length of context in the error message
+        -- BUT the original code has some extra characters in the constraint, some redundant characters appear after applying the fix
+        --
+        -- Example: The signature in code:          `f :: ( Monad m ) => m ()`    
+        --          The signature in error message: `f :: forall (m :: * -> *). Monad m => m ()`
+        -- The context in code           has 13 chars (stuff between `::` and `==`)
+        -- The context in error meessage has  9 chars
+        -- => The result of applying fix will have 13 - 9 = 4 extraneous characters left over from the original code
+        --                                          `f :: (Monad m, MonadIO m) m ) => m ()`
+        --                                                                    ^^^^-extraneous chars
           typeSignatureFirstChar + maybe 0 T.length mExistingConstraints
         range = Range startOfConstraint endOfConstraint
      in [(actionTitle missingConstraint typeSignatureName, [TextEdit range newConstraint])]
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -774,6 +786,17 @@ suggestFunctionConstraint contents Diagnostic{..} missingConstraint
...
         startOfConstraint = Position typeSignatureLine typeSignatureFirstChar
         endOfConstraint = Position typeSignatureLine $
+        -- Issue #2: The mExistingConstraints is based on type signature from GHC error message,
+        -- whose type signature is normalized. If we base the Range of edit on the length of context in the error message
+        -- BUT the original code has some extra characters in the constraint, some redundant characters appear after applying the fix
+        --
+        -- Example: The signature in code:          `f :: ( Monad m ) => m ()`    
+        --          The signature in error message: `f :: forall (m :: * -> *). Monad m => m ()`
+        -- The context in code           has 13 chars (stuff between `::` and `==`)
+        -- The context in error meessage has  9 chars
```
wz1000, line 796:
```suggestion
        -- The context in error message has  9 chars
```

alanz:
>  I'm trying to figure out how to make it more robust..

Use the AST :)

jhrcek:
Ok, I tried to use the AST and made the tests pass. It'll still need some work and I'll have to clean this up a lot.
I'll get back to this.

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -168,11 +168,11 @@ suggestAction dflags packageExports ideOptions parsedModule text diag = concat
     , suggestFixConstructorImport text diag
     , suggestModuleTypo diag
     , suggestReplaceIdentifier text diag
-    , suggestConstraint text diag
     , removeRedundantConstraints text diag
     , suggestAddTypeAnnotationToSatisfyContraints text diag
     ] ++ concat
-    [  suggestNewDefinition ideOptions pm text diag
+    [  suggestConstraint pm text diag
```
jhrcek, line 174: Since suggestContext now uses `pm :: ParsedModule` I moved it into the list of functions that have access to `pm`. Is it OK that I changed the priority like that just to get access to it?

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -770,15 +769,29 @@ suggestFunctionConstraint contents Diagnostic{..} missingConstraint
   | Just typeSignatureName <- findTypeSignatureName _message
   = let mExistingConstraints = findExistingConstraints _message
         newConstraint = buildNewConstraints missingConstraint mExistingConstraints
-        typeSignatureLine = findTypeSignatureLine contents typeSignatureName
-        typeSignatureFirstChar = T.length $ typeSignatureName <> " :: "
-        startOfConstraint = Position typeSignatureLine typeSignatureFirstChar
-        endOfConstraint = Position typeSignatureLine $
-          typeSignatureFirstChar + maybe 0 T.length mExistingConstraints
-        range = Range startOfConstraint endOfConstraint
```
jhrcek, line 778: The issue with the original Regex-based parsing of the source code was that
1. it could differ significantly in length (e.g. due to extraneous spaces of parens) compared to the "normalized" type signature in error message
2. The existing logic failed completely when there were newlines in the type signature, leading to the new Context being inserted at the end of file

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -770,15 +769,29 @@ suggestFunctionConstraint contents Diagnostic{..} missingConstraint
...
-     in [(actionTitle missingConstraint typeSignatureName, [TextEdit range newConstraint])]
+     in case findRangeOfContextForFunctionNamed typeSignatureName of 
+       Just range -> [(actionTitle missingConstraint typeSignatureName, [TextEdit range newConstraint])]
+       Nothing -> []
   | otherwise = []
     where
+      findRangeOfContextForFunctionNamed :: T.Text -> Maybe Range
```
jhrcek, line 777: I'm not well versed in GHC AST, I just crossed fingers and made the tests passed.
Any suggestion to get at the SrcSpan of type signature / context are welcome :)

test/exe/Main.hs
```diff
@@ -1714,16 +1733,24 @@ addFunctionConstraintTests = let
   in testGroup "add function constraint"
   [ check
     "Add `Eq a` to the context of the type signature for `eq`"
-    (missingConstraintSourceCode Nothing)
```
jhrcek, line 1717: I think that not using this Maybe indirection makes it more immediately obvious what's the difference between "before" and "after" (+ bonus: it makes the test source code few lines shorter :smile: )

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -770,15 +769,29 @@ suggestFunctionConstraint contents Diagnostic{..} missingConstraint
...
+      findRangeOfContextForFunctionNamed :: T.Text -> Maybe Range 
+      findRangeOfContextForFunctionNamed typeSignatureName = do
+          locatedType <- listToMaybe 
+              [ locatedType
+              | L _ (SigD _ (TypeSig _ identifiers (HsWC _ (HsIB _ locatedType)))) <- hsmodDecls
+              , any (`isSameName` T.unpack typeSignatureName) $ fmap unLoc identifiers
+              ]
+          srcSpanToRange $ case locatedType of
+            -- The type signature has explicit Context
+            L _ (HsQualTy _ (L contextSrcSpan _ ) _) -> contextSrcSpan
```
pepeiborra, line 786: Not every type with a context is rooted in an `HsQualTy` constructor. Did you consider using `splitLHsQualTy` instead ?

src/Development/IDE/Plugin/CodeAction.hs
pepeiborra: I'm no expert either, but the code looks fine to me

pepeiborra (changes requested):
Thank you for your contribution!
Apologies for the long delay in reviewing this, change looks fine to me with a minor change request.

src/Development/IDE/Plugin/CodeAction.hs
jhrcek: Good point, I wasn't aware a function like that existed.
I just tried it but it makes the test case where there's no initial context to start with fail.

Looking at how [splitLHsQualTy](https://hackage.haskell.org/package/ghc-8.10.1/docs/src/GHC.Hs.Types.html#splitLHsQualTy) is implemented it's clear why: It uses `noLHsContext = noLoc []` in the case where I use `SrcLoc` of the start of the type signature. So in that respect my "manual" pattern match seems better.

On the other hand `splitLHsQualTy` has an additional case `HsParTy` which I have no idea what kind of syntax it represents. I'll try to find out and tweak my implementation based on that (Ideally adding a test case which triggers that case).

jhrcek: I didn't find an example syntax which would trigger that `HsParTy` case, but I reimplemented my attempt like this, which made the tests pass

```haskell
          srcSpanToRange $ case splitLHsQualTy locatedType of
            (L contextSrcSpan _ , _) ->
              if isGoodSrcSpan contextSrcSpan
                then contextSrcSpan -- The type signature has explicit context
                else -- No explicit context, return SrcSpan at the start of type sig where we can write context
                     let start = srcSpanStart $ getLoc locatedType in mkSrcSpan start start
```

BUT it seems to be noticably slower:

Screenshot from 2020-10-04 16-14-14

compared to my previous impl:

Screenshot from 2020-10-04 16-14-01

Which implementation would you suggest using?

jhrcek: Nevermind, the speed difference seems to have been fluke. On repeated reruns both versions seems to perform the same. So I'm going with your suggestion of using the ghc helper function.

pepeiborra (approved):

Issue #107, opened 2019-09-20
Warnings emitted by the desugarer are not shown


maralorn:
I sometimes have some enabled warnings which don‘t get displayed even if all other warnings are and there are no errors in the code. A cabal run on the same file shows the warning. My feeling is it depends on the category of the warning. I noticed the problem with following categories:

* `all-missed-specialisations`
* `incomplete-patterns`

There might be more. I’d like to know if this warnings work for anybody else.

ndmitchell:
These warnings are generated by the desugarer. We only run as far as the type checker. We could go onwards, and have the code to do so, but it costs more CPU time. Should probably be added as a low priority rule and discard the output (to save memory).

cocreature:
IIRC GHC shows these warnings even if you pass `-fno-code`. Probably worth looking at what code path triggers the warnings here.

suhdonghwi:
So currently there is no workaround to make ghcide to emit those warnings?

emlautarom1:
Does this mean that no warnings work? Currently using `-Wall` and `-Wincomplete-record-updates` on a project and I don't get any warnings at all.

ndmitchell:
@emlautarom1 no, not by a very long way, 90%+ of warnings show up, just those about pattern matches. If you are seeing no warnings, its not because of this issue.

emlautarom1:
@ndmitchell Then please, take a look at #524 since I'm getting no warnings when using `-Wall` and `-Wincomplete-record-updates`. This is the only reason I keep using `ghcid` instead of `ghcide`

jneira:
@maralorn do you continue missing those warnings?

maralorn:
@jneira On

```
haskell-language-server --version
haskell-language-server version: 0.4.0.0 (GHC: 8.8.4) (PATH: /nix/store/ri34qr8kzx4qwwg3yw4sj0skv738723v-haskell-language-server-0.4.0.0/bin/haskell-language-server)
```

yes.

```
{-# OPTIONS_GHC -Wall #-}
data A = A | B
f :: A ->  ()
f A = ()
```
This file shows a warning when compiled, but is clean in the editor.

maralorn:
(Which is an especially important warning in my opinion. For example [this ghc proposal](https://github.com/ghc-proposals/ghc-proposals/pull/351) wants to make incompleteness an error via an extension.)

jneira:
[...]

PR #836, opened 2020-09-27
Use object code for Template Haskell

wz1000:
Rework of https://github.com/mpickering/ghcide/pull/43/

Should improve performance and also functions as a workaround for #614 

We now generate object code along with the interface file. Object code is only generated for a module if

1) The module uses TH
2) Any parents of the module use TH

For this to work, we require the complete module graph. Earlier, tests were failing because `TargetFile`'s were not included in the set of known files. I fixed this, but now tests are failing because the modules don't define a `main` function. Any idea why this is happening @pepeiborra? See my last commit.

FIxes #614 
Fixes #107

PR #840, opened 2020-09-28
Run desugarer (not working)

georgefst:
A very rudimentary first attempt at solving #107, which is the main thing I still miss from HIE.

This is my first foray in to the internals of Ghcide, and, frankly, I don't know what I'm doing... Although it was marked "good first issue", so I guess I can't be too far off.

Hopefully someone can tell me why this doesn't yet work - I'd expect to see an "incomplete pattern match" warning on:

```hs
f = \case
    True -> ()
```

### DIFF
--- a/src/Development/IDE/Core/RuleTypes.hs
+++ b/src/Development/IDE/Core/RuleTypes.hs
@@ -109,15 +109,17 @@ data HieAstResult
   , refMap :: !RefMap
   , importMap :: !(M.Map ModuleName NormalizedFilePath) -- ^ Where are the modules imported by this file located?
   }
- 
+
 instance NFData HieAstResult where
     rnf (HAR m hf rm im) = rnf m `seq` rwhnf hf `seq` rnf rm `seq` rnf im
- 
+
 instance Show HieAstResult where
     show = show . hieModule
 
 -- | The type checked version of this file, requires TypeCheck+
 type instance RuleResult TypeCheck = TcModuleResult
+
+type instance RuleResult Desugar = DesugaredModule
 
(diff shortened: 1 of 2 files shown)


wz1000:
Notice that we already have `generateCoreRule` which does exactly this.

However, I plan to address this in #836

wz1000:
BTW, the reason this doesn't work is because you don't actually call the rule from anywhere :)

georgefst:
> BTW, the reason this doesn't work is because you don't actually call the rule from anywhere :)

Ha, I did wonder about that. Although that would seem to mean that `typeCheckRule`, which this was based on, isn't called from anywhere either. 😕

> Notice that we already have `generateCoreRule` which does exactly this.

At a glance, the implementation looks quite different - does `generateCoreRule` go further?

> However, I plan to address this in #836

Wow that's weird timing - I'd been meaning to look in to this all year...

wz1000:
> Ha, I did wonder about that. Although that would seem to mean that `typeCheckRule`, which this was based on, isn't called from anywhere either. confused

`typecheckRule` and `generateCoreRule` merely define the shake rule. They are called when we invoke `use Typecheck` or `use GenerateCore`. So to fix your code, you would need to `use_ Desugar` somewhere.

`generateCoreRule` is essentially calling `compileModule`, which calls `desugarModule` as you did, but also optionally runs the simplifier, and also tidies the Core produced using `tidyProgram` (which makes the Core suitable for further use for compilation/bytecode generation).

georgefst:
> you would need to `use_ Desugar` somewhere

Aha, of course - thanks

PR #845, opened 2020-10-03
Pull in local bindings

isovector:
PR as requested from https://github.com/haskell/haskell-language-server/pull/431


src/Development/IDE/Spans/LocalBindings.hs
```diff
@@ -45,13 +53,42 @@ bindings refmap = Bindings $ L.foldl' (flip (uncurry IM.insert)) mempty  $ do
        , unitNameEnv name (name,ty)
        )
 
+------------------------------------------------------------------------------
+-- | Compute which identifiers are in scope at every point in the AST. Use
+-- 'getDefiningBindings' to find the results.
+bindSites
+    :: RefMap
+    -> IntervalMap Position (NameEnv (Name, Maybe Type))
```
wz1000, line 61: I think it would be good to combine this with the local bindings traversal. If not, at least use `foldl'` instead of `foldr`

isovector: Yeah probablly. I didn't have the mental energy to figure out how to navigate the do block for the two cases though. Will think about it further tomorrow.

isovector:
Sorry for the delay on this; got tangled up on tactics. PTAL.

wz1000 (approved):
Looks good. This should be good to merge after a rebase to disable the flaky windows tests and applying the hlint suggestions.

isovector:
Merged, but I'm not sure how to run hlint.

wz1000:
You can run `./fmt.sh`, but the only failing hlint is for `LambdaCase`, which is already a default extension in the cabal file, so you don't need to add it as a pragma.

isovector:
Done, thanks for the guidance here @wz1000!

isovector:
CI is green!

PR #846, opened 2020-10-04
Rerun Windows tests just like linux tests

pepeiborra:
1. Rerun only the failing tests
2. Rerun only the failing tests and produce an LSP log for diagnosis


jneira:
oh right, I forgot that nice option

jneira (approved):

PR #848, opened 2020-10-05
Downgrade file watch debug log to logDebug from logInfo

mpickering:
This gets quite noisy when cabal is building dependencies which makes it
hard to see what's going on.


jneira (approved):

wz1000:
rebase to disable the windows tests.

PR #853, opened 2020-10-07
Add test for th link failure

wz1000:


pepeiborra (approved):
We really need to make this test pass before release.

wz1000:
I won't have any time to look at this any further before next weekend. 

Swapping over to bytecode should fix it, along with calling `unload` before typechecking.

To swap over to bytecode, the changes in Compile.hs from https://github.com/haskell/ghcide/pull/836/commits/dc8526a70f752afb3fc2a615a9a04257193816b2 should be enough

For the call to unload, see https://github.com/wz1000/ghcide/commit/39d3fff427138be179e533b2779d05564671596c  from https://github.com/wz1000/ghcide/tree/reload-th
If we are using bytecode, we don't need to call `unloadObj`. However, the implementation could be better, by figuring out which modules are "stable", so don't need to be unloaded.

The other commit in the reload-th branch is also needed for the correctness of recompilation checking. It does need to be adapted for bytecode though.

Note that I can reproduce the test failure from this PR even before we started using object code.

PR #861, opened 2020-10-10
Locate Documentation html page using Haddock interface file

DunetsNM:
A WIP for #860 , not ready for review

So far can make it work only for GHC 8.10.2 , having issues with `haddock-api` in other versions.


wz1000:
This seems like the correct approach, but I'm a bit concerned about reading the interface file on each hover. Perhaps we should have a global `IORef` that maintains a cache of the `ifLinkEnv`s for all the interface files we have already read.

DunetsNM:
afaik tooltip is not generated per mouse hover but once per opened .hs file/module. I was concerned too that reading all interface files in advance can be a bottleneck but it looked fast enough to me, I didn't profile memory though, should be OK as it retains only paths to html doc files in memory not the whole interface data structure

pepeiborra:
> afaik tooltip is not generated per mouse hover but once per opened .hs file/module. I was concerned too that reading all interface files in advance can be a bottleneck but it looked fast enough to me, I didn't profile memory though, should be OK as it retains only paths to html doc files in memory not the whole interface data structure

Neither. `mkDocMap` is called on every keystroke, so I would prefer caching the results of `readInterfaceFile` in a Shake rule.  In case of doubt we have a benchmark suite: just run `cabal bench` and look at the results

wz1000:
I don't think we need a shake rule, since nothing has a 'dependency' on this, and nothing has to be recomputed if we read in an interface file. An `IORef` or a `Var` should be sufficient.

DunetsNM:
accidentally closed because of force-push, reopening

DunetsNM:
is it OK to cache locations of all names from haddock interface files, or only names that are actually used in the code
Former is simple and needs only one readInterfaceFile per package but requires more memory that will be mostly unused. 
Only used names will result in multiple reads of same file when code is being typed and more names brought into scope.

wz1000:
I think we should cache the complete `LinkEnv`

DunetsNM:
@wz1000  do you mind if I partially revert #875 to bring GhcMonad back to Documentation?  I need it to use`H.nameCacheFromGhc` deep inside - the doc lookup doesn't work with `H.freshNameCache` 

As Session.hs already mentions it's probably because "It's important to keep the same NameCache though for reasons that I do not fully understand" - and neither do I

wz1000:
You can use the NameCache in ShakeExtras?

DunetsNM:
`haddock-library` is designed to compile across multiple versions of GHC, but `haddock-api` is tightly bound to a particular version of GHC - and of `haddock-library`:

8.6:   haddock-api-2.22.*  haddock-library-1.7.*
8.8:   haddock-api-2.23.*  haddock-library-1.8.*
8.10: haddock-api-2.24.*  haddock-library-1.9.*

There are two ways to solve it:

1. variate haddock-library based on GHC version.   It works but requires a lot of #ifdef-ing (especially in tests) and can result in versioning hell on the long run, and also locks out older GHC from getting improvements of newer versions.

1. or, link `haddock-api` to version `haddock-library` that it needs but use one fixed `haddock-library` version for everything else (we use 1.8.0 at the moment)

I like the second option more, because haddock-api usage is very limited but had to resort to a nasty hack as I couldn't find a clean way to use two versions of same package in stack.yaml. Is it possible to do at all ?

DunetsNM:
Also 8.10-windows build is flaky now (I saw it red and green with no code change, with both compilation or test errors)
Could be also related to `haddock-library-1.9.0` so another reason to keep it at 1.8.0, just need to find a clean way to do it.

Apart that it's ready for review

src/Development/IDE/Core/Shake.hs
```diff
@@ -62,8 +62,9 @@ module Development.IDE.Core.Shake(
     DelayedAction, mkDelayedAction,
     IdeAction(..), runIdeAction,
     mkUpdater,
+    LinkEnvsCache,
     -- Exposed for testing.
-    Q(..),
+    Q(..),
```
jneira, line 67: trimming whitespace needed

jneira: the project has a `.editorconfg` that enforces that, do you have installed the needed plugin in your editor to honour it?
It saves a lot of time fixing those little things automatically.

DunetsNM: thanks for the hint

stack-ghc-lib.yaml
```diff
@@ -14,8 +14,11 @@ extra-deps:
 - shake-0.18.5
 - regex-base-0.94.0.0
 - regex-tdfa-1.3.1.0
-- haddock-library-1.8.0
-- haddock-api-2.23.1
+- haddock-library-1.7.0
+- github: haskell/haddock
+  commit: 1f704e30d5f40fa5113045b252c0e48aebd842be
+  subdirs:
+    - haddock-api
```
pepeiborra: Why and how long do we need this custom version of haddock-api? I'm not keen on merging as it will prevent a Hackage release

pepeiborra:
> There are two ways to solve it:
> 
> 1. variate haddock-library based on GHC version.   It works but requires a lot of #ifdef-ing (especially in tests) and can result in versioning hell on the long run, and also locks out older GHC from getting improvements of newer versions.
> 2. or, link `haddock-api` to version `haddock-library` that it needs but use one fixed `haddock-library` version for everything else (we use 1.8.0 at the moment)
> 
> I like the second option more, because haddock-api usage is very limited but had to resort to a nasty hack as I couldn't find a clean way to use two versions of same package in stack.yaml. Is it possible to do at all ?

Which of these two options is compatible with uploading a package to Hackage that builds with GHC 8.6.x, 8.8.x and 8.10.x?

stack-ghc-lib.yaml
DunetsNM: this was removed from stack-ghc-lib.yaml -  haddock-api uses ghc and can't be used from ghc-lib mode anyway (still uses old documentation lookup).

similar workaround with github dependency still there in stack-810.yaml:  haddock-api-2.24.0 from hackage doesn't compile with haddock-library-1.9.0 although it should.  
Same for 8.6 (stack.yaml)

This is easy to fix by making a new release of haddock-api upstream, but I'm not sure if that's the best way to go because it also will lock us with separate versions of `haddock-library` per GHC version

DunetsNM:
> Which of these two options is compatible with uploading a package to Hackage that builds with GHC 8.6.x, 8.8.x and 8.10.x?

First option is definitely feasible, but according to haddock's README the second option looks more correct:

> `haddock-library`: is concerned with the parsing and processing of the
>    Haddock markup language. Unlike the other packages, it is expected to build
>    on a fairly wide range of GHC versions.
> 

First is easy to do:  just need to release latest commits of upstream `haddock` branches ghc-8.6 and ghc-8.10 to hackage and use them here in ghcide (plus some #ifdef-ing here and there, which can get worse over time).

Second option also requires upstream release of mentioned `haddock-api` branches , plus some way to compile them with their private copies of `haddock-library`.  
I just forked haddock and renamed `haddock-library` to `haddock-library-sandbox` which did the trick but is obviously bad (see stack.yaml). 
I'm not aware of any clean way to do it, maybe putting it into yet another package can do it e.g. haddock-api-wrapper that exposes only `readInterfaceFile` function? Any hints are very welcome. 

@pepeiborra

jneira:
@DunetsNM I would go to use each combination of `haddock-api` and `haddock-library` official versions in the cabal file

```
haddock-api ^= 2.22 || ^= 2,23 || ^= 2.23
haddock-library ^=1.7 || ^=1.8 || ^=1.9
```

Cabal should pick the correct version based on  base library bounds.

And create a special module `Development.IDE.Haddock.Compat` similar to the existing[ `Development.IDE.GHC.Compat`](https://github.com/haskell/ghcide/blob/master/src/Development/IDE/GHC/Compat.hs) with all the cpp conditions so the rest of the code can be keep clean of them.

Not sure if i fully understand your proposals but i think it is your first option. Forks from official libraries should be avoided imo, moreover if they are established and widely used ones.

A new package (`haddock-compat`?) with `Development.IDE.Haddock.Compat` with other name could be useful outside ghcide but i would start putting it here.

DunetsNM:
> Not sure if i fully understand your proposals but i think it is your first option.

@jneira correct, this is my first option. OK I'll proceed with it then. 
Converting PR back to draft meanwhile: as of now haddock-api-2.22.0 and 2.24.0 that are on hackage simply fail to compile, I'll try my best but it can take a while before fixed version is released upstream (as I understood all dependencies for ghcide must be on hackage)

pepeiborra:

```
git fetch ghcide
```
Thanks!

DunetsNM:
alright, will take a look after the holidays. This PR is still blocked by upstream haddock which seems to lack ownership so not sure when that can be merged anyway.

PR #863, opened 2020-10-11
Use -fexternal-interpreter and WayDyn, fix recompilation checking for object files

### LINKED ISSUE #854: Link error when modifying TH
When modifying TH, it is easy to run into linker errors:

```
Program error: 
ByteCodeLink.lookupCE
During interactive linking, GHCi couldn't find the following symbol:
  fakezuuid_THA_th_closure
This may be due to you not asking GHCi to load extra object files,
archives or DLLs needed by your current session.  Restart GHCi, specifying
the missing library using the -L/path/to/object/dir and -lmissinglibname
flags, or simply by naming the relevant files on the GHCi command line.
Alternatively, this link failure might indicate a bug in GHCi.
If you suspect the latter, please send a bug report to:
  glasgow-haskell-bugs@haskell.org
```

Test here: https://github.com/haskell/ghcide/pull/853

wz1000:
Needs a lot of tests, especially on windows

Fixes #854 and #672 
No need for #858

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -275,7 +275,9 @@ generateObjectCode hscEnv summary guts = do
                                 fp
                       compileFile session' StopLn (outputFilename, Just (As False))
               let unlinked = DotO dot_o_fp
-              let linkable = LM (ms_hs_date summary) (ms_mod summary) [unlinked]
+              -- Need time to be the modification time for recompilation checking
+              t <- liftIO $ getModificationTime dot_o_fp
+              let linkable = LM t (ms_mod summary) [unlinked]
               pure (map snd warnings, linkable)
 
 generateByteCode :: HscEnv -> ModSummary -> CgGuts -> IO (IdeResult Linkable)
@@ -479,7 +481,7 @@ loadModuleHome
     -> HscEnv
     -> HscEnv
 loadModuleHome mod_info e =
-    e { hsc_HPT = addToHpt (hsc_HPT e) mod_name mod_info }
+    e { hsc_HPT = addToHpt (hsc_HPT e) mod_name mod_info, hsc_type_env_var = Nothing }
     where
       mod_name = moduleName $ mi_module $ hm_iface mod_info
 
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -31,9 +31,9 @@ module Development.IDE.Core.Compile
   , setupFinderCache
   , getDocsBatch
   , lookupName
+  , modifyPLS
   ) where
 
-import Debug.Trace
 import Development.IDE.Core.RuleTypes
 import Development.IDE.Core.Preprocessor
 import Development.IDE.Core.Shake
@@ -94,8 +94,8 @@ import Control.Exception (evaluate)
 import Exception (ExceptionMonad)
 import TcEnv (tcLookup)
 import Data.Time (UTCTime)
-import Linker (unload, showLinkerState)
-import GHCi (unloadObj)
+import Linker (unload)
+-- import GHCi (unloadObj)
 import LinkerTypes
 import Control.Concurrent.MVar
 
@@ -163,16 +163,7 @@ tcRnModule pmod = do
   let ms = pm_mod_summary pmod
   hsc_env <- getSession
   let hsc_env_tmp = hsc_env { hsc_dflags = ms_hspp_opts ms }
-  traceM "******* UNLOADING ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^"
-  let pdflags = (hsc_dflags hsc_env_tmp) {log_action = \df _ _ _ st m -> hPutStrLn stderr $ renderWithStyle df m st}
-  liftIO $ showLinkerState (hsc_dynLinker hsc_env_tmp) pdflags
-  liftIO $ modifyPLS (hsc_dynLinker hsc_env) $ \pls@PersistentLinkerState{..} -> do
-    mapM_ (unloadObj hsc_env_tmp) [f | lnk <- objs_loaded, DotO f <- linkableUnlinked lnk]
-    pure (pls,())
   liftIO $ unload hsc_env_tmp []
-  liftIO $ showLinkerState (hsc_dynLinker hsc_env_tmp) pdflags
-  traceM "******* UNLOADING Over ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^"
-  traceShowM ("1Typechecking", pmod, showSDocUnsafe $ pprHPT $ hsc_HPT hsc_env_tmp)
   (tc_gbl_env, mrn_info)
         <- liftIO $ hscTypecheckRename hsc_env_tmp ms $
                        HsParsedModule { hpm_module = parsedSource pmod,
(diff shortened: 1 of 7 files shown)


pepeiborra:
In my experience the external interpreter doesn't work well with ghcide out of the box, see https://github.com/haskell/haskell-language-server/issues/433

Does this work in anger with a project that uses TH?

src/Development/IDE/Core/Compile.hs
```diff
@@ -148,11 +153,17 @@ typecheckModule (IdeDefer defer) hsc pm = do
...
+modifyPLS dl f =
+  modifyMVar (dl_mpls dl) (fmapFst pure . f . fromMaybe undefined)
+  where fmapFst f = fmap (\(x, y) -> (f x, y))
+
 tcRnModule :: GhcMonad m => ParsedModule -> m TcModuleResult
 tcRnModule pmod = do
   let ms = pm_mod_summary pmod
   hsc_env <- getSession
   let hsc_env_tmp = hsc_env { hsc_dflags = ms_hspp_opts ms }
+  liftIO $ unload hsc_env_tmp []
```
pepeiborra, line 166: IIUC:

- In 8.8.x unload is buggy and can trigger the crash outlined in GHC #16525
- In 8.10.x unload does nothing and we rely on the system linker loading new objects with a newest-one-wins behaviour. But will it unload old ones? Probably not
- In 9.2 (or hopefully 9.0) unload will be fixed and reenabled and all will be good

wz1000: Yes, your understanding is largely correct.

> In 8.10.x unload does nothing

It also clears the `lookupSymbol` cache for the external interpreter, and also clears the `PersistentLinkerState` (mainly the `objs_loaded` field which is used by `getLinkDeps` to figure out which linkables (from the `HomePackageTable`) we need to load.

wz1000:
@pepeiborra  your observation was correct, an untimely cancellation can indeed get kill the `iserv` process and leave us unable to execute any TH. 

The cause of this is the following code in `GHCi`:

```haskell
  read/write/callPipe iservPipe put
    `catch` \(e :: SomeException) -> handleIServFailure iserv e

handleIServFailure :: IServ -> SomeException -> IO a
handleIServFailure IServ{..} e = do
  ex <- getProcessExitCode iservProcess
  case ex of
    Just (ExitFailure n) ->
      throw (InstallationError ("ghc-iserv terminated (" ++ show n ++ ")"))
    _ -> do
      terminateProcess iservProcess
      _ <- waitForProcess iservProcess
      throw e
```

This calls `terminateProcess` on `ghc-iserv` on any exception, including an `AsyncCancelled`.

My last commit adds `uninteruptibleMask` to a few key places using ghc `Hooks`. I think this covers all places where an asynchronous exception might hurt us, except for `runRemoteModFinalizers` which is responsible for running finalizers added through [`addModFinalizer`](https://hackage.haskell.org/package/template-haskell-2.16.0.0/docs/Language-Haskell-TH-Syntax.html#v:addModFinalizer)

Another option is to add a `uninteruptibleMask` around all of `tcRnModule` (which doesn't seem like a very good idea). We could only do this if the module uses TH.

The third option is to use [`createIservProcessHook`](https://hackage.haskell.org/package/ghc-8.10.1/docs/Hooks.html#v:createIservProcessHook) to return a dummy process handle to ghc, so GHC can't actually kill the iserv process. I tried using `{ createIservProcessHook = Just $ \cp ->  createProcess cp >> spawnCommand "true" }` but this doesn't seem to work (I guess the communication between `ghc-iserv` and `ghc` is being disrupted).

wz1000:
I'm testing using the following declaration:

```haskell
th_ff = do
    liftIO $ threadDelay 1000000
    [d| b = () |]
```

wz1000:
Forcing `GHC` to start a new `iserv` process by writing `Nothing` to `hsc_iserv` results in segfaults while trying to evaluate any `iserv` commands.

src/Development/IDE/Core/Compile.hs
```diff
@@ -148,11 +160,55 @@ typecheckModule (IdeDefer defer) hsc pm = do
...
+         ; simpl_expr <- simplifyExpr dflags ds_expr
+
+           {- Tidy it (temporary, until coreSat does cloning) -}
+         ; let tidy_expr = tidyExpr emptyTidyEnv simpl_expr
+
+           {- Prepare for codegen -}
+         ; prepd_expr <- corePrepExpr dflags hsc_env tidy_expr
+
+           {- Lint if necessary -}
+         ; lintInteractiveExpr "hscCompileExpr" hsc_env prepd_expr
```
pepeiborra, line 195: skip?

wz1000: I guess we can. I copied the entire function verbatim from `GHC` except for the `uninteruptibleMask` around `linkExpr`.

The function is pretty harmless anyway:

```haskell
lintInteractiveExpr what hsc_env expr
  | not (gopt Opt_DoCoreLinting dflags)
  = return ()
```

src/Development/IDE/Core/Compile.hs
```diff
@@ -148,11 +160,55 @@ typecheckModule (IdeDefer defer) hsc pm = do
...
+
+           {- Convert to BCOs -}
+         ; bcos <- coreExprToBCOs hsc_env
+                     (icInteractiveModule (hsc_IC hsc_env)) prepd_expr
```
pepeiborra, line 199: I'm confused. Aren't we using object code now?

wz1000: TH is always run by compiling to bytecode. This function is copied almost exactly from GHC.

src/Development/IDE/Core/Compile.hs
```diff
@@ -479,7 +534,7 @@ loadModuleHome
     -> HscEnv
     -> HscEnv
 loadModuleHome mod_info e =
-    e { hsc_HPT = addToHpt (hsc_HPT e) mod_name mod_info }
+    e { hsc_HPT = addToHpt (hsc_HPT e) mod_name mod_info, hsc_type_env_var = Nothing }
```
pepeiborra, line 537: what's this change for?

wz1000: While debugging, I was just making sure everything was consistent with the way `GHC` was doing things: https://hackage.haskell.org/package/ghc-8.10.1/docs/src/GhcMake.html#local-6989586621681025499

pepeiborra (commented):
It would be great if @bgamari or @mpickering can take a look too

pepeiborra:
You have tested cancellation. Any thoughts about concurrency?

wz1000:
> Any thoughts about concurrency?

Shouldn't be a problem since `withIServ` uses the `hsc_iserv` `MVar` as a lock, using `takeMVar` to acquire the lock and `putMVar` to release it.

Note that we are *not* safe from cancelling if [addModFinalizer](https://hackage.haskell.org/package/template-haskell-2.16.0.0/docs/Language-Haskell-TH-Syntax.html#v:addModFinalizer) is used.

wz1000:
I don't think the approach taken in the PR is a very good idea. A deadlock, or a loop while evaluating TH will leave the IDE in an unkillable state.

wz1000:
I have removed the masking code. Now we are using `createIservProcessHook` to install a cleanup action for the iserv, so that things can run when it is restarted.

wz1000:
GHC <= 8.8 doesn't export many things we need :(

We will have to use bytecode for those.

wz1000:
Also, seems like windows doesn't support dynamic linking (or at least doesn't ship/build with --dynamic-too). @jneira can you confirm.

If this is the case, we will need to use bytecode for windows too :(

jneira:
If the wiki page about is up-to-date it definitely doesnt support it: https://gitlab.haskell.org/ghc/ghc/-/wikis/windows-dynamic-linking

wz1000:
Closed in favour of #873 

I've opened https://gitlab.haskell.org/ghc/ghc/-/issues/18852 to track the GHC issues surfaced by this PR.

PR #864, opened 2020-10-11
Run benchmarks on a list of examples 

pepeiborra:
- Cabal 3.0.0.0 
- haskell-lsp-types 0.22.0.0


pepeiborra:
@wz1000 let's get this reviewed and merged for convenient TH benchmarks

wz1000 (approved):

wz1000:
Looks good, thanks.

As a minor quality of life improvement, can we have the results interleaved so they are easy to compare?

Also, if we use `column -s, -t < results.csv` instead of cat to display the results in CI, we will get a nice aligned table.

PR #867, opened 2020-10-12
Add code action for remove all redundant imports

berberman:
This PR implements https://github.com/haskell/haskell-language-server/issues/339. The code removing single redundant import can be reused.

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -90,7 +90,8 @@ codeAction lsp state (TextDocumentIdentifier uri) _range CodeActionContext{_diag
     contents <- LSP.getVirtualFileFunc lsp $ toNormalizedUri uri
     let text = Rope.toText . (_text :: VirtualFile -> Rope.Rope) <$> contents
         mbFile = toNormalizedFilePath' <$> uriToFilePath uri
-    (ideOptions, parsedModule, join -> env) <- runAction "CodeAction" state $
+    allDiagnostics <- fmap (\(_, _, d) -> d) . filter (\(p, _, _) -> mbFile == Just p) <$> getDiagnostics state
+    (ideOptions, join -> parsedModule, join -> env) <- runAction "CodeAction" state $
       (,,) <$> getIdeOptions
             <*> getParsedModule `traverse` mbFile
             <*> use GhcSession `traverse` mbFile
@@ -99,11 +100,11 @@ codeAction lsp state (TextDocumentIdentifier uri) _range CodeActionContext{_diag
     localExports <- readVar (exportsMap $ shakeExtras state)
     let exportsMap = localExports <> fromMaybe mempty pkgExports
     let dflags = hsc_dflags . hscEnv <$> env
-    pure $ Right
+    pure . Right $
         [ CACodeAction $ CodeAction title (Just CodeActionQuickFix) (Just $ List [x]) (Just edit) Nothing
-        | x <- xs, (title, tedit) <- suggestAction dflags exportsMap ideOptions ( join parsedModule ) text x
+        | x <- xs, (title, tedit) <- suggestAction dflags exportsMap ideOptions parsedModule text x
         , let edit = WorkspaceEdit (Just $ Map.singleton uri $ List tedit) Nothing
-        ]
+        ] <> removeAllRedundantImports parsedModule text allDiagnostics uri
 
 -- | Generate code lenses.
 codeLens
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -201,6 +202,21 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
...
+    r <- join $ map (suggestRemoveRedundantImport pm contents) diagnostics,
+    tes <- nub [ e | (_, edits) <- r, e <- edits],
+    _changes <- Just $ Map.singleton uri $ List tes,
+    _title <- "Remove all redundant imports",
+    _kind <- Just CodeActionQuickFix,
+    _diagnostics <- Nothing,
+    _documentChanges <- Nothing,
+    _edit <- Just WorkspaceEdit{..},
+    _command <- Nothing
+   =  [CACodeAction CodeAction {..}]
```
pepeiborra, line 217: Will the code action show when there are no redundant imports?

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -201,6 +202,21 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
         = [("Remove import", [TextEdit (extendToWholeLineIfPossible contents _range) ""])]
     | otherwise = []
 
+removeAllRedundantImports :: Maybe ParsedModule -> Maybe T.Text -> [Diagnostic] -> Uri -> [CAResult]
+removeAllRedundantImports m contents diagnostics uri
+  | Just pm <- m,
+    r <- join $ map (suggestRemoveRedundantImport pm contents) diagnostics,
+    tes <- nub [ e | (_, edits) <- r, e <- edits],
```
pepeiborra, line 209: let's avoid nub, it's quadratic and I want ghcide to stay speedy in a module with 100 redundant imports

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -201,6 +202,21 @@ suggestRemoveRedundantImport ParsedModule{pm_parsed_source = L _  HsModule{hsmod
         = [("Remove import", [TextEdit (extendToWholeLineIfPossible contents _range) ""])]
     | otherwise = []
 
+removeAllRedundantImports :: Maybe ParsedModule -> Maybe T.Text -> [Diagnostic] -> Uri -> [CAResult]
+removeAllRedundantImports m contents diagnostics uri
+  | Just pm <- m,
+    r <- join $ map (suggestRemoveRedundantImport pm contents) diagnostics,
```
pepeiborra, line 208: Is there any way to avoid calling `suggestRemoveRedundantImport` twice of every diagnostic?

pepeiborra (changes requested):

pepeiborra:
Does it work? Can you add a test please?

berberman:
Any instruction of tests? I'm not sure what to add in `removeImportTests`.

src/Development/IDE/Plugin/CodeAction.hs
berberman: It was a mistake, and I don't think `TextEdit` will duplicate, so just remove it.

src/Development/IDE/Plugin/CodeAction.hs
berberman: Added a `not null` guard.

pepeiborra:
All the tests live in test/exe/Main.hs
You can find the tests for the redundant imports code action there too

berberman:
```haskell
removeImportTests :: TestTree
removeImportTests = testGroup "remove import actions"
  [ testSession "redundant" $ do
      let contentA = T.unlines
            [ "module ModuleA where"
            ]
      _docA <- createDoc "ModuleA.hs" "haskell" contentA
      let contentB = T.unlines
            [ "{-# OPTIONS_GHC -Wunused-imports #-}"
            , "module ModuleB where"
            , "import ModuleA"
            , "stuffB :: Integer"
            , "stuffB = 123"
            ]
      docB <- createDoc "ModuleB.hs" "haskell" contentB
      _ <- waitForDiagnostics
      -- two code actions are expected
      --one for remove the single redundant import in range, the ohter for remove all
      [CACodeAction action@CodeAction { _title = actionTitle }] 
          <- getCodeActions docB (Range (Position 2 0) (Position 2 5))
      liftIO $ "Remove import" @=? actionTitle
      executeCodeAction action
      -- the import has been removed so far...
      contentAfterAction <- documentContents docB
      let expectedContentAfterAction = T.unlines
            [ "{-# OPTIONS_GHC -Wunused-imports #-}"
            , "module ModuleB where"
            , "stuffB :: Integer"
            , "stuffB = 123"
            ]
      liftIO $ expectedContentAfterAction @=? contentAfterAction
  , -- ......
```

`executeCodeAction` here apples the edit which removes the single redundant import to the document, so I have no where to place the test of remove all redundant imports. Can I extract `TextEdit` from `CodeAction` then use `applyEdit` to avoid modifying the document in session directly, or I have to create separate tests?

pepeiborra:
Please create a separate test

berberman:
Ok. For `removeImportTests`, we have to modify each pattern matching on `getCodeActions`, because there will be a new code action. I think `[CACodeAction action@CodeAction { _title = actionTitle }, _]` would be enough, since we know that "Remove all redundant imports" always appears later than "Remove import".

test/exe/Main.hs
```diff
@@ -929,6 +929,48 @@ removeImportTests = testGroup "remove import actions"
...
+            , "staffC = ()"
+            , "staffD = False"
+            , "data E = F"
+            ]
+      _docB <- createDoc "ModuleB.hs" "haskell" contentB
+      let contentC = T.unlines
+            [ "{-# OPTIONS_GHC -Wunused-imports #-}"
+            , "module ModuleC where"
+            , "import ModuleA (staffA, staffB, D(A, B))"
+            , "import ModuleB (staffC, staffD, E(F))"
```
pepeiborra, line 952: You might be able to simplify this test by using imports from `base` instead of defining auxiliary modules

pepeiborra (approved):
Looks good with some minor suggestions. Thank you for adding a test!

berberman:
Thanks for the review and suggestions! I enum more cases in the test.

pepeiborra:
Merged! Thanks again, this is an awesome contribution I've been wanting for a long time.

PR #871, opened 2020-10-15
Fix pretty printer for diagnostic ranges

hurryabit:
With the current implementation, VS Code will show "1:1" for the top
left corner, but the pretty printer renders this poisition to "1:0".
This is particularly interesting for people building command line tools
using `ghcide`, like the our DAML compiler at Digital Asset.
tools with command line drivers, like us at Digital Asset.

I would argue that VS Code has the ultimate authority on this since we
can't change what it displays without also moving the squiggly lines.

This PR fixes the discrepance by simply adding one to the column number
in the prtty printer, like we do for the line number.

### DIFF
--- a/src/Development/IDE/Types/Diagnostics.hs
+++ b/src/Development/IDE/Types/Diagnostics.hs
@@ -89,7 +89,7 @@ type FileDiagnostic = (NormalizedFilePath, ShowDiagnostic, Diagnostic)
 
 prettyRange :: Range -> Doc Terminal.AnsiStyle
 prettyRange Range{..} = f _start <> "-" <> f _end
-    where f Position{..} = pretty (_line+1) <> colon <> pretty _character
+    where f Position{..} = pretty (_line+1) <> colon <> pretty (_character+1)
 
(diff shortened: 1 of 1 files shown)


pepeiborra (approved):
Seems very reasonable, thanks

ndmitchell:
Is it worth including a test somewhere - these off-by-one things could certainly recur by accident quite easily.

hurryabit:
> Is it worth including a test somewhere - these off-by-one things could certainly recur by accident quite easily.

@ndmitchell Oh yes, they do. Do you remember our `damlc` integration tests? It got so bad, that I decided to add a test for the test utilities: https://github.com/digital-asset/daml/pull/7708.

I suppose this giant file is the right place for a test: https://github.com/haskell/ghcide/blob/master/test/exe/Main.hs ?

ndmitchell:
Yep, that giant file is the one. You've answered the question who tests the tests, but now who tests the tests that test the tests?

hurryabit:
@ndmitchell Here we go: https://github.com/haskell/ghcide/pull/878

PR #875, opened 2020-10-19
simplify things unnecessarily running in GhcM

wz1000:
Most of the times we don't need to mutate the `HscEnv`, so we don't need to store it in an `IORef`. The few times we do, it is better to be explicit, instead of risking implicitly throwing away changes by using `evalGhcEnv`.


pepeiborra:
I fully agree with the sentiment but just don't have the energy to review right now

pepeiborra (approved):

PR #877, opened 2020-10-19
Move HIE files stuff to a new hie-compat package

wz1000:
Last remaining hurdle for hiedb.

Once this is merged I will upload it on hackage, followed by the long awaited hackage release of hiedb.

I need volunteers for co-maintainers.


fendor:
@wz1000 I am a volunteer :)

wz1000:
ghc-lib job is very misleading :(

It says its for `ghc-lib-8.8`, which is true, but it is actually built using `ghc-8.6.5`

pepeiborra (approved):

PR #878, opened 2020-10-19
Add a test to check diagnostic ranges are printed 1-based

hurryabit:
Recently, we fixed a bug in `prettyRange` where lines where rendered
1-based but columns 0-based. Let's make sure we don't get into such
weird situations again by adding a test.

---
This change is [Reviewable](https://reviewable.io/reviews/haskell/ghcide/878)


jneira (approved):
Many thanks, adding new tests in a pr is a nice surprise

PR #879, opened 2020-10-20
Fix cabal check for hie-compat

wz1000:
Workaround for https://github.com/haskell/cabal/issues/4629 so we can upload to hackage


wz1000:
Package candidate here: https://hackage.haskell.org/package/hie-compat-0.1.0.0/candidate

pepeiborra (approved):

wz1000:
@pepeiborra shall I add you as a hackage maintainer?

PR #882, opened 2020-10-23
Don't need to invoke full typechecking logic for completions

wz1000:
Calling `tcRnImportDecls` is sufficient

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -2,6 +2,7 @@
 -- SPDX-License-Identifier: Apache-2.0
 
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE CPP #-}
 #include "ghc-api-version.h"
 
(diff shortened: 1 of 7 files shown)


pepeiborra (commented):
This is great, but it would be remiss of me not to ask for benchmarks. Can reuse the CI output if we just look at the allocations column

wz1000:
Doesn't seem to make much of a impact:

```
Cabal-3.0.0.0                upstream   completions after edit       True      100       24.113088235000003   0.0                  5.959781221999999     38.010221214            43.97857776           385393264      68306954608
Cabal-3.0.0.0                HEAD       completions after edit       True      100       24.239612734         0.0                  6.3938819959999975    37.65904611600002       44.061182254          331660184      68452162808
haskell-lsp-types-0.22.0.0   upstream   completions after edit       True      100       10.255219029000001   0.0                  26.246635755          40.81364477799999       67.074192193          255080480      71624692160
haskell-lsp-types-0.22.0.0   HEAD       completions after edit       True      100       10.176196605000001   0.0                  25.936275921999993    41.38951759199998       67.342077532          268654664      71544794864
```

pepeiborra (approved):
This cleans up the codebase, enhances the `ModSummaryWithoutTimestamps` rule result to include imports (which can be useful for other things, like the import lens HLS plugin) and doesn't impact performance negatively, so it's a win.

PR #886, opened 2020-10-27
Expose Development.IDE.Core.Preprocessor

tittoassini:
Preprocessor functionality is required by the Eval Plugin.

### DIFF
--- a/src/Development/IDE/GHC/Orphans.hs
+++ b/src/Development/IDE/GHC/Orphans.hs
@@ -15,7 +15,7 @@ import           Control.DeepSeq
 import           Data.Hashable
 import           Development.IDE.GHC.Compat
 import           Development.IDE.GHC.Util
-import           GHC
+import           GHC                        ()
 import           GhcPlugins
 import qualified StringBuffer               as SB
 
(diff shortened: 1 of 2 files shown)


src/Development/IDE/GHC/Orphans.hs
```diff
@@ -10,14 +10,14 @@
...
 
-import GHC
-import GhcPlugins
-import Development.IDE.GHC.Compat
-import qualified StringBuffer as SB
-import Control.DeepSeq
-import Data.Hashable
-import Development.IDE.GHC.Util
-import Bag
+import           Bag
```
ndmitchell, line 13: As a project we don't like this style - better not to indent things more than necessary.

ndmitchell (commented):
This seems a perfectly reasonable thing to expose - its pretty well designed as an API.

src/Development/IDE/GHC/Orphans.hs
tittoassini: I seem to have made a mistake, the first two commits have already been accepted as part of a previous PR.

The one I meant to submit is only the last one (exposing Development.IDE.Core.Preprocessor).

tittoassini: Sorry for the confusion, I have opened a different PR with just the required change.

PR #889, opened 2020-10-28
Use hole fit plugins to record holes instead of parsing error messages

wz1000:
Only works on 8.10+, so we would need to keep the parsing logic by CPPing the rest.

/cc @isovector I think this approach will provide a more robust way to get the local environment (`TcGblEnv`/`TcLclEnv`) at a hole, instead of going through `LocalBindings`. Thoughts? The tactic logic could work entirely inside `TcRn`.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -71,6 +71,7 @@ import           StringBuffer                   as SB
 import           TcRnMonad
 import           TcIface                        (typecheckIface)
 import           TidyPgm
+import Constraint
 
 import Control.Exception.Safe
 import Control.Monad.Extra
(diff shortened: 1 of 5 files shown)


wz1000:
@pepeiborra do you think it is a good idea to use this approach on 8.10 and keep the parsing logic for earlier ghcs? Or should we just stick with a single code path throughout?

Hopefully we can get rid of all the parsing stuff when errors as values lands.

pepeiborra:
Errors as values will take a while to arrive, so the plugin approach is good for 8.10, 9.0 and probably 9.2. By 9.4 we will be looking to migrate to errors as values, dropping support for 8.8 and hopefully avoiding having 3 code paths for dealing with hole fits. 

On the other hand, if we think that errors as values is going to land in 9.2, then I would say don't bother.

pepeiborra:

```
git fetch ghcide
```
Thanks!

PR #894, opened 2020-11-06
Expose Development.IDE.GHC.Orphans

jneira:
* To make available instances downstream
* Needed to use ghcide master in hls: https://github.com/haskell/haskell-language-server/pull/568


ndmitchell (approved):
Fine by me - I don't think we're too precious about what we expose.

jneira:
At the end it is not strictly needed for hls, but i think import specifically `Orphans` to get its instances makes sense.

PR #893, opened 2020-11-05
Disable CI benchmark suite

pepeiborra:
It appears that the benchmark suite gets stuck due to `stack exec` grabbing a file lock that it didn't use to. 

This PR is to debug the issues to try find a fix or workaround

```

ghcide> still blocking for directory lock on /home/vsts/work/1/s/.stack-work/dist/x86_64-linux/Cabal-2.4.0.1/build-lock; maybe another Stack process is running?
ghcide> still blocking for directory lock on /home/vsts/work/1/s/.stack-work/dist/x86_64-linux/Cabal-2.4.0.1/build-lock; maybe another Stack process is running?
ghcide> still blocking for directory lock on /home/vsts/work/1/s/.stack-work/dist/x86_64-linux/Cabal-2.4.0.1/build-lock; maybe another Stack process is running?
```

PR #895, opened 2020-11-07
GitHub actions

pepeiborra:
Adds a Github action for benchmarking, using Cabal instead of stack (since the stack implementation gets stuck, see #893)

Builds on top of #893


bubba (approved):
I can't see any run for this PR but the workflow file makes sense to me

ndmitchell (approved):

pepeiborra:
Runs: https://github.com/pepeiborra/ghcide/actions

jneira (approved):
LGTM
It would be great to run it in windows but it is matter of a another pr 😄

pepeiborra:
Benchmarks do not build in Windows because of the Cairo requirement

PR #897, opened 2020-11-07
Add code actions for disabling a warning in the current file

### LINKED ISSUE #571: Suggest adding -Wno-orphans when orphans are encountered
It's one of the warning I disable most frequently (in application code only of course!).

I wonder if it's even worth not suggesting this in library code, or if the inconsistency would just be confusing.

Just an idea, I'm sure it's not to everyone's taste!

georgefst:
This is a little ugly - we need some way to keep track of the [WarningFlag](https://hackage.haskell.org/package/ghc-8.10.1/docs/DynFlags.html#t:WarningFlag), so we store it as a string in the (previously unused by Ghcide) `_code` field of [Diagnostic](https://hackage.haskell.org/package/haskell-lsp-types-0.23.0.0/docs/Language-Haskell-LSP-Types.html#t:Diagnostic). I'm open to a better way.

Closes haskell/haskell-language-server#571.

### DIFF
--- a/src/Development/IDE/GHC/Warnings.hs
+++ b/src/Development/IDE/GHC/Warnings.hs
@@ -12,6 +12,7 @@ import qualified           Data.Text as T
 import           Development.IDE.Types.Diagnostics
 import           Development.IDE.GHC.Error
 import           Language.Haskell.LSP.Types (NumberOrString (StringValue))
+import Data.Char
 
 
 -- | Take a GHC monadic action (e.g. @typecheckModule pm@ for some
@@ -39,6 +40,7 @@ attachReason wr d = d{_code = StringValue <$> showReason wr}
   where
     showReason = \case
         NoReason -> Nothing
-        Reason flag -> Just $ showT flag
-        ErrReason flag -> showT <$> flag
-    showT = T.pack . show
+        Reason flag -> showFlag flag
+        ErrReason flag -> showFlag =<< flag
+    showFlag = fmap (("-W" <>) . camelToHyphenCase) . T.stripPrefix "Opt_Warn" . T.pack . show
+    camelToHyphenCase = T.dropWhile (== '-') . T.concatMap (\c -> if isUpper c then "-" <> T.singleton (toLower c) else T.singleton c)
--- a/src/Development/IDE/GHC/Warnings.hs
+++ b/src/Development/IDE/GHC/Warnings.hs
@@ -3,6 +3,7 @@
 
 module Development.IDE.GHC.Warnings(withWarnings) where
 
+import Data.List
 import ErrUtils
 import GhcPlugins as GHC hiding (Var, (<>))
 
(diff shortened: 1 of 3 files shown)


src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -181,6 +182,17 @@ suggestAction dflags packageExports ideOptions parsedModule text diag = concat
     ] ++
     suggestFillHole diag                   -- Lowest priority
 
+suggestDisableWarning :: Diagnostic -> [(T.Text, [TextEdit])]
+suggestDisableWarning Diagnostic{..}
+    | Just (StringValue (showFlag -> Just w)) <- _code =
+        pure
+            ( "Disable \"" <> w <> "\" warnings"
```
georgefst, line 189: Exact wording of message is bikeshed-able

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -181,6 +182,17 @@ suggestAction dflags packageExports ideOptions parsedModule text diag = concat
...
+            , [TextEdit (Range (Position 0 0) (Position 0 0)) $ "{-# OPTIONS_GHC -Wno-" <> w <> " #-}\n"]
```
georgefst, line 190: Alternatively, we could try to insert beneath any existing pragmas. If e.g. we have an `{-# OPTIONS_GHC -Wall #-}` in the file, then inserting this pragma at the top may have no effect.

pepeiborra: Does the order of ghc options matter? I'm not quite sure
Either way, I would prefer if this pragma was inserted beneath the existing pragmas, specially because it's quite common to have a copyright notice *above* the pragmas

pepeiborra (commented):
A test would be great

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -181,6 +182,17 @@ suggestAction dflags packageExports ideOptions parsedModule text diag = concat
...
+            )
+    | otherwise = []
+  where
+    showFlag = fmap camelToHyphenCase . T.stripPrefix "Opt_Warn"
+    camelToHyphenCase = T.dropWhile (== '-') . T.concatMap (\c -> if isUpper c then "-" <> T.singleton (toLower c) else T.singleton c)
```
georgefst, line 195: We need to go from e.g. "Opt_WarnUnusedImports" to "unused-imports". I'd expect this to be in the GHC API somewhere, but I can't find anything.

src/Development/IDE/Plugin/CodeAction.hs
georgefst: Yeah, I've just checked. With:
```hs
{-# OPTIONS_GHC -Wno-unused-imports #-}
{-# OPTIONS_GHC -Wall #-}
```

the first pragma has no effect.

src/Development/IDE/GHC/Warnings.hs
```diff
@@ -27,8 +28,17 @@ withWarnings diagSource action = do
...
       newAction dynFlags wr _ loc style msg = do
-        let wr_d = fmap (wr,) $ diagFromErrMsg diagSource dynFlags $ mkWarnMsg dynFlags loc (queryQual style) msg
+        let wr_d = map ((wr,) . third3 (attachReason wr)) $ diagFromErrMsg diagSource dynFlags $ mkWarnMsg dynFlags loc (queryQual style) msg
         modifyVar_ warnings $ return . (wr_d:)
   res <- action $ \x -> x{ms_hspp_opts = (ms_hspp_opts x){log_action = newAction}}
   warns <- readVar warnings
   return (reverse $ concat warns, res)
+
+attachReason :: WarnReason -> Diagnostic -> Diagnostic
+attachReason wr d = d{_code = StringValue <$> showReason wr}
```
wz1000, line 38: This seems pretty weird, especially since the specification says this field may be displayed to the user. Perhaps the reason can be recorded in the ide state, and the code action can be generated based off of that?

georgefst: > This seems pretty weird, especially since the specification says this field may be displayed to the user.

Hmm, yeah, that is unfortunate. I was originally intending to use the more human-readable form here, analogously to how GHC shows e.g. `-Wunused-imports` in the CLI output. But this way it was easier to be sure that a particular diagnostic comes from a GHC warning (using `T.stripPrefix "Opt_Warn"`).

> Perhaps the reason can be recorded in the ide state, and the code action can be generated based off of that?

That sounds promising. There's a `diagnostics` field in `ShakeExtras` but the `DiagnosticStore` type doesn't seem to give us anywhere to store extra information, so I guess it would need replacing - I haven't yet looked at how much of a knock-on effect that might have.

georgefst: I'm struggling to see a nice way to put this information in `ShakeExtras`.

Besides, based on other LSP servers, I think this actually is more or less the intended use case for the `code` field. I'm thinking we put the full flag there e.g. `-Wunused-imports`, and trigger the action for diagnostics whose code begins with "-W" Still a little bit ugly, but the very nature of LSP means we're bound to be resigned to the occasional but of string-ly typed code.

The way VScode at least shows this is actually quite nice: `typecheck(-Wunused-imports)`. Regardless of the original purpose of this PR, it's nice to get this extra information about where a warning is coming from. In fact, it's the one thing in GHC's output that we're currently missing.

src/Development/IDE/Plugin/CodeAction.hs
georgefst: Is there a clean way to find the pragma locations? AFAICT, they don't appear in `ParsedModule`.

Best thing I can think of otherwise is to get the location of the module declaration (falling back to first import, then first declaration, if there isn't one), and insert immediately before that. Or even backtrack from there by looking at the raw text for lines beginning with `{-#`...

georgefst: Presumably we want pretty much the same logic for adding language pragmas? Currently those also go right to the top of the file. So perhaps this belongs in a follow-up PR.

src/Development/IDE/GHC/Warnings.hs
georgefst: > I'm thinking we put the full flag there e.g. -Wunused-imports

This is what c2cd237 does.

src/Development/IDE/Plugin/CodeAction.hs
pepeiborra: From the parsed module you can get the line for the `module Foo where...` declaration, and just put the new pragmas above it.

pepeiborra (changes requested):
This looks good, but let's fix the location of the inserted pragmas to make sure the new flags are not overridden by later pragmas

src/Development/IDE/Plugin/CodeAction.hs
pepeiborra: There is `fmap flagName . findSpecOf`.

georgefst: Where does `findSpecOf` come from? I can't find that anywhere.

pepeiborra (approved):

src/Development/IDE/Plugin/CodeAction.hs
pepeiborra: Sorry, I meant `flagSpecOf`: https://hackage.haskell.org/package/ghc-8.10.1/docs/src/DynFlags.html#flagSpecOf

But it turns out it's not imported :(

georgefst: You mean not exported? Perhaps we should ask GHC to change that.

Turns out this logic is only valid for 68 out of 82 warnings...

pepeiborra:
What's left to merge this?

georgefst:
Assuming @wz1000 is now happy with the `_code` field being used, after my comments, then the remaining issue is what to do about `showFlag`. I think what I'll have to do is just spell out all the cases (effectively duplicating `flagSpecOf`). Or spell out the 14 broken cases, then fall back to the current implementation - this is similar to what we do for passing flags to Ormolu/Fourmolu in HLS, where `-XCPP` needs to be special cases, but it's a little brittle.

Meanwhile, shall I open a PR on GHC's GitLab about exporting `flagSpecOf`? I don't know enough about that part of the code to know whether it would be sensible.

ndmitchell:
> Meanwhile, shall I open a PR on GHC's GitLab about exporting flagSpecOf? I don't know enough about that part of the code to know whether it would be sensible.

Yes. As a general approximation, if its useful, it should be exported from GHC. GHC isn't particularly careful about presenting a nice carefully crafted API. We've asked for things exported in the past - if it will be exported, adding it to the Compat layer for now and then introducing an ifdef to make it available later is a good way to go.

georgefst:
Ok, thanks @ndmitchell, I'll do that.

Just to note, I'm *very* busy at the moment - in the middle of changing jobs and moving between cities. I'll probably get to this next weekend.

src/Development/IDE/GHC/Warnings.hs
alanz: The `code` field is intended to refer to the "handbook code" for a diagnostic, as provided by some compilers.  Version 3.16 of the spec has a `data` field for this, prompted from https://github.com/microsoft/language-server-protocol/issues/887.  Perhaps use that instead.

georgefst: I think seeing as the flag is the closest thing to a "handbook code" GHC is likely to have any time soon, this is still the right place to put it. Especially given that the result is visually appealing in VScode (haven't tried other editors, but I'd hope they show something relatively similar).

Obviously I'll change this if you or any of the other core maintainers feel strongly.

src/Development/IDE/Plugin/CodeAction.hs
georgefst: Closed by 71da233

georgefst:
Done.

georgefst:
Ok, now this really is finished. Thanks @mpickering for pointing out `wWarningFlags`.

georgefst:
What 5a2ffa4 does is essentially inline the definition of `flagSpecOf`.

It is a bit inefficient as it effectively uses a list as a map. But we don't have `Ord` for `WarningFlag`. `IntMap` could be an option...

Then again, it's a list with bounded length (number of warnings - somewhere under 100), so it's not disastrous.

src/Development/IDE/Plugin/CodeAction.hs
georgefst: > Closed by 71da233

Actually resolved in a more maintainable way by 5a2ffa4.

pepeiborra:
Thank you @georgefst ! Please take a look at the merge conflicts so that we can merge this ASAP

georgefst:
@pepeiborra Rebased and resolved.

pepeiborra:
Looks like there are still some test failures

georgefst:
> Looks like there are still some test failures

Ah, whoops. I'd only been running the ones I added. Turns out three of the old ones are broken by the presence of an additional code action.

64a0aa5 does the bare minimum to get the tests fixed but in some cases they should perhaps actually be made more robust instead...

test/exe/Main.hs
```diff
@@ -1042,7 +1044,7 @@ removeImportTests = testGroup "remove import actions"
             ]
       doc <- createDoc "ModuleC.hs" "haskell" content
       _ <- waitForDiagnostics
-      [_, _, _, _, CACodeAction action@CodeAction { _title = actionTitle }]
+      [_, _, _, _, _, _, _, _, CACodeAction action@CodeAction { _title = actionTitle }]
```
georgefst, line 1047: This is particularly nasty.

pepeiborra:

```
git fetch ghcide
```
Thanks!

PR #898, opened 2020-11-08
Integrate hiedb

wz1000:
Includes support for references, workspace symbols, faster loading by loading (stale) .hie files from disk ~~and jump to definition for arbitrary dependencies.~~

As described [here](https://www.tweag.io/blog/2020-10-07-ghcide-fellowship-summary/) and [here](https://mpickering.github.io/ide/posts/2020-05-29-hiedb.html)

TODO:

- [ ] Write documentation
- [ ] Tests
- [x] Release `hiedb` on hackage

I'll be pressed for time in the near future, so would appreciate any help to get this over the finish line, especially with the first two points above.

### DIFF
--- a/exe/Arguments.hs
+++ b/exe/Arguments.hs
@@ -1,15 +1,18 @@
 -- Copyright (c) 2019 The DAML Authors. All rights reserved.
 -- SPDX-License-Identifier: Apache-2.0
 
-module Arguments(Arguments(..), getArguments) where
+module Arguments(Arguments, Arguments'(..), getArguments, IdeCmd(..)) where
 
 import Options.Applicative
+import HieDb.Run
 
+type Arguments = Arguments' IdeCmd
 
-data Arguments = Arguments
+data IdeCmd = Typecheck [FilePath] | DbCmd Command | LSP
+
+data Arguments' a = Arguments
     {argLSP :: Bool
     ,argsCwd :: Maybe FilePath
-    ,argFiles :: [FilePath]
     ,argsVersion :: Bool
     ,argsShakeProfiling :: Maybe FilePath
     ,argsOTMemoryProfiling :: Bool
@@ -17,6 +20,7 @@ data Arguments = Arguments
     ,argsDisableKick :: Bool
     ,argsThreads :: Int
     ,argsVerbose :: Bool
+    ,argFilesOrCmd :: a
     }
 
 getArguments :: IO Arguments
@@ -24,14 +28,12 @@ getArguments = execParser opts
   where
     opts = info (arguments <**> helper)
       ( fullDesc
-     <> progDesc "Used as a test bed to check your IDE will work"
      <> header "ghcide - the core of a Haskell IDE")
 
 arguments :: Parser Arguments
 arguments = Arguments
       <$> switch (long "lsp" <> help "Start talking to an LSP server")
       <*> optional (strOption $ long "cwd" <> metavar "DIR" <> help "Change to this directory")
-      <*> many (argument str (metavar "FILES/DIRS..."))
       <*> switch (long "version" <> help "Show ghcide and GHC versions")
       <*> optional (strOption $ long "shake-profiling" <> metavar "DIR" <> help "Dump profiling reports to this directory")
       <*> switch (long "ot-memory-profiling" <> help "Record OpenTelemetry info to the eventlog. Needs the -l RTS flag to have an effect")
@@ -39,3 +41,12 @@ arguments = Arguments
       <*> switch (long "test-no-kick" <> help "Disable kick. Useful for testing cancellation")
       <*> option auto (short 'j' <> help "Number of threads (0: automatic)" <> metavar "NUM" <> value 0 <> showDefault)
       <*> switch (long "verbose" <> help "Include internal events in logging output")
+      <*> ( hsubparser (command "typecheck" (info (Typecheck <$> fileCmd) fileInfo)
+                   <> command "hiedb" (info (DbCmd <$> cmdParser <**> helper) hieInfo)
+                   <> command "lsp" (info (pure LSP <**> helper) lspInfo)  )
+         <|> Typecheck <$> fileCmd )
+  where
+    fileCmd = many (argument str (metavar "FILES/DIRS..."))
+    lspInfo = fullDesc <> progDesc "Start talking to an LSP server"
+    fileInfo = fullDesc <> progDesc "Used as a test bed to check your IDE will work"
+    hieInfo = fullDesc <> progDesc "Query .hie files"
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -2,6 +2,8 @@
 -- SPDX-License-Identifier: Apache-2.0
 {-# OPTIONS_GHC -Wno-dodgy-imports #-} -- GHC no longer exports def in GHC 8.6 and above
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE CPP #-}
+#include "ghc-api-version.h"
 
 module Main(main) where
 
@@ -31,7 +33,7 @@ import Development.IDE.Plugin
 import Development.IDE.Plugin.Completions as Completions
 import Development.IDE.Plugin.CodeAction as CodeAction
 import Development.IDE.Plugin.Test as Test
-import Development.IDE.Session (loadSession)
+import Development.IDE.Session (loadSession, cacheDir)
 import qualified Language.Haskell.LSP.Core as LSP
 import Language.Haskell.LSP.Messages
 import Language.Haskell.LSP.Types
@@ -55,6 +57,23 @@ import Text.Printf
 import Development.IDE.Core.Tracing
 import Development.IDE.Types.Shake (Key(Key))
 
+import HieDb.Create
+import HieDb.Types
+import HieDb.Utils
+import Database.SQLite.Simple
+import qualified Data.ByteString.Char8 as B
+import qualified Crypto.Hash.SHA1 as H
+import Control.Concurrent.Async
+import Control.Exception
+import System.Directory
+import Data.ByteString.Base16
+import HieDb.Run (Options(..), runCommand)
+import Maybes (MaybeT(runMaybeT))
+import HIE.Bios.Types (CradleLoadResult(..))
+import HIE.Bios.Environment (getRuntimeGhcLibDir)
+import DynFlags
+
+
 ghcideVersion :: IO String
 ghcideVersion = do
   path <- getExecutablePath
@@ -66,6 +85,31 @@ ghcideVersion = do
              <> ") (PATH: " <> path <> ")"
              <> gitHashSection
 
+-- | Wraps `withHieDb` to provide a database connection for reading, and a `HieWriterChan` for
+-- writing. Actions are picked off one by one from the `HieWriterChan` and executed in serial
+-- by a worker thread using a dedicated database connection.
+-- This is done in order to serialize writes to the database, or else SQLite becomes unhappy
+runWithDb :: FilePath -> (HieDb -> HieWriterChan -> IO ()) -> IO ()
+runWithDb fp k =
+  withHieDb fp $ \writedb -> do
+    execute_ (getConn writedb) "PRAGMA journal_mode=WAL;"
+    initConn writedb
+    chan <- newChan
+    race_ (writerThread writedb chan) (withHieDb fp (flip k chan))
+  where
+    writerThread db chan = forever $ do
+      k <- readChan chan
+      k db `catch` \e@SQLError{} -> do
+        hPutStrLn stderr $ "Error in worker, ignoring: " ++ show e
+
+getHieDbLoc :: FilePath -> IO FilePath
+getHieDbLoc dir = do
+  let db = dirHash++"-"++takeBaseName dir++"-"++VERSION_ghc <.> "hiedb"
+      dirHash = B.unpack $ encode $ H.hash $ B.pack dir
+  cDir <- IO.getXdgDirectory IO.XdgCache cacheDir
+  createDirectoryIfMissing True cDir
+  pure (cDir </> db)
+
 main :: IO ()
 main = do
     -- WARNING: If you write to stdout before runLanguageServer
@@ -75,15 +119,47 @@ main = do
     if argsVersion then ghcideVersion >>= putStrLn >> exitSuccess
     else hPutStrLn stderr {- see WARNING above -} =<< ghcideVersion
 
+    whenJust argsCwd IO.setCurrentDirectory
+
+    -- We want to set the global DynFlags right now, so that we can use
+    -- `unsafeGlobalDynFlags` even before the project is configured
+    dir <- IO.getCurrentDirectory
+    dbLoc <- getHieDbLoc dir
+    hieYaml <- runMaybeT $ yamlConfig dir
+    cradle <- maybe (loadImplicitCradle $ addTrailingPathSeparator dir) loadCradle hieYaml
+    libDirRes <- getRuntimeGhcLibDir cradle
+    libdir <- case libDirRes of
+        CradleSuccess libdir -> pure $ Just libdir
+        CradleFail err -> do
+          hPutStrLn stderr $ "Couldn't load cradle for libdir: " ++ show err
+          return Nothing
+        CradleNone -> return Nothing
+    dynFlags <- mapM (dynFlagsForPrinting . LibDir) libdir
+    mapM_ setUnsafeGlobalDynFlags dynFlags
+
+    case argFilesOrCmd of
+      DbCmd cmd -> do
+        let opts :: Options
+            opts = Options
+              { database = dbLoc
+              , trace = False
+              , quiet = False
+              , virtualFile = False
+              }
+        runCommand (LibDir $ fromJust libdir) opts cmd
+      Typecheck (Just -> argFilesOrCmd) | not argLSP -> runWithDb dbLoc $ runIde dir Arguments{..}
+      _ -> let argFilesOrCmd = Nothing in runWithDb dbLoc $ runIde dir Arguments{..}
+
+
+runIde :: FilePath -> Arguments' (Maybe [FilePath]) -> HieDb -> HieWriterChan -> IO ()
+runIde dir Arguments{..} hiedb hiechan = do
+    command <- makeLspCommandId "typesignature.add"
+
     -- lock to avoid overlapping output on stdout
     lock <- newLock
     let logger p = Logger $ \pri msg -> when (pri >= p) $ withLock lock $
             T.putStrLn $ T.pack ("[" ++ upper (show pri) ++ "] ") <> msg
 
-    whenJust argsCwd IO.setCurrentDirectory
-
-    dir <- IO.getCurrentDirectory
-    command <- makeLspCommandId "typesignature.add"
 
     let plugins = Completions.plugin <> CodeAction.plugin
             <> if argsTesting then Test.plugin else mempty
@@ -97,8 +173,8 @@ main = do
         options = def { LSP.executeCommandCommands = Just [command]
                       , LSP.completionTriggerCharacters = Just "."
                       }
-
-    if argLSP then do
+    case argFilesOrCmd of
+      Nothing -> do
         t <- offsetTime
         hPutStrLn stderr "Starting LSP server..."
         hPutStrLn stderr "If you are seeing this in a terminal, you probably should have run ghcide WITHOUT the --lsp option!"
@@ -127,8 +203,8 @@ main = do
                   unless argsDisableKick $
                     action kick
             initialise caps rules
-                getLspId event wProg wIndefProg (logger logLevel) debouncer options vfs
-    else do
+                getLspId event wProg wIndefProg (logger logLevel) debouncer options vfs hiedb hiechan
+      Just argFiles -> do
         -- GHC produces messages with UTF8 in them, so make sure the terminal doesn't error
         hSetEncoding stdout utf8
         hSetEncoding stderr utf8
@@ -162,7 +238,7 @@ main = do
                     , optCheckProject      = CheckProject False
                     }
             logLevel = if argsVerbose then minBound else Info
-        ide <- initialise def mainRule (pure $ IdInt 0) (showEvent lock) dummyWithProg (const (const id)) (logger logLevel) debouncer options vfs
+        ide <- initialise def mainRule (pure $ IdInt 0) (showEvent lock) dummyWithProg (const (const id)) (logger logLevel) debouncer options vfs hiedb hiechan
 
         putStrLn "\nStep 4/4: Type checking the files"
         setFilesOfInterest ide $ HashMap.fromList $ map ((, OnDisk) . toNormalizedFilePath') files
--- a/src/Development/IDE/Core/RuleTypes.hs
+++ b/src/Development/IDE/Core/RuleTypes.hs
@@ -5,6 +5,7 @@
 {-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE GADTs #-}
 
 -- | A Shake implementation of the compiler service, built
 --   using the "Shaker" abstraction layer for in-memory use.
@@ -134,18 +135,28 @@ instance Show HiFileResult where
 
 -- | Save the uncompressed AST here, we compress it just before writing to disk
 data HieAstResult
-  = HAR
+  = forall a. HAR
   { hieModule :: Module
-  , hieAst :: !(HieASTs Type)
-  , refMap :: RefMap
+  , hieAst :: !(HieASTs a)
+  , refMap :: RefMap a
   -- ^ Lazy because its value only depends on the hieAst, which is bundled in this type
   -- Lazyness can't cause leaks here because the lifetime of `refMap` will be the same
   -- as that of `hieAst`
+  , hieKind :: !(HieKind a)
+  -- ^ Is this hie file loaded from the disk, or freshly computed?
   }
 
-instance NFData HieAstResult where
-    rnf (HAR m hf _rm) = rnf m `seq` rwhnf hf
+data HieKind a where
+  HieFromDisk :: !HieFile -> HieKind TypeIndex
+  HieFresh :: HieKind Type
 
+instance NFData (HieKind a) where
+    rnf (HieFromDisk hf) = rnf hf
+    rnf HieFresh = ()
+ 
+instance NFData HieAstResult where
+    rnf (HAR m hf _rm kind) = rnf m `seq` rwhnf hf `seq` rnf kind
+ 
 instance Show HieAstResult where
     show = show . hieModule
 
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -79,7 +79,6 @@ import Development.IDE.Core.Shake
 import Development.Shake.Classes hiding (get, put)
 import Control.Monad.Trans.Except (runExceptT)
 import Control.Concurrent.Async (concurrently)
-import Control.Concurrent.Chan
 import Control.Monad.Reader
 import Control.Exception.Safe
 
@@ -96,8 +95,6 @@ import Control.Concurrent.Extra
 import Module
 
 import qualified HieDb
-import System.Directory
-import System.FilePath
 
 -- | This is useful for rules to convert rules that can only produce errors or
 -- a result into the more general IdeResult type that supports producing
@@ -130,32 +127,16 @@ defineEarlyCutOffNoFile f = defineEarlyCutoff $ \k file -> do
 ------------------------------------------------------------
 -- Exposed API
 
-lookupMod :: HieWriterChan -> FilePath -> ModuleName -> UnitId -> Bool -> MaybeT IdeAction Uri
-lookupMod dbchan hie_f mod uid boot
-  | stringToUnitId "fake_uid" ==  uid = MaybeT $ pure Nothing
-  | otherwise = do
-      ide <- ask
-      liftIO $ L.logInfo (logger ide) $ T.pack $ "LookupMod " ++ show hie_f
-      cachedir <- liftIO getCurrentDirectory
-      let dir = cachedir </> "cached-deps" </> unitIdString uid
-          src = dir </> map (\x -> if x == '.' then '-' else x) (moduleNameString mod) <.> ext
-          ext = if boot then "hs-boot" else "hs"
-      hieTimestamp <- MaybeT $ either (\(_ :: IOException) -> Nothing) Just <$> (liftIO $ try $ getModificationTime hie_f)
-      srcTimestamp   <- either (\(_ :: IOException) -> Nothing) Just <$> (liftIO $ try $ getModificationTime src)
-      case srcTimestamp of
-        Just t | t >= hieTimestamp -> do
-          liftIO $ L.logInfo (logger ide) $ T.pack $ "Returning file " ++ show src
-          pure $ fromNormalizedUri $ filePathToUri' $ toNormalizedFilePath' src
-        _ -> do
-          hf <- readHieFileFromDisk hie_f
-          liftIO $ do
-            createDirectoryIfMissing True dir
-            removePathForcibly src
-            BS.writeFile src $ hie_hs_src hf
-            p <- getPermissions src
-            setPermissions src (p {writable = False})
-            writeChan dbchan $ \db -> HieDb.addSrcFile db hie_f src False
-            pure $ fromNormalizedUri $ filePathToUri' $ toNormalizedFilePath' src
+-- | Eventually this will lookup/generate URIs for files in dependencies, but not in the
+-- project. Right now, this is just a stub.
+lookupMod
+  :: HieWriterChan -- ^ access the database
+  -> FilePath -- ^ The `.hie` file we got from the database
+  -> ModuleName
+  -> UnitId
+  -> Bool -- ^ Is this file a boot file?
+  -> MaybeT IdeAction Uri
+lookupMod _dbchan _hie_f _mod _uid _boot = MaybeT $ pure Nothing
 
 -- | Get all transitive file dependencies of a given module.
 -- Does not include the file itself.
@@ -189,10 +170,7 @@ getDefinition file pos = runMaybeT $ do
     !pos' <- MaybeT (return $ fromCurrentPosition mapping pos)
     hiedb <- lift $ asks hiedb
     hiedbChan <- lift $ asks hiedbWriter
-    let origFile
-         | "cached-deps" `isInfixOf` fromNormalizedFilePath file = Just file
-         | otherwise = Nothing
-    toCurrentLocations mapping <$> AtPoint.gotoDefinition hiedb (lookupMod $ channel hiedbChan) opts imports hf origFile pos'
+    toCurrentLocations mapping <$> AtPoint.gotoDefinition hiedb (lookupMod $ channel hiedbChan) opts imports hf pos'
 
 getTypeDefinition :: NormalizedFilePath -> Position -> IdeAction (Maybe [Location])
 getTypeDefinition file pos = runMaybeT $ do
@@ -202,10 +180,7 @@ getTypeDefinition file pos = runMaybeT $ do
     !pos' <- MaybeT (return $ fromCurrentPosition mapping pos)
     hiedb <- lift $ asks hiedb
     hiedbChan <- lift $ asks hiedbWriter
-    let origFile
-         | "cached-deps" `isInfixOf` fromNormalizedFilePath file = Just file
-         | otherwise = Nothing
-    toCurrentLocations mapping <$> AtPoint.gotoTypeDefinition hiedb (lookupMod $ channel hiedbChan) opts hf origFile pos'
+    toCurrentLocations mapping <$> AtPoint.gotoTypeDefinition hiedb (lookupMod $ channel hiedbChan) opts hf pos'
 
 highlightAtPoint :: NormalizedFilePath -> Position -> IdeAction (Maybe [DocumentHighlight])
 highlightAtPoint file pos = runMaybeT $ do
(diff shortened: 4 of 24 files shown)


wz1000:
Go to definition for dependencies works in the following way:

First, the user must index the `.hie` files for their dependencies by running the command `ghcide hiedb index $DIR` in the root directory of their project.

Then, whenever we report a definition in one of the dependencies(say `Control.Lens`), a read only file `Control.Lens.hs` is written out on demand in `root/.cached-deps/lens-XXX/Control.Lens.hs`. The contents of the file are obtained from the `.hie` file that was indexed in the previous step. The logic responsible for this is in `lookupMod` in `Rules.hs`.

The `.cached-deps` directory has to be inside the project directory, as otherwise `vscode` and other editors will try to spin up another instance of ghcide to

exe/Arguments.hs
```diff
@@ -1,37 +1,48 @@
...
 import Options.Applicative
+import HieDb.Run
 
+type Arguments = Arguments' IdeCmd
 
-data Arguments = Arguments
+data IdeCmd = Typecheck [FilePath] | DbCmd Command | LSP
+
+data Arguments' a = Arguments
     {argLSP :: Bool
```
pepeiborra, line 14: why is this field still needed?

exe/Arguments.hs
```diff
@@ -1,37 +1,48 @@
...
     ,argsCwd :: Maybe FilePath
-    ,argFiles :: [FilePath]
     ,argsVersion :: Bool
     ,argsShakeProfiling :: Maybe FilePath
     ,argsTesting :: Bool
     ,argsThreads :: Int
     ,argsVerbose :: Bool
+    ,argFilesOrCmd :: a
```
pepeiborra, line 21: Can we call this simply `argCmd` ? Or even take it out of this record, and change the type of `getArguments` to `(Arguments, IdeCmd)`.

src/Development/IDE/Core/RuleTypes.hs
```diff
@@ -23,20 +24,39 @@ import           Data.Hashable
...
+  deriving (Show, Generic, Eq, Ord)
+instance Hashable GetFOIReferences
+instance NFData   GetFOIReferences
+instance Binary   GetFOIReferences
+type instance RuleResult GetFOIReferences = FOIReferences
+
+newtype FOIReferences
+  = FOIReferences
+  { foiAsts :: HM.HashMap NormalizedFilePath (HieAstResult, PositionMapping)
+  }
```
pepeiborra, line 53: Doc comments

src/Development/IDE/Core/Shake.hs
```diff
@@ -164,6 +177,9 @@ data ShakeExtras = ShakeExtras
     ,exportsMap :: Var ExportsMap
     -- | A work queue for actions added via 'runInShakeSession'
     ,actionQueue :: ActionQueue
+    , hiedb :: HieDb -- ^ Use only to read
+    , hiedbWriter :: HieDbWriter -- ^ use to write
+    , persistentKeys :: Var (HMap.HashMap Key GetStalePersistent)
```
pepeiborra, line 182: doc comment.

Would it make sense to group these 3 new fields into a `HieDbState` record?

src/Development/IDE/Core/Shake.hs
```diff
@@ -195,6 +213,13 @@ getShakeExtrasRules = do
     Just x <- getShakeExtraRules @ShakeExtras
     return x
 
+addPersistentRule :: IdeRule k v => k -> (NormalizedFilePath -> IdeAction (Maybe v)) -> Rules ()
```
pepeiborra, line 216: doc comment

src/Development/IDE/Core/Shake.hs
```diff
@@ -268,30 +293,44 @@ instance NFData v => NFData (Value v)
...
-        Failed -> Nothing
+lastValueIO :: IdeRule k v => ShakeExtras -> k -> NormalizedFilePath -> IO (Maybe (v, PositionMapping))
+lastValueIO s@ShakeExtras{positionMapping,persistentKeys,state} k file = do
+  modifyVar state $ \hm -> do
+    let readPersistent = do
+          pmap <- readVar persistentKeys
+          mv <- runMaybeT $ do
+            liftIO $ Logger.logInfo (logger s) $ T.pack $ "LOOKUP UP PERSISTENT FOR" ++ show k
+            f <- MaybeT $ pure $ HMap.lookup (Key k) pmap
+            dv <- MaybeT $ runIdeAction "lastValueIO" s $ f file
```
pepeiborra, line 308: Isn't this a potential deadlock if `f file` tries to access the state?

src/Development/IDE/Core/Shake.hs
```diff
@@ -778,16 +819,16 @@ useWithStaleFast' key file = do
   liftIO $ case r of
```
pepeiborra, line 819: Isn't the call to `getValues` made redundant by `lastValueIO`?

src/Development/IDE/Core/Shake.hs
```diff
@@ -854,8 +895,8 @@ uses key files = map (\(A value) -> currentValue value) <$> apply (map (Q . (key
 usesWithStale :: IdeRule k v
     => k -> [NormalizedFilePath] -> Action [Maybe (v, PositionMapping)]
 usesWithStale key files = do
-    values <- map (\(A value) -> value) <$> apply (map (Q . (key,)) files)
-    zipWithM lastValue files values
+    _ <- apply (map (Q . (key,)) files)
+    mapM (lastValue key) files
```
pepeiborra, line 899: does this change the semantics of `usesWithStale`?

src/Development/IDE/Spans/Common.hs
```diff
@@ -30,23 +29,22 @@ import ConLike
...
-showName :: Outputable a => a -> T.Text
-showName = T.pack . prettyprint
-  where
-    prettyprint x = renderWithStyle unsafeGlobalDynFlags (ppr x) style
-    style = mkUserStyle unsafeGlobalDynFlags neverQualify AllTheWay
+showGhc :: Outputable a => a -> T.Text
+showGhc = showSD . ppr
+
+showSD :: SDoc -> T.Text
+showSD = T.pack . unsafePrintSDoc
```
pepeiborra, line 47: This used to render with `neverQualify AllTheWay`. Why remove those?

exe/Main.hs
```diff
@@ -62,6 +84,33 @@ ghcideVersion = do
              <> ") (PATH: " <> path <> ")"
              <> gitHashSection
 
+runWithDb :: Maybe DynFlags -> FilePath -> (HieDb -> HieWriterChan -> IO ()) -> IO ()
```
pepeiborra, line 87: Some comments here would be helpful

pepeiborra (commented):
I've taken a first look, there's loads of changes.

exe/Arguments.hs
wz1000: Backward compatibility - we don't want to break all existing `ghcide` command lines!

cabal.project
```diff
@@ -14,7 +14,15 @@ allow-newer:
...
+
 
 -- To ensure the build get the version with the fix for
 -- https://github.com/Avi-D-coder/implicit-hie/issues/17
 constraints: implicit-hie >= 0.1.2.0
+
+source-repository-package
+  type: git
+  location: https://github.com/wz1000/hiedb
+  tag: 7010bfd67f5f7735d46ce755112dea36cd693009
```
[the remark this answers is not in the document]
wz1000: No, I'm just in the process of finalizing the release.

See https://hackage.haskell.org/package/hiedb-0.1.0.0/candidate

src/Development/IDE/Core/Shake.hs
wz1000: Yes, but not very different from a cycle in shake dependencies?

src/Development/IDE/Core/Shake.hs
wz1000: No,  I don't think so

src/Development/IDE/Spans/Common.hs
wz1000: Same as `showSDoc`, but maybe I can re-add it for explicitness: https://hackage.haskell.org/package/ghc-8.10.1/docs/src/Outputable.html#showSDoc

exe/Arguments.hs
pepeiborra: But that can be handled in the parser, can't it?

src/Development/IDE/Core/Shake.hs
pepeiborra: But there is no cycle involved here, right? 

Put differently, why does this lookup+update need to be atomic?

src/Development/IDE/Spans/Common.hs
pepeiborra: Given how much the ghc api changes across versions, my vote is on explicitness

src/Development/IDE/Core/Shake.hs
wz1000: Hmm, I see

exe/Arguments.hs
wz1000: I couldn't get `optparse-applicative` to do this any other way.

pepeiborra: To do what, parse `--lsp`?

wz1000: Yeah

src/Development/IDE/Core/Shake.hs
wz1000: We still need to know whether the key is the map so that we can decide to block or not?

exe/Main.hs
```diff
@@ -62,6 +84,33 @@ ghcideVersion = do
...
+        hPutStrLn stderr $ "Error in worker, ignoring: " ++ show e
+        exitWith $ ExitFailure 1
+
+getHieDbLoc :: FilePath -> IO FilePath
+getHieDbLoc dir = do
+  let db = dirHash++"-"++takeBaseName dir++"-"++VERSION_ghc <.> "hiedb"
+      dirHash = B.unpack $ encode $ H.hash $ B.pack dir
+  cDir <- IO.getXdgDirectory IO.XdgCache cacheDir
+  createDirectoryIfMissing True cDir
+  pure (cDir </> db)
```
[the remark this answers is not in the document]
wz1000: Since `ghcide` now integrates the `hiedb` command line, I would say the way to interact with this database is via the `ghcide hiedb ...` subcommands.

So this lives in the usual ghcide cache directory, along with the `.hi`/`.hie` files.

src/Development/IDE/Core/RuleTypes.hs
wz1000: Moved to `AtPoint` and added.

src/Development/IDE/Core/Shake.hs
wz1000: Added

src/Development/IDE/Core/Shake.hs
wz1000: Fixed I think. This bit requires some careful consideration.

wz1000:
The situation with the `cached-deps` files is unsatisfactory right now - we want a nice way to detect if a file is "fake" in this manner (something better than `"cached-deps" `isInfixOf` filepath`, and we also want to compute only a given whitelist of rules for such files, like `GetHieAsts`, but not things like `GenerateCore`.

exe/Main.hs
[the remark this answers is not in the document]
pepeiborra: oh yeah, that's what I meant.

exe/Arguments.hs
pepeiborra: I find that using `ApplicativeDo` to construct parsers clarifies things massively. Or just ue a smart constructor

src/Development/IDE/Core/Rules.hs
```diff
@@ -127,88 +131,102 @@ defineEarlyCutOffNoFile f = defineEarlyCutoff $ \k file -> do
 ------------------------------------------------------------
 -- Exposed API
 
+lookupMod :: HieWriterChan -> FilePath -> ModuleName -> UnitId -> Bool -> MaybeT IdeAction Uri
+lookupMod dbchan hie_f mod uid boot
```
pepeiborra, line 135: doc comment

src/Development/IDE/Core/Rules.hs
```diff
@@ -127,88 +131,102 @@ defineEarlyCutOffNoFile f = defineEarlyCutoff $ \k file -> do
...
 getDefinition file pos = runMaybeT $ do
     ide <- ask
     opts <- liftIO $ getIdeOptionsIO ide
-    (HAR _ hf _ , mapping) <- useE GetHieAst file
+    (HAR _ hf _ _, mapping) <- useE GetHieAst file
     (ImportMap imports, _) <- useE GetImportMap file
     !pos' <- MaybeT (return $ fromCurrentPosition mapping pos)
-    AtPoint.gotoDefinition (getHieFile ide file) opts imports hf pos'
+    hiedb <- lift $ asks hiedb
+    hiedbChan <- lift $ asks hiedbWriter
```
pepeiborra, line 192: What sort of guarantees are given about the hiedb contents? Is it up-to-date?

I'm surprised it isn't managed by a Shake rule

src/Development/IDE/Core/Shake.hs
```diff
@@ -164,6 +180,11 @@ data ShakeExtras = ShakeExtras
     ,exportsMap :: Var ExportsMap
     -- | A work queue for actions added via 'runInShakeSession'
     ,actionQueue :: ActionQueue
+    , hiedb :: HieDb -- ^ Use only to read.
+    , hiedbWriter :: HieDbWriter -- ^ use to write
```
pepeiborra, line 184: Read and write what exactly? What is the hiedb used for? Source files in external packages or also in local packages?

src/Development/IDE/Core/Shake.hs
```diff
@@ -268,30 +302,55 @@ instance NFData v => NFData (Value v)
...
+
+        -- Update preserving 'monotonicity'
+        -- Don't want to overwrite a newer value with an older one
+        upd _new@(Failed False) old = old
+        upd new@(Failed True) (Failed False) = new
+        upd _new@(Failed True) old = old
+        upd new@(Stale _ _) Failed{} = new
+        upd new@(Stale v _) old@(Stale v' _) = if v >= v' then new else old
+        upd new@(Stale v _) old@(Succeeded v' _) = if v >= v' then new else old
+        upd new _old = new
```
pepeiborra, line 330: Could this be extracted to a `Monoid` instance?

src/Development/IDE/Core/Shake.hs
```diff
@@ -268,30 +302,55 @@ instance NFData v => NFData (Value v)
...
-    pure $ case v of
-        Succeeded ver v -> Just (v, mappingForVersion allMappings file ver)
-        Stale ver v -> Just (v, mappingForVersion allMappings file ver)
-        Failed -> Nothing
+lastValueIO :: IdeRule k v => ShakeExtras -> k -> NormalizedFilePath -> IO (Maybe (v, PositionMapping))
+lastValueIO s@ShakeExtras{positionMapping,persistentKeys,state} k file = do
+    let readPersistent = do
+          pmap <- readVar persistentKeys
+          mv <- runMaybeT $ do
+            liftIO $ Logger.logInfo (logger s) $ T.pack $ "LOOKUP UP PERSISTENT FOR" ++ show k
```
pepeiborra, line 314:
```suggestion
            liftIO $ Logger.logDebug (logger s) $ T.pack $ "LOOKUP UP PERSISTENT FOR" ++ show k
```

src/Development/IDE/Core/Shake.hs
pepeiborra: I don't see it, could you walk me through the cases?

src/Development/IDE/Core/Rules.hs
```diff
@@ -547,15 +557,17 @@ getImportMapRule = define $ \GetImportMap f -> do
 getBindingsRule :: Rules ()
 getBindingsRule =
   define $ \GetBindings f -> do
-    har <- use_ GetHieAst f
-    pure ([], Just $ bindings $ refMap har)
+    HAR _ _ rm kind <- use_ GetHieAst f
```
pepeiborra, line 560:
```suggestion
    HAR{hieKind=kind, refMap=rm} <- use_ GetHieAst f
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -547,15 +557,17 @@ getImportMapRule = define $ \GetImportMap f -> do
...
+      HieFromDisk _ -> pure ([], Nothing)
 
 getDocMapRule :: Rules ()
 getDocMapRule =
     define $ \GetDocMap file -> do
       (tmrTypechecked -> tc,_) <- useWithStale_ TypeCheck file
-      (hscEnv -> hsc,_) <-useWithStale_ GhcSessionDeps file
-      (refMap -> rf, _) <- useWithStale_ GetHieAst file
+      (hscEnv -> hsc,_) <- useWithStale_ GhcSessionDeps file
+      (HAR _ _ rf _, _) <- useWithStale_ GetHieAst file
```
pepeiborra, line 570:
```suggestion
      (HAR{refMap=rf}, _) <- useWithStale_ GetHieAst file
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -707,6 +743,17 @@ getModIfaceFromDiskRule = defineEarlyCutoff $ \GetModIfaceFromDisk f -> do
...
+                se@ShakeExtras{hiedb,hiedbWriter} <- getShakeExtras
+                mrow <- liftIO $ HieDb.lookupHieFileFromSource hiedb (fromNormalizedFilePath f)
+                case mrow of
+                  Just row | ms_hs_date ms <= HieDb.modInfoTime (HieDb.hieModInfo row) -> pure ()
+                  _ -> liftIO $ do
+                    mhf <- runIdeAction "GetModIfaceFromDisk" se $ runMaybeT $
+                      readHieFileFromDisk (ml_hie_file $ ms_location ms)
+                    case mhf of
+                      Nothing -> pure ()
+                      Just hf -> indexHieFile hiedbWriter ms f hf (mi_decl_docs $ hirModIface x)
```
pepeiborra, line 756: Could you expand on this comment? What is this side effect needed for? 

Consider extracting it to a helper function

src/Development/IDE/LSP/HoverDefinition.hs
```diff
@@ -29,16 +31,33 @@ gotoDefinition :: IdeState -> TextDocumentPositionParams -> IO (Either ResponseE
...
+      let filePath = toNormalizedFilePath' path
+      logInfo (ideLogger ide) $
+        "References request at position " <> T.pack (showPosition pos) <>
+        " in file: " <> T.pack path
+      Right . List <$> (runAction "references" ide $ refsAtPoint filePath pos)
+    Nothing -> pure $ Left $ ResponseError InvalidParams ("Invalid URI " <> T.pack (show uri)) Nothing
+
+wsSymbols :: IdeState -> WorkspaceSymbolParams -> IO (Either ResponseError (List SymbolInformation))
+wsSymbols ide (WorkspaceSymbolParams query _) = do
+  logInfo (ideLogger ide) $ "Workspace symbols request: " <> query
```
pepeiborra, line 52:
```suggestion
  logDebug (ideLogger ide) $ "Workspace symbols request: " <> query
```

src/Development/IDE/LSP/HoverDefinition.hs
```diff
@@ -29,16 +31,33 @@ gotoDefinition :: IdeState -> TextDocumentPositionParams -> IO (Either ResponseE
...
 gotoTypeDefinition = request "TypeDefinition" getTypeDefinition (MultiLoc []) MultiLoc
 hover          = request "Hover"      getAtPoint     Nothing      foundHover
 documentHighlight = request "DocumentHighlight" highlightAtPoint (List []) List
 
+references :: IdeState -> ReferenceParams -> IO (Either ResponseError (List Location))
+references ide (ReferenceParams (TextDocumentIdentifier uri) pos _ _) =
+  case uriToFilePath' uri of
+    Just path -> do
+      let filePath = toNormalizedFilePath' path
+      logInfo (ideLogger ide) $
```
pepeiborra, line 44:
```suggestion
      logDebug (ideLogger ide) $
```

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -23,35 +29,110 @@ import Development.IDE.Spans.Common
...
+  -> Position
+  -> FOIReferences
+  -> m [Location]
+referencesAtPoint hiedb nfp pos refs = do
+  let (names, foisLocs, exclude) = foiReferencesAtPoint nfp pos refs
+  locs <- forM names $ \name ->
+    case nameModule_maybe name of
+      Nothing -> pure []
+      Just mod -> do
+         rows <- liftIO $ search hiedb True (nameOccName name) (Just $ moduleName mod) (Just $ moduleUnitId mod) exclude
```
pepeiborra, line 100: What does `search` do?

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -23,35 +29,110 @@ import Development.IDE.Spans.Common
...
+         pure $ mapMaybe rowToLoc rows
+  typelocs <- forM names $ \name ->
+    case nameModule_maybe name of
+      Just mod | isTcClsNameSpace (occNameSpace $ nameOccName name) -> do
+        refs <- liftIO $ findTypeRefs hiedb (nameOccName name) (moduleName mod) (moduleUnitId mod)
+        pure $ mapMaybe typeRowToLoc refs
+      _ -> pure []
+  pure $ nubOrd $ foisLocs ++ concat locs ++ concat typelocs
```
pepeiborra, line 108: `foisLocs`: file of interest locations
`typeLocs`: type references
`locs`: non FOI locations ? 

if right, consider renaming `locs` for clarity

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -23,35 +29,110 @@ import Development.IDE.Spans.Common
...
+          toCurrentLocation mapping (Location uri range) = Location uri <$> toCurrentRange mapping range
+        in (names, adjustedLocs,map fromNormalizedFilePath $ HM.keys asts)
+
+referencesAtPoint
+  :: MonadIO m
+  => HieDb
+  -> NormalizedFilePath
+  -> Position
+  -> FOIReferences
+  -> m [Location]
```
pepeiborra, line 93: doc comments

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -134,61 +220,124 @@ atPoint IdeOptions{} hf (DKMap dm km) pos = listToMaybe $ pointCommand hf pos ho
...
    . MonadIO m
-  => (Module -> MaybeT m (HieFile, FilePath))
+  => HieDb
+  -> LookupModule m
   -> IdeOptions
   -> M.Map ModuleName NormalizedFilePath
   -> Position
-  -> HieASTs Type
+  -> HieASTs a
+  -> Maybe NormalizedFilePath
```
pepeiborra, line 271: doc comments please - `trustFile` is not very meaningful

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -134,61 +220,124 @@ atPoint IdeOptions{} hf (DKMap dm km) pos = listToMaybe $ pointCommand hf pos ho
...
-      modToLocation m = fmap (\fs -> Location (fromNormalizedUri $ filePathToUri' fs) zeroRange) $ M.lookup m imports
-    in mapMaybeM (either (pure . modToLocation) $ nameToLocation getHieFile) ns
+      modToLocation m = fmap (\fs -> pure $ Location (fromNormalizedUri $ filePathToUri' fs) zeroRange) $ M.lookup m imports
+    in fmap concat $ mapMaybeM (either (pure . modToLocation) $ nameToLocation hiedb lookupModule trustFile) ns
 
 -- | Given a 'Name' attempt to find the location where it is defined.
-nameToLocation :: Monad f => (Module -> MaybeT f (HieFile, String)) -> Name -> f (Maybe Location)
-nameToLocation getHieFile name = fmap (srcSpanToLocation =<<) $
+nameToLocation :: MonadIO m => HieDb -> LookupModule m -> Maybe NormalizedFilePath -> Name -> m (Maybe [Location])
+nameToLocation hiedb lookupModule trustFile name = runMaybeT $
```
pepeiborra, line 282: doc comments for function arguments please

src/Development/IDE/Spans/AtPoint.hs
```diff
@@ -134,61 +220,124 @@ atPoint IdeOptions{} hf (DKMap dm km) pos = listToMaybe $ pointCommand hf pos ho
...
+      case erow of
+        [] -> MaybeT $ pure Nothing
+        xs -> lift $ mapMaybeM (runMaybeT . defRowToLocation lookupModule) xs
+
+defRowToLocation :: Monad m => LookupModule m -> Res DefRow -> MaybeT m Location
+defRowToLocation lookupModule (row:.info) = do
+  let start = Position (defSLine row - 1) (defSCol row - 1)
+      end   = Position (defELine row - 1) (defECol row - 1)
+      range = Range start end
+  traceM $ "DEFROW TO LOC ******************" ++ show (range, modInfoSrcFile info)
```
pepeiborra, line 315:
```suggestion
--  traceM $ "DEFROW TO LOC ******************" ++ show (range, modInfoSrcFile info)
```

src/Development/IDE/Core/Rules.hs
```diff
@@ -570,7 +582,32 @@ getDocMapRule =
...
+  pure $ HAR (hie_module res) (hie_asts res) refmap (HieFromDisk res)
+
+readHieFileForSrcFromDisk :: NormalizedFilePath -> MaybeT IdeAction HieFile
+readHieFileForSrcFromDisk file = do
+  db <- asks hiedb
+  log <- asks $ L.logInfo . logger
+  row <- MaybeT $ liftIO $ HieDb.lookupHieFileFromSource db $ fromNormalizedFilePath file
+  let hie_loc = HieDb.hieModuleHieFile row
+  liftIO $ log $ "LOADING HIE FILE :" <> T.pack (show file)
+  readHieFileFromDisk hie_loc
```
pepeiborra, line 598: I am trying to understand what this buys us. Ghcide already reuses `.hie` files from the cache dir at startup if they are up-to-date, so in what cases does this make a difference?

src/Development/IDE/Core/Rules.hs
```diff
@@ -570,7 +582,32 @@ getDocMapRule =
...
+persistentHieFileRule :: Rules ()
+persistentHieFileRule = addPersistentRule GetHieAst $ \file -> runMaybeT $ do
+  res <- readHieFileForSrcFromDisk file
+  let refmap = generateReferencesMap . getAsts . hie_asts $ res
+  pure $ HAR (hie_module res) (hie_asts res) refmap (HieFromDisk res)
+
+readHieFileForSrcFromDisk :: NormalizedFilePath -> MaybeT IdeAction HieFile
+readHieFileForSrcFromDisk file = do
+  db <- asks hiedb
+  log <- asks $ L.logInfo . logger
```
pepeiborra, line 594:
```suggestion
  log <- asks $ L.logDebug . logger
```

pepeiborra (commented):
Second pass done.

Can you write a high-level comment describing the integration with hiedb?

In particular, regarding the paragraph below, for non FOIs, what things are now obtained from hiedb instead of from `.hie` files?

> Under this paradigm, ghcide acts as an indexing service for hiedb, generating .hi and .hie files which are indexed and saved in the database, available for all future queries, even across restarts. A local cache of .hie files/typechecked modules is maintained on top of this to answer queries for the files the user is currently editing, while non-local information about other files in the project is accessed through the database.

pepeiborra:
> The situation with the `cached-deps` files is unsatisfactory right now - we want a nice way to detect if a file is "fake" in this manner (something better than `"cached-deps" `isInfixOf` filepath`, and we also want to compute only a given whitelist of rules for such files, like `GetHieAsts`, but not things like `GenerateCore`.

Sounds like extending the `IsFileOfInterest` rule and its result would be good for clarity and allow a better mechanism down the line

pepeiborra:
> First, the user must index the `.hie` files for their dependencies by running the command `ghcide hiedb index $DIR` in the root directory of their project.

What should `$DIR` point to? 

If it's just the same folder where `hie.yaml` lives, can `ghcide` do it in the background?

src/Development/IDE/Spans/AtPoint.hs
wz1000: It returns all references from the database, except for the ones that occur in `exclude`.

src/Development/IDE/Spans/AtPoint.hs
wz1000: Correct.

src/Development/IDE/Core/Rules.hs
wz1000: > Ghcide already reuses .hie files from the cache dir at startup if they are up-to-date

It doesn't, at least not for things like hover/definition etc.

With the persistent rule stuff we get:

- Instant availability of select IDE features (hover/definition...) on startup, even before `cabal configure` etc. finish
- If the project doesn't compile on startup, we can still use the above features from the stale `.hie` file.

src/Development/IDE/Core/Rules.hs
wz1000: No, the invariant is that it is up to date when `pendingIndexes` is empty and we've received no more edits.

src/Development/IDE/Core/Shake.hs
wz1000: Both. References are looked up in local packages, and definitions are looked up in both.

src/Development/IDE/Core/Rules.hs
pepeiborra: Can we document this somewhere?
What depends on this invariant and how is it enforced?

wz1000: OK. It's enforced by us queuing up an index operation every time a `.hie` file is generated - which is every time the `GetModIface` rule is recomputed. Also, when reading a `.hi` file from disk, we check the database to ensure that the the corresponding `.hie` file has been indexed, and if not, we re-index it. This maintains database consistency if ghcide is killed before all the indexing completes.

Nothing "depends" on the invariant as such, since we are after eventual consistency - the results may be out of date at any given time.

src/Development/IDE/Core/Rules.hs
pepeiborra: That's awesome. 

But surely ghcide will block until the cradle is loaded, as it usually happens on a Notification?

src/Development/IDE/Core/Rules.hs
pepeiborra: It would be good to write this down in a comment somewhere

Do we make any effort to avoid redundant re-indexings ?

src/Development/IDE/Core/Shake.hs
```diff
@@ -268,30 +302,55 @@ instance NFData v => NFData (Value v)
...
+    let readPersistent = do
+          pmap <- readVar persistentKeys
+          mv <- runMaybeT $ do
+            liftIO $ Logger.logDebug (logger s) $ T.pack $ "LOOKUP UP PERSISTENT FOR" ++ show k
+            f <- MaybeT $ pure $ HMap.lookup (Key k) pmap
+            dv <- MaybeT $ runIdeAction "lastValueIO" s $ f file
+            MaybeT $ pure $ fromDynamic dv
+          modifyVar state $ \hm -> case mv of
+            Nothing -> pure (HMap.insertWith upd (file,Key k) (Failed True) hm,Nothing)
+            Just v -> pure (HMap.insertWith upd (file,Key k) (Stale Nothing (toDyn v)) hm, Just (v,zeroMapping))
```
pepeiborra, line 320: This returns `zeroMapping` because we cannot do better, right?

src/Development/IDE/Core/Shake.hs
wz1000: I think its best to keep this near the `insertWith` because any such monoid would be non-commutative, so it becomes confusing figuring out which the "new" one, and which is the "old" one.

src/Development/IDE/Core/Shake.hs
wz1000: Yes, with a bit more work we can compute the position mapping from the diff.

wz1000:
> What should $DIR point to?

What ever `-hiedir` was set to.

You would put something like the following in a `cabal.project(.local)`:

```
package *
    ghc-options: -fwrite-ide-info -hiedir <some-directory>
```

One limitation is that `<some-directory>` has to be an absolute path, or `cabal` will strew `.hie` files in random places.

I don't know if there is a way for ghcide to pass such options to cabal/stack, but it might not be a good idea anyway since it would force recompilation of all dependencies. If it is possible, we can have a special `ghcide` command that does this and indexes the resulting files.

/cc @fendor

src/Development/IDE/Core/Shake.hs
pepeiborra: Add a TODO then!

src/Development/IDE/Core/Rules.hs
wz1000: > But surely ghcide will block until the cradle is loaded, as it usually happens on a Notification?

Ah, this is indeed the behavior in vscode. I didn't notice because my client sends the request immediately after initialize if the server isn't already running, so I was getting instant responses most of the time.

I think we can serialize notifications with respect to each other, but handle requests concurrently without blocking?

pepeiborra: We decided to process notifications synchronously to ensure consistency after seeing test failures showing real inconsistency issues around e.g. completions. I don't see how to avoid such inconsistencies if we handle requests concurrently with notifications.

pepeiborra:
> > What should $DIR point to?
> 
> What ever `-hiedir` was set to.
> 
> You would put something like the following in a `cabal.project(.local)`:
> 
> ```
> package *
>     ghc-options: -fwrite-ide-info -hiedir <some-directory>
> ```
> 
> One limitation is that `<some-directory>` has to be an absolute path, or `cabal` will strew `.hie` files in random places.
> 
> I don't know if there is a way for ghcide to pass such options to cabal/stack, but it might not be a good idea anyway since it would force recompilation of all dependencies. If it is possible, we can have a special `ghcide` command that does this and indexes the resulting files.
> 
> /cc @fendor

This might just about work, but in the long term, do we want `.hie` files to be part of ghc package databases, just like interface files are/

exe/Arguments.hs
googleson78: The field is still used in `Main.hs` though, isn't it?

Anyway, something like this should work:
```haskell
diff --git a/exe/Arguments.hs b/exe/Arguments.hs
index ddffd35..afa56e6 100644
--- a/exe/Arguments.hs
+++ b/exe/Arguments.hs
@@ -11,8 +11,7 @@ type Arguments = Arguments' IdeCmd
 data IdeCmd = Typecheck [FilePath] | DbCmd Command | LSP
 
 data Arguments' a = Arguments
-    {argLSP :: Bool
-    ,argsCwd :: Maybe FilePath
+    {argsCwd :: Maybe FilePath
     ,argsVersion :: Bool
     ,argsShakeProfiling :: Maybe FilePath
     ,argsTesting :: Bool
@@ -30,7 +29,7 @@ getArguments = execParser opts
 
 arguments :: Parser Arguments
 arguments = Arguments
-      <$> switch (long "lsp" <> help "Start talking to an LSP server")
+      <$ switch (long "lsp" <> help "Start talking to an LSP server" <> internal)
       <*> optional (strOption $ long "cwd" <> metavar "DIR" <> help "Change to this directory")
       <*> switch (long "version" <> help "Show ghcide and GHC versions")
       <*> optional (strOption $ long "shake-profiling" <> metavar "DIR" <> help "Dump profiling reports to this directory")
```
The `internal` is so that `--lsp` doesn't show up in help texts

src/Development/IDE/Core/Rules.hs
wz1000: https://github.com/mpickering/hie-bios/pull/265 will make startup much faster when it lands, so we will get most of the benefits of quick startup even without async notifications. Do we still want to keep this mechanism to provide code navigation when the project doesn't compile on startup, or should I just remove it?

If we are keeping it, I think it needs a re-design. I'm thinking another variant of `define` in which we can return a stale value if the regular codepath fails.

pepeiborra: I think you've answered yourself! Get rid of this version and leave the redesign for future work

pepeiborra:
@wz1000 when do you expect to be able to push this forward?

src/Development/IDE/Core/Rules.hs
pepeiborra: I think that my assumption that ghcide blocks on a FileOpened Notification might be wrong. 
The handler calls `restartShakeSession` but doesn't wait on the results, so it might not be actually blocking. 
The open telemetry stuff will make it very easy to find out whether it blocks or not.

wz1000:
Unfortunately I won't be able to get back to this for another 1-2 weeks at least.

Meanwhile, @jhrcek has very kindly volunteered to help out with tests and documentation for the hiedb library itself, before its hackage release.

wz1000:
I've removed all the stuff to do with non-project dependencies for now, will re-add in another PR.

src/Development/IDE/Core/Shake.hs
wz1000: There are three cases:

1a) The value isn't in the store, and we can't (or fail to) compute a stale persistent value - we want to wait for the rule to compute a fresh value (matching the previous semantics)
1b) The value isn't in the store but we can compute a stale persistent value - return the value
2) The value is in the store - we can just return it. 

(sorry for the delay, hope you will be able to page this back in)

src/Development/IDE/Core/Rules.hs
wz1000: removed this bit for now.

src/Development/IDE/Core/Rules.hs
wz1000: > Do we make any effort to avoid redundant re-indexings ?

Yes, we only re-index only if one of the following conditions hold:

-  The `.hi` file had to be regenerated
- The `.hs` file is newer than the `.hie` file.

pepeiborra:

```
git fetch ghcide
```
Thanks!

PR #900, opened 2020-11-09
Record completions snippets

gdevanla:
This PR addresses the issue in https://github.com/haskell/haskell-language-server/issues/333.

record_snippet_demo2-2


src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -343,17 +363,63 @@ localCompletionsForParsedModule pm@ParsedModule{pm_parsed_source = L _ HsModule{
...
+
 ppr :: Outputable a => a -> T.Text
 ppr = T.pack . prettyPrint
 
 newtype WithSnippets = WithSnippets Bool
 
 toggleSnippets :: ClientCapabilities -> WithSnippets -> CompletionItem -> CompletionItem
 toggleSnippets ClientCapabilities { _textDocument } (WithSnippets with) x
-  | with && supported = x
+  | with || supported = x
```
gdevanla, line 419: Note that I had to make this change to let snippets show up. I think we need to make a more relevant change to the code https://github.com/haskell/ghcide/blob/dcf680402722e7660377bf7da3fca27c2bde988f/src/Development/IDE/Plugin/Completions.hs#L134-L135

wz1000:
> How do I add tests that test non-local record definitions and completions. Please could you point to places where I can look for some examples of those.

See https://github.com/haskell/ghcide/blob/master/test/exe/Main.hs#L2662

> How does one go about debugging/printing things inside ghcide

You can use the logging mechanism, `Debug.Trace` or just `hPutStrLn stderr`, whichever is convenient.

gdevanla:
> > How do I add tests that test non-local record definitions and completions. Please could you point to places where I can look for some examples of those.
> 
> See https://github.com/haskell/ghcide/blob/master/test/exe/Main.hs#L2662
> 
> > How does one go about debugging/printing things inside ghcide
> 
> You can use the logging mechanism, `Debug.Trace` or just `hPutStrLn stderr`, whichever is convenient.

@wz1000 So, I used `trace` inside say `findRecordCompl` and ran tests. i did not see any output. i also ran the code through hls and still did not see any output in` hls.log`. Does the output get written to some temp file? I cannot see any output on the terminal.  What would be a typical workflow to print and debug stuff? Or should I wrap this functions in `IO` to see the trace statements?

Than

pepeiborra:
For tests, check the lsp-test [haddocks](http://hackage.haskell.org/package/lsp-test-0.11.0.7/docs/Language-Haskell-LSP-Test.html#t:SessionConfig). By default stderr is not echoed and you need to change the settings or define an env var

For hls, I don't know how the log file works. I usually find stderr is logged to the haskell Output window in VSCode

gdevanla:
Thanks @pepeiborra. I will explore those options. For now, I was able to figure out what was happening by just staring at the code!

I managed to add tests (also had to extend `completionTest` to compare the `insertText`).

One test I would like to add and which I am not able to figure out is,  declare a Record in one module, and check if snippet gets generated in another module.  I see 2 options.

1. Have a test declare 2 modules and then create a test. (by importing the record from one module onto another)
2. If there an `record` in standard haskell library(like Data.List, say), then I can use that in tests. That way I do not have to declare an separate module.  I looked around but failed.

wz1000:
> If there an record in standard haskell library(like Data.List, say), then I can use that in tests. That way I do not have to declare an separate module. I looked around but failed.

How about this? If there an record in standard haskell library(like Data.List, say), then I can use that in tests. That way I do not have to declare an separate module. I looked around but failed.

gdevanla:
Thanks, @wz1000. I was able to use examples from that library to add the tests I wanted.

gdevanla:
Looks like this failure is because older version do not have this type: https://github.com/haskell/ghcide/pull/900/checks?check_run_id=1402628784.

Please can I have some suggestions on how to fix this in prior versions?

wz1000:
Use `ConDeclField` instead of `SrcSpanLess LConDeclField`

gdevanla:
Also, I am a not able to address this hlint failure.

https://dev.azure.com/digitalasset/ghcide/_build/results?buildId=60930&view=logs&j=8c509560-e3d6-573f-8a24-ae9e290d8e4f&t=fdec7397-253f-5ca6-ce5e-10d811d2034a&l=16

```
src/Development/IDE/Plugin/Completions/Logic.hs:271:10-26: Suggestion: Use tuple-section
Found:
  \ x -> (x, mempty)
Perhaps:
  (, mempty)
Note: may require `{-# LANGUAGE TupleSections #-}` adding to the top of the file
```

When I use this suggestion, then I get a hlint warning regarding use of `TupleSections`, which seems circular.

wz1000:
You don't need to add `TupleSections`, it is already enabled in `default-extensions`.

wz1000:
Why was this closed?

gdevanla:
I think I closed the wrong issue!

pepeiborra:
Could you rebase this on top of master to pick up some fixes to the GitHub Actions CI?

gdevanla:
I am not able to figure out what the 2 faiures are. The logs don't say much. Is there somewhere else I can take a look.

```

@azure-pipelines
haskell.ghcide Failing after 120m — Build #20201121.5 failed
Details
@azure-pipelines
haskell.ghcide (ghcide_stack_windows) Cancelled after 120m — ghcide_stack_windows was canceled
```

pepeiborra:
The Windows CI jobs are unreliable (because GHC is unreliable in Windows)

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -342,17 +369,71 @@ localCompletionsForParsedModule pm@ParsedModule{pm_parsed_source = L _ HsModule{
...
+    _ -> []
+
+findRecordCompl :: ParsedModule -> T.Text -> Located (IdP GhcPs) -> HsDataDefn GhcPs -> [CompItem]
+findRecordCompl pmod mn lname dd = catMaybes name_type''
+    where
+        conDecl = unLoc <$> dd_cons dd
+        h98 = catMaybes $ findH98 <$> conDecl
+        findH98 conDecl = case conDecl of
+          ConDeclH98{..} -> Just (unLoc con_name, con_args)
+          ConDeclGADT{} -> Nothing  -- TODO: Expand this out later
```
pepeiborra, line 386: Are you planning to look at this TODO?

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -342,17 +369,71 @@ localCompletionsForParsedModule pm@ParsedModule{pm_parsed_source = L _ HsModule{
...
+
 ppr :: Outputable a => a -> T.Text
 ppr = T.pack . prettyPrint
 
 newtype WithSnippets = WithSnippets Bool
 
 toggleSnippets :: ClientCapabilities -> WithSnippets -> CompletionItem -> CompletionItem
 toggleSnippets ClientCapabilities { _textDocument } (WithSnippets with) x
-  | with && supported = x
+  | with || supported = x
```
pepeiborra, line 433:
```suggestion
  | with = x
```

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -342,17 +369,71 @@ localCompletionsForParsedModule pm@ParsedModule{pm_parsed_source = L _ HsModule{
...
   | otherwise = x { _insertTextFormat = Just PlainText
-                  , _insertText       = Nothing
-                  }
+                , _insertText       = Nothing
+                }
   where supported = Just True == (_textDocument >>= _completion >>= _completionItem >>= _snippetSupport)
```
pepeiborra, line 437:
```suggestion
                , _insertText       = Nothing
                }
```

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -45,6 +47,13 @@ import Development.IDE.Spans.Common
 import Development.IDE.GHC.Util
 import Outputable (Outputable)
 import qualified Data.Set as Set
+import Maybes (catMaybes)
```
pepeiborra, line 50:
```suggestion
import Data.Maybe (catMaybes)
```

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -600,3 +680,37 @@ prefixes =
...
+        types = map (conLikeFieldType dc . flLabel) flds
+        ctxStr = T.pack . showGhc $ name
+        fld_pairs = [(T.pack . unpackFS . flLabel $ x, T.pack . showGhc $ y) | x <- flds | y <- types]
+    in
+        Just (ctxStr, fld_pairs)
+        --[(showGhc name, [("x", "y")])]
+safeTyThingForRecord _ = Nothing
+
+mkRecordSnippetCompItem :: T.Text -> [(T.Text, T.Text)] -> T.Text -> SpanDoc -> CompItem
+mkRecordSnippetCompItem ctxStr compl mn docs = r
```
pepeiborra, line 699: Here `compl` contains a tuple (name, type) for every field, but It looks like the type is not used anywhere, so why bother to pass it through (and why bother collecting it)?

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -600,3 +680,37 @@ prefixes =
...
+          , insertText = buildSnippet
+          , importedFrom = importedFrom
+          , typeText = Nothing
+          , label = ctxStr
+          , isInfix = Nothing
+          , docs = docs
+          , isTypeCompl = False
+          }
+
+      t = zip compl ([1..]::[Int])
```
pepeiborra, line 712: I'd prefer a better name here, `t` is a bit too short for me

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -342,17 +369,71 @@ localCompletionsForParsedModule pm@ParsedModule{pm_parsed_source = L _ HsModule{
...
+    --recordCompls = localRecordSnippetProducer pm thisModName
+
+findRecordCompl' :: ParsedModule -> T.Text -> TyClDecl GhcPs -> [CompItem]
+findRecordCompl' pmod mn x = case x of
+    DataDecl {tcdLName, tcdDataDefn} -> findRecordCompl pmod mn tcdLName tcdDataDefn
+    _ -> []
+
+findRecordCompl :: ParsedModule -> T.Text -> Located (IdP GhcPs) -> HsDataDefn GhcPs -> [CompItem]
+findRecordCompl pmod mn lname dd = catMaybes name_type''
+    where
```
pepeiborra, line 381: this function would write quite well as a big list comprehension

pepeiborra (changes requested):
Thank you, this is awesome. I have a minor nitpick, can you change the templates to suggest holes? 
That is, given a record `data Person = Person { name :: String}`: 
 - your action currently suggests `Person{name = name}`,
 - instead, it would be better to suggest `Person{name = _name}`.

I have left some other comments and suggestions.

src/Development/IDE/Plugin/Completions/Logic.hs
gdevanla: I was not planning to as part of this PR.  I can take a look and see if it is easy.  Would it be OK to do it as part of another PR?

src/Development/IDE/Plugin/Completions/Logic.hs
gdevanla: I am not sure what you mean by this. Could you elaborate a little. Or may be also point to other examples? Thanks

src/Development/IDE/Plugin/Completions/Logic.hs
gdevanla: I left it here since I was not sure if it would be nice to attach `type` information in the code snippet.

This is from your suggestion:

> That is, given a record `data Person = Person { name :: String}`:
> 
>     * your action currently suggests `Person{name = name}`,
> 
>     * instead, it would be better to suggest `Person{name = _name}`.

Should we consider, 
```
Person {name = (_name:String)}  
```
With this the user has information about the type as well. I do understand that the `type` is also displayed as part of `codeLens` documentation. So, in that sense, it could be redundant.

For now, I will go with your suggestion, unless you feel adding `type` would be an improvement.

pepeiborra: My feeling is that the type is redundant, and including it will not be trivial for polymorphic types. Moreover, the type might include identifiers that are not in scope, creating a build error.

gdevanla:
The code clean up suggestions are great. Thanks for taking the time to show me idiomatic uses. Appreciate it!

src/Development/IDE/Plugin/Completions/Logic.hs
pepeiborra: I think this function could be rewritten quite nicely in the following style:
```
findRecordCompl pmod mn lname dd = 
  [ mkRecordSnippetCompItem ctxStr flds mn doc
  | ConDeclH98{..} <- dd_cons dd
  , con_arg <- con_args
  , Just name_flds <- [getFlds con_fields]
  , ...
  ] 
```

src/Development/IDE/Plugin/Completions/Logic.hs
gdevanla: Sounds good. I will go with your suggestion. I think it is an improvement over what I have currently.

gdevanla: Any ideas on how to handle field names which already have a prefixed `_`.

For example, `Person { _name:: String }`.  Should we leave it as is or add another `__name`?

src/Development/IDE/Plugin/Completions/Logic.hs
gdevanla: I see. Ok. Let me try!

gdevanla (commented):
I think I have addressed all the requested changes.

I am happy to recieve further suggestions on improving how I implemented the list comprehension logic in findCompletions.

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -371,57 +369,31 @@ localCompletionsForParsedModule pm@ParsedModule{pm_parsed_source = L _ HsModule{
...
-                          Nothing -> Nothing
-
-        name_flds :: [(RdrName, [ConDeclField GhcPs])]
-        name_flds = catMaybes $ decompose <$> h98
-
+        result = [mkRecordSnippetCompItem (T.pack . showGhc . unLoc $ con_name) field_labels mn doc
+                 | ConDeclH98{..} <- unLoc <$> dd_cons tcdDataDefn
+                 , Just  con_details <- [getFlds con_args]
+                 , field_names <- [mapMaybe extract con_details]
+                 , field_labels <- [T.pack . showGhc . unLoc <$> field_names]
```
pepeiborra, line 379:
```suggestion
                 , let field_names = mapMaybe extract con_details
                 , let field_labels = T.pack . showGhc . unLoc <$> field_names
```

src/Development/IDE/Plugin/Completions/Logic.hs
pepeiborra: Not a big deal

pepeiborra (approved):
Many thanks for incorporating the feedback, this is ready to merge now

PR #902, opened 2020-11-11
Hide warnings which user hasn't explicitly enabled

georgefst:
Patches #815. 

This seems to result in the right behaviour, but I don't quite understand the logic in `hideDiag`. @serras Why exactly do we match on the existence of a tag in the diagnostic, without caring which tag it is?

Closes haskell/haskell-language-server#572.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -332,12 +332,12 @@ upgradeWarningToError (nfp, sh, fd) =
   warn2err = T.intercalate ": error:" . T.splitOn ": warning:"
 
 hideDiag :: DynFlags -> (WarnReason, FileDiagnostic) -> (WarnReason, FileDiagnostic)
-hideDiag originalFlags (Reason warning, (nfp, sh, fd))
+hideDiag originalFlags (Reason warning, (nfp, _sh, fd))
   | not (wopt warning originalFlags)
   = if null (_tags fd)
        then (Reason warning, (nfp, HideDiag, fd))
             -- keep the diagnostic if it has an associated tag
-       else (Reason warning, (nfp, sh, fd{_severity = Just DsInfo}))
+       else (Reason warning, (nfp, HideDiag, fd{_severity = Just DsInfo}))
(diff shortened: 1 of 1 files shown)


serras:
The way it works right now was part of the discussion of #815, and related to the fact that in LSP you can only mark an element as "unused" if it comes with an associated diagnostic. So it was decided to enable always the warning about unused elements, and demote them to `Info` if not enabled by the user. That is the logic behind the weird `hideDiag`: hide it if it was not present in the original set of extensions *unless* it is an "unused warning" (so the tags are not empty), then we only demote it.

If you really think that this behavior is wrong, and unused variables should only appear if the user really asked for it, the patch should be the following:
- Do not add `unnecessaryDeprecationWarningFlags` to the list of enabled warning any more by default (just remove the call to `enableUnnecessaryAndDeprecationWarnings` and that function altogether);
- The `hideDiag` would then not need the `if` at all, just put the `HideDiag` on if it was not in the list of original warnings.

The current patch only works around the implementation, but if we really don't want the functionality to be on even if the user didn't ask for it, just let's never enable it!

ndmitchell:
CC @alanz as our resident LSP expert. It's really sad that we can't both have unused variables come up in grey, but at the same time not have them report to the info tab. TypeScript in VS Code does exactly that, but I'm not sure if they are doing extension-side stuff.

alanz:
Typescript might be using the new in LSP 3.16 feature of [Semantic Tokens](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_semanticTokens) which is a very complex feature, but basically allows you to tweak the colour of any token at will.

Another thing is you might be able to disable the display of certain categories of diagnostic, on a client-by-client basis.

Personally I love the dimmed out unused stuff, but I would say default to having it on, and provide a means to disable it for the set of users who would rather not have it.

Which ties back to having a standardized way to expose plugin settings.

georgefst:
> This seems to result in the right behaviour

As has now been pointed out, it really doesn't. Turns out I messed up my versions when testing this.

georgefst:
Closing in favour of #907.

PR #904, opened 2020-11-11
Avoid calling kick explicitly

pepeiborra:
This change leverages that rules are rerun by shakeRunDatabase.
It allows users of ghcide as a library to use their own kick

### DIFF
--- a/exe/Main.hs
+++ b/exe/Main.hs
@@ -50,6 +50,7 @@ import qualified Data.HashMap.Strict as HashMap
 import qualified Data.Aeson as J
 
 import HIE.Bios.Cradle
+import Development.IDE (action)
 
 ghcideVersion :: IO String
 ghcideVersion = do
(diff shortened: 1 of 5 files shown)


wz1000:
I was skeptical of this in the `hlint` plugin as well. Do we really want to run `kick` so often? Why should we run kick when we actually want to compute an unrelated rule? If we are computing a rule, its dependencies will be checked for freshness anyway.

Can we at least deprioritize `kick`?

The documentation for `action` explicitly mentions "only cheap operations should be performed"

pepeiborra:
This is a pure refactoring, isn't it? My recollection is that we always call `kick` on a `shakeRestart`.

wz1000:
Yes, but I'm concerned about `runAction`/`shakeEnqueue`.

pepeiborra:
What is your concern? There is no change in `runAction` / `shakeEnqueue` as far as I can see, since neither of them calls `shakeRunDatabase`. 

This change is motivated by a custom LSP server we have at work, half-built on top of ghcide 0.0.5, which uses this approach to install a custom `kick`.

wz1000 (approved):

PR #905, opened 2020-11-12
Update implicit-hie

Avi-D-coder:
> # Changelog for implicit-hie-cradle
> 
> ## 0.3.0.0
> Match `implicit-hie`'s cradle type ordering.
> This is technically a breaking change.

### DIFF
--- a/ghcide.cabal
+++ b/ghcide.cabal
@@ -90,7 +90,7 @@ library
         ghc-paths,
         cryptohash-sha1 >=0.11.100 && <0.12,
         hie-bios >= 0.7.1 && < 0.8.0,
-        implicit-hie-cradle >= 0.2.0.1 && < 0.3,
+        implicit-hie-cradle >= 0.2.0.1 && < 0.4,
         base16-bytestring >=0.1.1 && <0.2
     if os(windows)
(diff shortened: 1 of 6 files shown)


Avi-D-coder:
`allow-newer` is needed until https://github.com/haskell/ghcide/pull/905 lands.
Should we just wait and upgrade ghcide then?

ghcide.cabal
```diff
@@ -90,7 +90,7 @@ library
         ghc-paths,
         cryptohash-sha1 >=0.11.100 && <0.12,
         hie-bios >= 0.7.1 && < 0.8.0,
-        implicit-hie-cradle >= 0.2.0.1 && < 0.3,
+        implicit-hie-cradle >= 0.2.0.1 && < 0.4,
```
jneira, line 93: Maybe we should remove the upper bound? If we are no gonna change the api frequently, maybe it would be handy

Avi-D-coder: The API will change if `hie-bios`'s changes, I am more conformable with a upper bound.

jneira (approved):
Many thanks for updating the lib

PR #906, opened 2020-11-12
Parenthesize operators when exporting

berberman:
## Subject
A tiny fix to export operators properly. If I understand correctly, identifiers can be either totally symbolic or alpha num. And it might be safe to use `type ...` in the export list:

> The extension -XExplicitNamespaces is implied by -XTypeOperators and (for some reason) by -XTypeFamilies.

we could define those operators only when TypeOperators is enabled.

Closes haskell/haskell-language-server#589. Perhaps I should add some tests :)

## Screenshot
Peek 2020-11-12 18-37

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -382,7 +382,7 @@ suggestExportUnusedTopBinding srcOpt ParsedModule{pm_parsed_source = L _ HsModul
 
     parenthesizeIfNeeds :: Bool -> T.Text -> T.Text
     parenthesizeIfNeeds needsTypeKeyword x 
-      | any (`elem` opLetter) . T.unpack $ x = (if needsTypeKeyword then "type " else "") <> "(" <> x <>")"
+      | T.head x `elem` opLetter = (if needsTypeKeyword then "type " else "") <> "(" <> x <>")"
       | otherwise = x
 
     getLocatedRange :: Located a -> Maybe Range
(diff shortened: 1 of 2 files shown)


jneira:
> Perhaps I should add some tests 

Even one of them would be great :wink:

berberman:
I created them separately one by one, consistent with former tests.

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -377,6 +377,14 @@ suggestExportUnusedTopBinding srcOpt ParsedModule{pm_parsed_source = L _ HsModul
         _ -> False
     needsComma _ _ = False
 
+    opLetter :: String
+    opLetter = ":!#$%&*+./<=>?@\\^|-~"
+
+    parenthesizeIfNeeds :: Bool -> T.Text -> T.Text
+    parenthesizeIfNeeds needsTypeKeyword x 
+      | any (`elem` opLetter) . T.unpack $ x = (if needsTypeKeyword then "type " else "") <> "(" <> x <>")"
```
pepeiborra, line 385:
```suggestion
      | any (`elem` opLetter) . take 1 . T.unpack $ x = (if needsTypeKeyword then "type " else "") <> "(" <> x <>")"
```

pepeiborra (approved):
Thank you for fixing this and writing a thorough test suite!

src/Development/IDE/Plugin/CodeAction.hs
berberman: Yes, looking at the first char would be enough.

PR #907, opened 2020-11-14
Do not enable every "unnecessary" warning by default

serras:
This is the result of the discussion in https://github.com/haskell/haskell-language-server/issues/572. As such, it closes #902 and supersedes it. Most of the warnings are anyway enabled by `-Wall`, which is taken as a good practice in the community.

@georgefst @jneira you might be interested in this.

### DIFF
--- a/src/Development/IDE/Core/Compile.hs
+++ b/src/Development/IDE/Core/Compile.hs
@@ -136,7 +136,6 @@ typecheckModule (IdeDefer defer) hsc keep_lbls pm = do
         modSummary' <- initPlugins hsc modSummary
         (warnings, tcm) <- withWarnings "typecheck" $ \tweak ->
             tcRnModule hsc keep_lbls $ enableTopLevelWarnings
-                                     $ enableUnnecessaryAndDeprecationWarnings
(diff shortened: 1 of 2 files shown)


georgefst:
LGTM. I tend to enable `-Wall` everywhere and only disable individual warnings for very good reason. I believe this is pretty much standard practice.

serras:
It seems that tests are not passing due to:

```
highlight
     value:                                                                                                FAIL
       Exception: Timed out waiting to receive a message from the server.
```

but I have no idea how to fix this, or if this is really a problem. It would be great if I could get a helping hand here :)

jneira:
ser `LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true` before running the test might help

wz1000:
I think some of the tests rely on diagnostics which are no longer produced (in order to check that typechecking has completed), so you either have to enable the appropriate warnings for those tests, or add some other kind of construct which will result in some diagnostic.

serras:
Thanks @wz1000, I've just pushed what I think is the right fix.

pepeiborra (approved):
Thanks Alejandro!

PR #908, opened 2020-11-16
Remove allow-newer

jneira:
* In my machine stack finds a build plan without allow-newer, let's see what have to say ci


ndmitchell (approved):
Always good to avoid where we can

PR #912, opened 2020-11-17
Bump up implicit-hie-cradle lower bound

jneira:
* To ensure we get the desired version


jneira:
The build is failing due to https://github.com/actions/setup-haskell/issues/44, we'll have to wait to the fix

pepeiborra (approved):

pepeiborra:
> The build is failing due to [actions/setup-haskell#44](https://github.com/actions/setup-haskell/issues/44), we'll have to wait to the fix

Many thanks for digging this out!

PR #914, opened 2020-11-17
Make Filetargets absolute before continue using them

fendor:
The filetargets are not correctly absolute. The hie-bios API did not articulate it correctly. 

So, in certain cases, the following error is shown for file-targets in sub-directory:
```
/tmp/extra-dir-30793437515428/src/Main.hs: openBinaryFile: does not exist (No such file or directory)
```

Adds a test that this was the case and is now fixed.


pepeiborra:
Thanks! We didn't see any bug reports about this because the cabal and stack cradles produce module targets, right?

pepeiborra (approved):

fendor:
For executables you get file targets. I got a bug report on IRC, but the module actually continues working, but there is a problem in problems tab. You can actually provoke it in the test project I added.

PR #915, opened 2020-11-18
Switch to sliding tags in Github actions

pepeiborra:


ndmitchell (approved):
Seems much better than continually updating

jneira (approved):

PR #916, opened 2020-11-20
Support extending constructors

berberman:
Perhaps we no longer need to parse the extended binding, completely depending on the diagnostic message, since we now have `ExportsMap`. Closes haskell/haskell-language-server#609:

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -19,53 +19,48 @@ module Development.IDE.Plugin.CodeAction
     -- * For testing
     , blockCommandId
     , typeSignatureCommandId
+    , matchRegExMultipleImports
     ) where
 
-import           Bag                                               (isEmptyBag)
-import           Control.Applicative                               ((<|>))
-import           Control.Arrow                                     ((>>>))
-import           Control.Concurrent.Extra                          (readVar,
-                                                                    threadDelay)
-import           Control.Monad                                     (guard, join)
-import           Data.Aeson.Types                                  (Result (..),
-                                                                    Value (..),
-                                                                    fromJSON,
-                                                                    toJSON)
-import           Data.Char
-import           Data.Function
-import           Data.Functor
-import qualified Data.HashMap.Strict                               as Map
-import qualified Data.HashSet                                      as Set
-import           Data.List.Extra
-import           Data.Maybe
-import qualified Data.Rope.UTF16                                   as Rope
-import qualified Data.Text                                         as T
-import           Data.Tuple.Extra                                  ((&&&))
-import           Development.IDE.Core.RuleTypes
-import           Development.IDE.Core.Rules
-import           Development.IDE.Core.Service
-import           Development.IDE.Core.Shake
-import           Development.IDE.GHC.Compat
-import           Development.IDE.GHC.Error
-import           Development.IDE.LSP.Server
-import           Development.IDE.Plugin
-import           Development.IDE.Plugin.CodeAction.PositionIndexed
-import           Development.IDE.Plugin.CodeAction.RuleTypes
-import           Development.IDE.Plugin.CodeAction.Rules
-import           Development.IDE.Types.Exports
-import           Development.IDE.Types.Location
-import           Development.IDE.Types.Options
-import           Development.Shake                                 (Rules)
-import           GHC.LanguageExtensions.Type                       (Extension)
-import qualified Language.Haskell.LSP.Core                         as LSP
-import           Language.Haskell.LSP.Messages
-import           Language.Haskell.LSP.Types
-import           Language.Haskell.LSP.VFS
-import           Outputable                                        (ppr,
-                                                                    showSDocUnsafe)
-import           Safe                                              (atMay)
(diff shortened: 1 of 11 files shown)


pepeiborra:
Thank you for this change, this was so annoying!
Do we need a better test?

Could you undo all the auto formatting? It will lead to unnecessary merge conflicts for other contributions, and its best done by maintainers when the time is right. Thank you!

berberman:
Sorry for that, I've restored changes done by the formatter. 
As for tests, hmmm, I don't have a good idea. Old tests can pass after the changes. Should I add the case in the screenshot?

pepeiborra:
Yes, I think encoding the screenshot as a test would be good enough. Thanks!

berberman:
I have a question about this test case,
```
,  (`xfail` "known broken") $ testSession "extend single line import with constructor" $ template
      (T.unlines
            [ "module ModuleA where"
            , "data A = Constructor"
            ])
      (T.unlines
            [ "module ModuleB where"
            , "import ModuleA (A)"
            , "b :: A"
            , "b = Constructor"
            ])
      (Range (Position 2 5) (Position 2 5))
      "Add Constructor to the import list of ModuleA"
      (T.unlines
            [ "module ModuleB where"
            , "import ModuleA (A(Constructor))"
            , "b :: A"
            , "b = Constructor"
            ])
```
which is marked as "known broken". Expected failure here means "user should not write code like that" or "currently we don't support this feature"?

pepeiborra:
That test case is marked as known broken because it reproduces a known bug that needs to be fixed. Which your PR does

pepeiborra (approved):

berberman:
Now we can extend `()` to `(A(Constructor))`, but not `(A)` to `(A(Constructor))`. So this test is failing, breaking the CI, since the reult becomes `(A(Constructor),A)`. The later one could be implemented, however, I don't think the user would add the constructor to the import list without the parent like this...

berberman:
The test which was marked as  "known broken" previously can pass now, and I added a new test to demonstrate extending constructors. However, current implementation breaks the "extend multi line import with value", because spaces are dropped before removing the trailing comma. This can be fixed by removing the comma in the end of the import list but keeping spaces.

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1108,17 +1109,31 @@ rangesForBinding' _ _ = []
...
+        binding' = (if T.null parent then id else T.init) renderedBinding
+     in removeTrailingComma $ T.concat [pre, "(", binding', rest']
+  _ ->
+    error $
+      "importLine does not have the expected structure: "
         <> T.unpack importLine
-
+  where
+    removeTrailingComma s = case T.breakOnEnd ")" s of
+      (T.unsnoc -> Just (h, _), t) -> T.dropWhileEnd (\x -> x == ',' || isSpace x) h <> ")" <> t
```
berberman, line 1132: Don't drop spaces

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1108,17 +1109,31 @@ rangesForBinding' _ _ = []
...
+    let rest' = case parent of
+          "" -> ", " <> rest
+          _ -> case T.breakOn parent rest of
+            (h, T.stripPrefix parent -> Just r) -> case T.uncons (T.dropWhile isSpace r) of
+              Just (')', _) -> ")" <> h <> r
+              Just ('(', xs) -> let imported = T.takeWhile (/= ')') xs in T.concat ["," ,imported , "), " , h , removeHeadingComma (T.tail (T.dropWhile (/= ')') r))]
+              _ -> "), " <> h <> r
+            _ -> "), " <> rest
+        binding' = (if T.null parent then id else T.init) renderedBinding
+     in removeTrailingComma $ T.concat [pre, "(", binding', rest']
```
pepeiborra, line 1125: I find this code very hard to read. Could you try to golf it a bit to make it more readable? Some concrete suggestions:

- whitespace
- both `rest'` and `binding'` perform branching on whether `parent` is null. Make this more apparent
- introduce more names or add more comments, there's a lot of character matching that is hard to follow
- consider replacing some of the character matching with regexes if that makes things simpler.

pepeiborra (changes requested):

src/Development/IDE/Plugin/CodeAction.hs
berberman: Thanks for your suggestions! I totally agree. It lacks of extensibility and is hard to read 😢

pepeiborra: @berberman when are you planning to implement this suggestion?

adamse:
Anything I can do to help move this forward? :)

berberman:
Sorry for the delay; I‘ve been pretty busy recently...
I may be free next weekend, and helps are welcome :)
Current implementation works but is ugly, so my plan is something like handling two branches (whether the thing we are going to extend is a constructor of a data type) separately, and using regexs or parsers to improve the readability and robustness.

pepeiborra:
I'll merge this as it is, please send a new PR for clean-ups.

pepeiborra (approved):

PR #918, opened 2020-11-22
Add a GitHub action for the Nix build

pepeiborra:
/cc @domenkozar

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -15,7 +15,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: cachix/install-nix-action@v12
       with:
-        nix_path: nixpkgs=channel:nixos-unstable
+        nix_path: nixpkgs=channel:nixos-20.03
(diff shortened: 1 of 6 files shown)


.github/workflows/nix.yml
```diff
@@ -0,0 +1,23 @@
...
+      fail-fast: false
+      matrix:
+        ghc: ['default', 'ghc8102', 'ghc884', 'ghc865']
+        os: [ubuntu-latest, macOS-latest]
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: cachix/install-nix-action@v12
+      with:
+        nix_path: nixpkgs=channel:nixos-unstable
```
domenkozar, line 18: I'd recommend [pinning a commit](https://nix.dev/reference/pinning-nixpkgs.html#possible-url-values) here, otherwise you'll possibly see non-deterministic breakages. Rarely, but not worth the hassle.

jneira (approved):

PR #919, opened 2020-11-22
Ignore import list while producing completions

### LINKED ISSUE #576: Ignore import lists for producing completions

gdevanla:
Fixes #576

### DIFF
--- a/src/Development/IDE/Plugin/Completions.hs
+++ b/src/Development/IDE/Plugin/Completions.hs
@@ -71,14 +71,26 @@ produceCompletions = do
         case (ms, sess) of
             (Just (ms,imps), Just sess) -> do
               let env = hscEnv sess
-              res <- liftIO $ tcRnImportDecls env imps
+              -- We do this to be able to provide completions of items that are not restricted to the explicit list
+              let imps' = map dropListFromImportDecl imps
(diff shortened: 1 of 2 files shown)


pepeiborra (approved):
Nice, thank you! 

Did you look into extending the import declaration as part of the completion? I guess we can leave it for optional future work.

gdevanla:
That's a good idea. Right now we need the user to choose an explicit code action to do that. Should this be done automatically?

pepeiborra:
Automatically extending the import is part of #576: 

> It would be great if selecting a completion would also check and expand the corresponding import list if needed.

pepeiborra:
What do you plan to do about automatically expanding imports? I'm happy to merge this now and leave that as future work, but then we should not close #576

gdevanla:
I would suggest merging this and leave that to next PR.  I have a couple of questions. I will ask those in #576.  I think it is fine to leave that issue open for now. Thanks!

PR #924, opened 2020-11-28
Clean up some of the linky-links in the README

samuela:
Add a link to the vscode extension, and minor cleanup.


jneira (approved):
many thanks for improving docs!

PR #926, opened 2020-11-28
Favor `lookupPathToId` over `pathToId`

### LINKED ISSUE #921: pathToId is partial
When editing Shake I was changing an import from `import Data.Either` to `import Data.Either.Extra`. It froze at `import Data.Either.Extr` with the message:

```
Data.HashMap.Internal.(!): key not found
CallStack (from HasCallStack):
  error, called at .\\Data\\HashMap\\Internal.hs:756:16 in unordered-con_-0.2.13.0-6786623f456cdab63dfcd11cc799ca8c6a341945:Data.HashMap.Internal
  !, called at src\Development\IDE\Import\DependencyInformation.hs:108:40 in ghcide-0.5.0-inplace:Development.IDE.Import.DependencyInformation
```

That seems to correspond to `pathToId`. Looking at the use cases of `pathToId`, I see:

* `reportImportCyclesRule`, but that seems OK since we've just asked for the dependency information for the file we're passing, so I assume the invariant is OK.
* `transitiveReverseDependencies` called by `typecheckParentsAction` which seems to rely on the current module graph having a given file in it. I don't see why that's guaranteed, and I think having `transitiveReverseDependencies` return `Nothing` if file resolution fail makes more sense.
* `immediateReverseDependencies` called by `needsCompilationRule` - same issue as the transitive case, and I think we'd be better served by a Nothing.
* `transitiveDeps`, which uses a dependency information based on a file we have at hand.

I'd suggest that `transitiveReverseDependencies` and `immediateReverseDependencies` be rewritten on top of `lookupPathToId` with `Nothing` returns. I don't know which broke for me, but I suspect its one of them.

Given that this occurred during editing, there are likely race conditions between completions or something and checking, so I don't see reproducing as a particularly fruitful endeavour, and I suspect writing a test is hard without the ability to reproduce.

samuela:
Fixes #921.

Done: `transitiveReverseDependencies` and `immediateReverseDependencies` have been refactored to return `Maybe [...]`.

I haven't changed `reportImportCyclesRule` or `transitiveDeps` as they seem to be acquitted in https://github.com/haskell/ghcide/issues/921#issue-749520646, although I can't say that I personally understand the logic there.

TODO: Predictably these types changing has stirred up type errors elsewhere. Although there are obvious workarounds, I'm not familiar with the desired behavior for these call sites to propose just converting `Nothing` -> `[]`.

https://github.com/haskell/ghcide/blob/30a46e8a18263c7b76384b854f40b21e6e88f7ba/src/Development/IDE/Core/FileStore.hs#L245

I imagine this should error out if we fail to get the reverse dependencies? If so, what's the appropriate way to error out of the `Action` monad?

https://github.com/haskell/ghcide/blob/f4bfe9c103e34de1d61d28f459eddf9e87f5c712/src/Development/IDE/Core/Rules.hs#L913

I imagine that `needsCompilationRule` should never fail? I don't understand what's going on here.

### DIFF
--- a/src/Development/IDE/Import/DependencyInformation.hs
+++ b/src/Development/IDE/Import/DependencyInformation.hs
@@ -317,23 +317,23 @@ partitionSCC (AcyclicSCC x:rest) = first (x:)   $ partitionSCC rest
 partitionSCC []                  = ([], [])
 
 -- | Transitive reverse dependencies of a file
-transitiveReverseDependencies :: NormalizedFilePath -> DependencyInformation -> [NormalizedFilePath]
-transitiveReverseDependencies file DependencyInformation{..} =
-  let FilePathId cur_id = pathToId depPathIdMap file
-  in map (idToPath depPathIdMap . FilePathId) (IntSet.toList (go cur_id IntSet.empty))
+transitiveReverseDependencies :: NormalizedFilePath -> DependencyInformation -> Maybe [NormalizedFilePath]
+transitiveReverseDependencies file DependencyInformation{..} = do
+    FilePathId cur_id <- lookupPathToId depPathIdMap file
+    return $ map (idToPath depPathIdMap . FilePathId) (IntSet.toList (go cur_id IntSet.empty))
   where
     go :: Int -> IntSet -> IntSet
     go k i =
-      let outwards = fromMaybe IntSet.empty (IntMap.lookup k depReverseModuleDeps  )
+      let outwards = fromMaybe IntSet.empty (IntMap.lookup k depReverseModuleDeps)
           res = IntSet.union i outwards
           new = IntSet.difference i outwards
       in IntSet.foldr go res new
 
 -- | Immediate reverse dependencies of a file
-immediateReverseDependencies :: NormalizedFilePath -> DependencyInformation -> [NormalizedFilePath]
-immediateReverseDependencies file DependencyInformation{..} =
-  let FilePathId cur_id = pathToId depPathIdMap file
-  in map (idToPath depPathIdMap . FilePathId) (maybe mempty IntSet.toList (IntMap.lookup cur_id depReverseModuleDeps))
+immediateReverseDependencies :: NormalizedFilePath -> DependencyInformation -> Maybe [NormalizedFilePath]
+immediateReverseDependencies file DependencyInformation{..} = do
+  FilePathId cur_id <- lookupPathToId depPathIdMap file
+  return $ map (idToPath depPathIdMap . FilePathId) (maybe mempty IntSet.toList (IntMap.lookup cur_id depReverseModuleDeps))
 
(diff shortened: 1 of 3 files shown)


pepeiborra:
The call site in `FileStore.hs` is being asked to perform typechecking of the "parents" of the module in the given path. A return value of `Nothing` in `transitiveReverseDependencies` would mean that this file is not in the module graph, which can happen if a `ModSummary` could not be parsed out of it. It would be ok to do nothing in this case.

The call site in `Rules.hs` should never return `Nothing` because only a few lines above there is a successful call to `GetModSummaryWithoutTimestamps`, and it would be ok to handle a `Nothing` with an `error`.

src/Development/IDE/Import/DependencyInformation.hs
```diff
@@ -317,23 +317,23 @@ partitionSCC (AcyclicSCC x:rest) = first (x:)   $ partitionSCC rest
...
 
 -- | Immediate reverse dependencies of a file
-immediateReverseDependencies :: NormalizedFilePath -> DependencyInformation -> [NormalizedFilePath]
-immediateReverseDependencies file DependencyInformation{..} =
-  let FilePathId cur_id = pathToId depPathIdMap file
-  in map (idToPath depPathIdMap . FilePathId) (maybe mempty IntSet.toList (IntMap.lookup cur_id depReverseModuleDeps))
+immediateReverseDependencies :: NormalizedFilePath -> DependencyInformation -> Maybe [NormalizedFilePath]
+immediateReverseDependencies file DependencyInformation{..} = do
+  FilePathId cur_id <- lookupPathToId depPathIdMap file
+  return $ map (idToPath depPathIdMap . FilePathId) (maybe mempty IntSet.toList (IntMap.lookup cur_id depReverseModuleDeps))
```
wz1000, line 336: I think `idToPath` is also partial

samuela: Ah you're correct... it is by virtue of `idTomodLocation` being partial: https://github.com/haskell/ghcide/blob/51f70d410aba4eebc5ad371e929fb5c5595029d9/src/Development/IDE/Import/DependencyInformation.hs#L113-L117

I'd prefer to defer fixing that half of things to a separate PR if possible.

OOC, why so many partial functions flying around here in the first place?

pepeiborra: Back before `GetModuleGraph` and `checkParents` existed, these partial functions were assumed to be safe in every callsite.

samuela:
Ok, @pepeiborra thanks for your feedback! I believe I've implemented your suggestions in https://github.com/haskell/ghcide/pull/926/commits/b2675d31534a45ff7d911a17c6bf036717146931 and https://github.com/haskell/ghcide/pull/926/commits/d95dd1fcc0d88aabc58621f9b36ac27b42d6bf4c. I'll mark this as ready for review now.

pepeiborra (approved):
Very nice, thanks for improving the code readability.

PR #928, opened 2020-11-29
Custom cradle loading

pepeiborra:
When using ghcide as a library, it may be desirable to host the hie.yaml file
in a location other than the project root, or even avoid the file system altogether


jneira (approved):
Nice point of configurable behaviour

PR #929, opened 2020-11-29
Return completion snippets only when client supports it

pepeiborra:
Fixing a wart that came up during #900


ndmitchell (approved):
Looks good

PR #931, opened 2020-12-01
Update implicit-hie

Avi-D-coder:


cabal.project
```diff
@@ -17,5 +17,5 @@ allow-newer:
 
 -- To ensure the build get the version with the fix for
 -- https://github.com/Avi-D-coder/implicit-hie/issues/17
-constraints: implicit-hie >= 0.1.2.3
+constraints: implicit-hie >= 0.1.2.5
 constraints: implicit-hie-cradle >= 0.3.0.0
```
jneira, line 21: If we had a lower bound in impicit-hie-cradle over implicit-hie, with a hackage revision for example, we could get ride of those constraints (here and in hls too)

jneira: @Avi-D-coder what do you think about make the hackage revision? we can wait to the next release though

Avi-D-coder: I'll take a look tomorrow

jneira:
ummm, closed deliberately?

PR #934, opened 2020-12-03
Drop stack Windows CI

pepeiborra:
The Stack Windows build is problematic:
  https://github.com/haskell/ghcide/pull/922

Stack is already covered by the Azure CI
Windows is already covered by the Github Actions CI


.azure/windows-stack.yml
```diff
@@ -1,41 +0,0 @@
-jobs:
```
jneira, line 1: Agree in drop it temporary, but i would keep the config file and only remove the build from azure-pipelines.yaml

stack-windows.yaml
```diff
@@ -1,32 +0,0 @@
-resolver: nightly-2020-06-19
```
jneira, line 1: Same with this file, i would like to make it work.

.azure/windows-stack.yml
pepeiborra: You can revert this commit to bring it back

stack-windows.yaml
pepeiborra: Use git revert

jneira (approved):

pepeiborra:
Damn it, the rebase Brough `stack-windows.yaml` back. I will need to send another PR to delete it

jneira:
die hard 😝

Issue #726, opened 2020-08-11
Hole suggestions doesn't properly qualify the result


anka-213:
If I import a module qualified
```
import qualified Data.List as L
```
and then ask vscode to fill a hole with something from there
```
myTails :: [a] -> [[a]]
myTails = _
```
it will pretend that it was available unqualified
```
myTails :: [a] -> [[a]]
myTails = tails
```
which Haskell will complain about again, but fortunately ghcide gives a suggestion to automatically fix it, so after another `⌘-. <Enter>` I will end up with what I wanted.
```
myTails :: [a] -> [[a]]
myTails = L.tails
```
The last step shouldn't be needed though.

jneira:
@tritlo reported [this](https://github.com/haskell/ghcide/issues/932), including a live demo: https://youtu.be/lfrk0wyZHvg?t=2311

Tritlo:
> @Tritlo reported [this](https://github.com/haskell/ghcide/issues/932), including a live demo: https://youtu.be/lfrk0wyZHvg?t=2311

Yeah, except I didn't even get the suggestion to update it!

anka-213:
> Yeah, except I didn't even get the suggestion to update it!
> 
Right, because for you there was already another value in scope with that name, so it becomes a type error instead of an “undefined variable” error. Type errors doesn’t (currently) suggest similarly named functions with the correct type.

pepeiborra:
Qualifiers in GHC error messages are getting munged before code actions have a chance to see them. 
I traced this back to #546. Reverting that commit will fix this issue, #652 and probably others. However it will also make error messages materially worse.

pepeiborra:
I'm working on a fix for this

PR #938, opened 2020-12-06
Qualified error messages

### LINKED ISSUE #652: "insert type signature" does not respect qualified imports
For example, inserting a type signature for the expression `test` below causes an error because `test :: Ratio Integer` is inserted instead of `test :: Real.Ratio Integer`.

```haskell
import GHC.Real ((%))
import qualified GHC.Real as Real

test = 4 % 5
```

I'm using ghcide version: 0.2.0 with GHC: 8.6.5.

pepeiborra:
Fixes #652 and #726


jneira (approved):
Small pr, great fix 👍

berberman:
Also closes haskell/haskell-language-server#441 :heart:

Issue #936, opened 2020-12-06
Consider extracting completions and code actions into HLS plugins


pepeiborra:
The completions code in ghcide is structured as a single provider that serves multiple contexts (pragmas, imports, types, values, etc.). The static set of dependencies is particularly suboptimal: no completions for pragmas unless the file type checks (#631)! It would be advisable to break this down into a family of providers that can all contribute completions depending on the context. 

Haskell-language-server has a plugin model that solves exactly this problem. Rather than duplicating this model in ghcide, it would make sense to extract HLS plugins instead. It would constitute a loss of functionality for ghcide users, but note that we are considering to remove the ghcide executable (#935). If that doesn't go ahead, we could always extend the current executable with the ability to run HLS plugins.

Similarly, the code actions code in ghcide should be properly extracted out as HLS plugins, although this seems lower priority. The code is already loosely structured as a bunch of providers, 16 in total, and does not have the shortcomings listed above for completions. An HLS plugin translation would bring certain benefits -decoupling, parallelism and isolation- and hopefully no downsides.

gdevanla:
@pepeiborra I feel this is a good project I can focus my efforts on. If no one else has taken a stab at it, I can start work on this. I like the idea very much. I did feel the `Completions` code was doing to many things all at once.

Would this necessarily mean, the `Completions` and `CodeActions` section will be completely moved out of `ghcide`, except for say the implementation of the `Rule` types?

Based on my understanding, this is a multi-step process, which needs co-ordination for removal from ghcide and addition to respective hls plugins.  Do you have any initial thoughts on how we should be approaching this on a per task basis. Perhaps, we could have a list of TODO items under this task, to accomplish what we need.  I look forward to other suggestions as well. Meanwhile, I can prepare my own notes on the changes needed.

pepeiborra:
By all means, do work on this if you are interested!

My mental model is: 

- I would start with the completions. If it goes well continue with the code actions. There's no need to micromanage if you are the only person working on it

- Extract all the code to one or more packages which depend on `his-plugin-api` and include one or more plugins each. This package would live in the haskell-language-server repo. Something like:

  - `hls-ghc-completions-plugins`: A cabal package containing three (or more) HLS completion providers
    - "local-completions" provider: An HLS provider for local completions and the `LocalCompletions` rule
    - "non-local-completions" provider: An HLS provider for non-local completions and the `NonLocalCompletions` rule
    - "import-completions" provider: similar

[...]

PR #939, opened 2020-12-06
Deprecate ghcide tool and delete the VSCode extension

### LINKED ISSUE #935: Deprecate the ghcide executable
We have already told people to not use ghcide directly: 

https://neilmitchell.blogspot.com/2020/09/dont-use-ghcide-anymore-directly.html

The next logical step is to deprecate and eventually remove the ghcide executable.

/cc @cocreature @ndmitchell any concerns?

pepeiborra:
Closes #935


jneira (approved):

larskuhtz:
I am an happy user of the ghcide tool and I'd like to ask to not deprecate it.

I know that I am supposed to use the Haskell-language-server. Unfortunately, that's not an option for me. Just like hie in the past, it comes with too many features and dependencies that cause too much overhead to build and make them work in my development environment -- each time I tried it took me a long time to install it and get it to work and it would usually break soon due to dev-environment changes (like a new cabal or ghc version or other often non-obvious reasons). 

What I like about ghcide is that it provides a non-opinionated set of basic IDE features. Those features cover everything that I need most (type checking, completion, jump to definition, show type) in a single easy to build, efficient, and robust tool. Haskell-language-server is adding features on top of it that I consider less important. Moreover many of the features the additional features of haskell-language-server seem opinionated to me. And they come at the price of a much more difficult to install and to use product.

larskuhtz:
To reiterate, in my experience, ghcide is the first and only Haskell IDE tool (at least for vim) that just builds (even relatively quickly with a simple `cabal install`) and just works out of the box.

ndmitchell:
@larskuhtz thanks for the feedback. It seems looking at your list of issues, the main two are:

* HLS is less GHC/Cabal/whatever changes compatible. Is this still true? It certainly was with HIE, but I was hoping HLS was less the case. Are there any specific dependencies that have caused you problems? Most of the GHC/Cabal compatibilities should be coming from hie-bios (which is shared) or GHC bumps (which are tested to the same versions with HLS and Ghcide).
* HLS takes longer to build. Do we have numbers on just how long? Would be useful information to have.
* Are there any other reasons you don't want to use HLS, but prefer Ghcide?

It may be that with some flags you can trim down HLS to very little over and above Ghcide (maybe you already can and we just haven't documented), or it may be that the work we've done means its no longer as problematic as it used to be.

larskuhtz:
@ndmitchell I haven't used HLS yet. I tried to compile it using `cabal v2-run ./install.hs --project-file install/shake.project hls`. It failed with an TH related error for which it was not obvious to me how to fix it. ghcide builds with a simple `cabal build` on a plain ghc/cabal setup, which is a big plus for me. (To be fair, HLS actually failed while compiling ghcide, however when I build ghcide directly, I can just repeat the build and it passes on the second try. In the context of an HLS build that doesn't work). 

HLS has dependencies that have been causing build issues in the past or would only build with patches, like for instance brittany. Overall the project structure of HLS is more complex and my experience tells me that projects of that kind tent to be fragile -- e.g. can take a longer time to be ported to new ghc versions. Ideally, an IDE tool should be available broadly across many different dev-environments, because I can't afford to switch or reconfigure my development tools for each project that I work on.

I understand that more advanced IDE tools provide a lot of value to teams working in relatively homogeneous and stable environments where it pays off to spent the extra time to setup and maintain those tools. But I also think that its desirable to have tools that are focused on common basic needs and just work everywhere with little dependencies. To me it seems that HLS belongs to the former while Ghcide belongs to the latter category of tools. 

So, I don't want to argue against HLS -- I will continue to try it and gladly use it if it works for me. I just like to continue to use ghcide, at least as a fallback solution that is available when HLS isn't.

jneira:
@larskuhtz i agree with almost all arguments.
I think @pepeiborra plan is to slim down hls to get a version with lot less dependencies, see https://github.com/haskell/haskell-language-server/pull/647#issuecomment-739138203
That said, i think we will keep ghcide executable at mid (long?) term to use it in ghcide tests.

It would be great to open an issue in hls about your failing install but i want to note there are precompiled executables for all major os's and supported ghc versions that make you not need to even compile it: https://github.com/haskell/haskell-language-server/releases/tag/0.6.0

pepeiborra:
Exactly. As a user, you want to have a small ide with few dependencies. I want the same thing! 

But as discussed in #936 we want to break down some of the monolithic completions code into multiple providers using the nice plugin model in HLS. Thus it seems natural to move this out of ghcide and into an HLS plugin.

These two desires should not be in opposition. 

1. We need to make improvements to the HLS project definition and build system so that it's easy to select and build only a subset of the plugins. 
2. And we need to get the current ghcide users onto HLS so that we can move the completions and other code without impacting them. 

This PR is to get 2 in motion, and before merging I will add a link to this discussion for anyone interested in why we are doing this.

larskuhtz:
> I think @pepeiborra plan is to slim down hls to get a version with lot less dependencies

That would be great, too, if HLS would be modular in a way that a build would just ignore dependencies that one doesn't care about or that might be less stable. What I really like about ghcide is that it "just builds" and "just works" with a vanilla ghc dev environment.

In any case my comments are meant as a compliment for ghcide and not criticism of hls.

PR #940, opened 2020-12-06
Prepare for v0.6.0 release

pepeiborra:
As usual the changelog is an edited version of the git log with non user-visible changes trimmed out, making sure that
everyone who contributed is listed at least once.


wz1000:
If we can commit to a monthly release schedule for ghcide, I think it makes sense to sync ghcide and hls release numbers.

/cc @alanz

alanz:
> If we can commit to a monthly release schedule for ghcide, I think it makes sense to sync ghcide and hls release numbers.

I don't have strong feelings on this, the monthly releases are the important thing.

jneira (approved):
Thanks for the release, we are gonna use it for hls next one

PR #941, opened 2020-12-06
Extract the benchmarking Shake rules to a standalone Cabal package

pepeiborra:
There's some room for reusing the rules used in the historic benchmarking suite
in other projects. This change makes that a bit easier and improves the
documentation and code structure.

The new structure is:
- `lib:shake-bench` - a Cabal library with functions to generate Shake rules
- `ghcide:bench:benchHist` - the ghcide instantiation of the above Shake rules

That's not to say that `shake-bench` is completely decoupled from ghcide -
there are still plenty of assumptions on how the benchmarks are organized, their
outputs, etc. But with a little bit more effort it should be easy to make
these rules more reusable

### DIFF
--- a/shake-bench/LICENSE
+++ b/shake-bench/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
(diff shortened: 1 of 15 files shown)


wz1000:
Very nice. Thanks!

shake-bench/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2019 Digital Asset (Switzerland) GmbH and/or its affiliates
```
wz1000, line 189: Might want to change this

pepeiborra: well spotted, thanks!

wz1000 (approved):

PR #942, opened 2020-12-07
Cleanup addBindingToImportList

berberman:
#916 was tricky and didn't work in some cases. Now the procedure has become clearer, and I add some comments to show how it works. I'm not sure all cases are covered, and we may need more tests.

### DIFF
--- a/src/Development/IDE/Plugin/CodeAction.hs
+++ b/src/Development/IDE/Plugin/CodeAction.hs
@@ -657,10 +657,9 @@ suggestExtendImport exportsMap contents Diagnostic{_range=_range,..}
                    in x{_end = (_end x){_character = succ (_character (_end x))}}
                 _ -> error "bug in srcspan parser",
             importLine <- textInRange range c,
-            Just ident <- lookupExportMap binding mod
-            =
-                [("Add " <> renderImport ident <> " to the import list of " <> mod
-                , [TextEdit range (addBindingToImportList ident importLine)])]
+            Just ident <- lookupExportMap binding mod,
+            Just result <- addBindingToImportList ident importLine
+            = [("Add " <> renderImport ident <> " to the import list of " <> mod, [TextEdit range result])]
           | otherwise = []
         renderImport IdentInfo {parent, rendered}
           | Just p <- parent = p <> "(" <> rendered <> ")"
@@ -1110,13 +1109,13 @@ rangesForBinding' _ _ = []
 --       import (qualified) A (..) ..
 --   Places the new binding first, preserving whitespace.
 --   Copes with multi-line import lists
-addBindingToImportList :: IdentInfo -> T.Text -> T.Text
+addBindingToImportList :: IdentInfo -> T.Text -> Maybe T.Text
 addBindingToImportList IdentInfo {parent = _parent, ..} importLine =
   case T.breakOn "(" importLine of
     (pre, T.uncons -> Just (_, rest)) ->
       case _parent of
         -- the binding is not a constructor, add it to the head of import list
-        Nothing -> T.concat [pre, "(", rendered, addCommaIfNeeds rest]
+        Nothing -> Just $ T.concat [pre, "(", rendered, addCommaIfNeeds rest]
         Just parent -> case T.breakOn parent rest of
           -- the binding is a constructor, and current import list contains its parent
           -- `rest'` could be 1. `,...)`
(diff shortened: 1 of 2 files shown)


src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -657,17 +657,17 @@ suggestExtendImport exportsMap contents Diagnostic{_range=_range,..}
...
         renderImport IdentInfo {parent, rendered}
-          | Just p <- parent = (p, p <> "(" <> rendered <> ")")
-          | otherwise        = ("", rendered)
+          | Just p <- parent = p <> "(" <> rendered <> ")"
+          | otherwise        = rendered
         lookupExportMap binding mod 
-          | [(renderImport -> pair, _)] <- filter (\(_,m) -> mod == m) $ maybe [] Set.toList $ Map.lookup binding (getExportsMap exportsMap)
-           = Just pair
+          | [(ident, _)] <- filter (\(_,m) -> mod == m) $ maybe [] Set.toList $ Map.lookup binding (getExportsMap exportsMap)
+           = Just ident
```
pepeiborra, line 670:
```suggestion
          | Just match <- Map.lookup binding (getExportsMap exportsMap)
          , [(ident, _)] <- filter (\(_,m) -> mod == m) (Set.toList match)
           = Just ident
```

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -1109,31 +1109,40 @@ rangesForBinding' _ _ = []
...
+          --               or 2. `(),...)`
+          --               or 3. `(ConsA),...)`
+          --               or 4. `)`
+          (leading, T.stripPrefix parent -> Just rest') -> case T.uncons (T.stripStart rest') of
+            -- case 1: no children and parentheses, e.g. `import A(Foo,...)` --> `import A(Foo(Cons), ...)`
+            Just (',', rest'') -> T.concat [pre, "(", leading, parent, "(", rendered, ")", addCommaIfNeeds rest'']
+            -- case 2: no children but parentheses, e.g. `import A(Foo(),...)` --> `import A(Foo(Cons), ...)`
+            Just ('(', T.uncons -> Just (')', rest'')) -> T.concat [pre, "(", leading, parent, "(", rendered, ")", rest'']
+            -- case 3: children with parentheses, e.g. `import A(Foo(ConsA),...)` --> `import A(Foo(Cons, ConsA), ...)`
+            Just ('(', T.breakOn ")" -> (children, rest''))
```
pepeiborra, line 1131: `T.stripStart` doesn't strip comments, so

    import M(Foo({-Cons-}))` 

will fall through case 3 and unexpectedly become 

    import M(Foo(ConsB, {-Cons-})

Not your fault though, your code is as good as it can reasonably be. But the only fully correct way to make edits in the user AST is with ghc-exactprint

pepeiborra (commented):
Very nice! I really appreciate the new code, its' much easier to understand and to maintain. And more correct too!

pepeiborra (approved):

src/Development/IDE/Plugin/CodeAction.hs
berberman: Things become quite complicated involving comments:

* block comments are considered to be constructors
```haskell
import M(Foo({-Cons-}))
```
⇓
```haskell
import M(Foo(ConsB, {-Cons-})
```

* unexpected `-}` after we matched the parent
```haskell
import M({-Foo-}Foo)
```
⇓
```haskell
error "unknown case"
```

* line comments 
```haskell
  import M(
--Foo
  Foo)
```
⇓
```haskell
error "unknown case"
```

* ...

It's knotty to resolve these cases simply relying on `breakOn` logic. As you suggested, manipulating ASTs would be an appropriate way. I'll check `ghc-exactprint` out :)

pepeiborra:
The tests are unreliable because the code action needs the package exports map to be populated, which is done implicitly by `kick`. To wait until `kick` is done, I think it should be enough to change the test template to wait for progress after waiting for diagnostics:

```
      void (skipManyTill anyMessage message :: Session WorkDoneProgressEndNotification)
```

berberman:
Thanks! Should we adopt this implementation? If so, how to handle unknown cases? I don't think `error` is suitable here, since it will crash ghcide.

pepeiborra:
I would handle the unknown cases by returning a `Nothing` so that no code action is provided. After all, we don't know how to apply GHC's suggestion.
In a follow-up change, rework this using ghc-exactprint. For this, I suggest asking @isovector for permission to inline the very useful https://github.com/haskell/haskell-language-server/blob/master/plugins/tactics/src/Ide/TreeTransform.hs in ghcide

isovector:
@pepeiborra please do. It's only where it is for convenience to me, but always was intended to be upstreamed.

PR #944, opened 2020-12-08
Add support for customizing the hidir location

pepeiborra:
Not just the hidir but also the hiedir and the odir


jneira (approved):
Good things to be configured

PR #945, opened 2020-12-10
Use qualified module name from diagnostics in suggestNewImport

berberman:
GHC helps us to split the qualified module name from the identifier. Closes haskell/haskell-language-server#629


pepeiborra:
Haven't looked at the code yet, but haskell/haskell-language-server#629 is about operators and the title of this PR "Use qualified module name in suggestNewImport" doesn't seem related.

berberman:
Ah, the issue was caused by we mistakenly stripped the qualified import prefix of identifiers. For example, if I write `T.null` , where the `T` is not in scope, diagnostic will report:

> Not in scope: ‘T.null’
> No module named ‘T’ is imported.

The previous logic only took the first line, then tried to use `splitOn "."` to strip `T.`, obtaining the identifier `null`, which we use to lookup the export map. Everything goes well.

But operators like `.|.` were split unexpectedly, result in `|.` as the identifer. In this PR, I try to extract the qualified module name from the second line of the diagnostic message, so we no longer to use `splitOn` and `intercalate` manually, removing the qualified module name by `T.stripPrefix` instead.

berberman:
Looks like the CI broke for some unknown reasons, but all checks passed on my fork: https://github.com/berberman/ghcide/tree/patch-qual
Ready to be reviewed and merged :-)

pepeiborra (approved):
Thank you, this is an awesome bug fix.  

I don't know what the problem is with CI, could you perhaps rebase on top of master?

src/Development/IDE/Plugin/CodeAction.hs
```diff
@@ -624,7 +624,10 @@ suggestExtendImport exportsMap contents Diagnostic{_range=_range,..}
             Just result <- addBindingToImportList ident importLine
             = [("Add " <> renderIdentInfo ident <> " to the import list of " <> mod, [TextEdit range result])]
           | otherwise = []
-        lookupExportMap binding mod 
+        renderImport IdentInfo {parent, rendered}
```
berberman, line 627: haha, it came back

pepeiborra: That was my fault, trying to fix a merge conflict

PR #946, opened 2020-12-10
Extend CI matrix with all the GHC minor versions supported by HLS

pepeiborra:

PR #947, opened 2020-12-10
Use GADTs for all ghc versions in Development.IDE.Plugin.Completions.Logic

jneira:
* It fixes the build for ghc-8.8.3 and ghc-8.8.2 that was failing with:

```
Progress 166/173: ghcide ghcide > /root/build/ghcide/src/Development/IDE/Plugin/Completions/Logic.hs:267:5: error:
Progress 166/173: ghcide ghcide >     • Illegal equational constraint NoGhcTcPass pass
Progress 166/173: ghcide ghcide >                                     ~ NoGhcTcPass (NoGhcTcPass pass)
Progress 166/173: ghcide ghcide >       (Use GADTs or TypeFamilies to permit this)
Progress 166/173: ghcide ghcide >     • When checking the inferred type
Progress 166/173: ghcide ghcide >         f :: forall (pass :: Pass).
Progress 166/173: ghcide ghcide >              (Outputable.OutputableBndr
Progress 166/173: ghcide ghcide >                 (NameOrRdrName (IdP (GhcPass (NoGhcTcPass pass)))),
Progress 166/173: ghcide ghcide >               Outputable.OutputableBndr (IdP (GhcPass (NoGhcTcPass pass))),
Progress 166/173: ghcide ghcide >               Outputable.OutputableBndr (IdP (GhcPass pass)),
Progress 166/173: ghcide ghcide >               Outputable.OutputableBndr (NameOrRdrName (IdP (GhcPass pass))),
Progress 166/173: ghcide ghcide >               Outputable (XIPBinds (GhcPass pass)),
Progress 166/173: ghcide ghcide >               Outputable (XViaStrategy (GhcPass pass)),
Progress 166/173: ghcide ghcide >               Outputable (XIPBinds (GhcPass (NoGhcTcPass pass))),
Progress 166/173: ghcide ghcide >               Outputable (XViaStrategy (GhcPass (NoGhcTcPass pass))),
Progress 166/173: ghcide ghcide >               NoGhcTcPass pass ~ NoGhcTcPass (NoGhcTcPass pass)) =>
Progress 166/173: ghcide ghcide >              Maybe Range -> ImportDecl (GhcPass pass) -> Maybe [TextEdit]
Progress 166/173: ghcide ghcide >       In the expression:
Progress 166/173: ghcide ghcide >         let
Progress 166/173: ghcide ghcide >           f (Just range) ImportDecl {ideclHiding}
Progress 166/173: ghcide ghcide >             = case ideclHiding of
Progress 166/173: ghcide ghcide >                 Just (False, x)
Progress 166/173: ghcide ghcide >                   | Set.notMember name (Set.fromList ...) -> ...
Progress 166/173: ghcide ghcide >                   | otherwise -> ...
Progress 166/173: ghcide ghcide >                 _ -> ...
Progress 166/173: ghcide ghcide >           f _ _ = Nothing
Progress 166/173: ghcide ghcide >           src_span = srcSpanToRange . getLoc $ lDecl
Progress 166/173: ghcide ghcide >         in f src_span . unLoc $ lDecl
Progress 166/173: ghcide ghcide >       In an equation for ‘extendImportList’:
Progress 166/173: ghcide ghcide >           extendImportList name lDecl
Progress 166/173: ghcide ghcide >             = let
Progress 166/173: ghcide ghcide >                 f (Just range) ImportDecl {ideclHiding} = ...
Progress 166/173: ghcide ghcide >                 f _ _ = Nothing
Progress 166/173: ghcide ghcide >                 src_span = srcSpanToRange . getLoc $ lDecl
Progress 166/173: ghcide ghcide >               in f src_span . unLoc $ lDecl
Progress 166/173: ghcide ghcide >     |
Progress 166/173: ghcide ghcide > 267 |     f (Just range) ImportDecl {ideclHiding} = case ideclHiding of
Progress 166/173: ghcide ghcide >     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
Progress 166/173: ghcide ghcide > 
```

It unblocks the bump of ghcide in hls (and the incoming release): https://github.com/haskell/haskell-language-server/pull/658.
Will rebase #946 when it is ready

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,9 +8,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # build not tested ghc versions
+        # all versions to only build or test for non windows os's
+        # inclusions will modify the major ones to mark them as testables
         os: [macOS-latest, ubuntu-latest]
         ghc: ['8.10.2', '8.10.1', '8.8.4', '8.8.3', '8.8.2', '8.6.5', '8.6.4']
+        ghc-lib: [false]
         include:
           # one ghc-lib build
           - os: ubuntu-latest
@@ -87,7 +89,7 @@ jobs:
       run: cabal build || cabal build || cabal build
 
     - name: Test
+      if: ${{ !matrix.ghc-lib && matrix.test }}
       shell: bash
       # run the tests without parallelism to avoid running out of memory
       run: cabal test --test-options="-j1 --rerun-update" || cabal test --test-options="-j1 --rerun" || LSP_TEST_LOG_COLOR=0 LSP_TEST_LOG_MESSAGES=true LSP_TEST_LOG_STDERR=true cabal test --test-options="-j1 --rerun"
-      if: ${{ !matrix.ghc-lib}} && {{matrix.test}}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,13 +24,13 @@ jobs:
             ghc-lib: true
           # run the testsuite only once per ghc major version
           - os: [macOS-latest, ubuntu-latest, windows-latest]
-          - ghc: '8.10.2'
+            ghc: '8.10.2'
             test: true
           - os: [macOS-latest, ubuntu-latest, windows-latest]
-          - ghc: '8.8.4'
+            ghc: '8.8.4'
             test: true
           - os: [macOS-latest, ubuntu-latest, windows-latest]
-          - ghc: '8.6.5'
+            ghc: '8.6.5'
             test: true
 
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,10 +23,13 @@ jobs:
             ghc: '8.10.1'
             ghc-lib: true
           # run the testsuite only once per ghc major version
+          - os: [macOS-latest, ubuntu-latest, windows-latest]
           - ghc: '8.10.2'
             test: true
+          - os: [macOS-latest, ubuntu-latest, windows-latest]
           - ghc: '8.8.4'
             test: true
+          - os: [macOS-latest, ubuntu-latest, windows-latest]
           - ghc: '8.6.5'
             test: true
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,30 +8,45 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macOS-latest, ubuntu-latest, windows-latest]
+        # build not tested ghc versions
+        os: [macOS-latest, ubuntu-latest]
         ghc: ['8.10.2', '8.10.1', '8.8.4', '8.8.3', '8.8.2', '8.6.5', '8.6.4']
-        test: [false]
-        ghc-lib: [false]
-        exclude:
-          - os: windows-latest
-            ghc: '8.10.2' # broken due to https://gitlab.haskell.org/ghc/ghc/-/issues/18550
-          - os: windows-latest
-            ghc: '8.8.4' # also fails due to segfault :(
         include:
           # one ghc-lib build
           - os: ubuntu-latest
             ghc: '8.10.1'
             ghc-lib: true
-          # run the testsuite only once per ghc major version
-          - os: [macOS-latest, ubuntu-latest, windows-latest]
+          # only test supported ghc major versions
+          - os: macOS-latest
             ghc: '8.10.2'
             test: true
-          - os: [macOS-latest, ubuntu-latest, windows-latest]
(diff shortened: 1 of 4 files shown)


ndmitchell (approved):
Why not just include the GADTs  extension unconditionally? CPP seems significantly worse than a fairly harmless extension.

jneira:
> 
> 
> Why not just include the GADTs extension unconditionally? CPP seems significantly worse than a fairly harmless extension.

ghc-8.6 defintely does not need it but we can give a try

ndmitchell:
Yes, it might be not required, but it's not harmful, so might as well simplify and include it.

pepeiborra:
Can we move this to #946? Otherwise, how do we know that it works?

jneira:
> 
> 
> Can we move this to #946? Otherwise, how do we know that it works?

my plan is rebase master after merging onto #946
(it has worked for me locally though)

jneira:
well in fact we should rebase #943 onto this one, to make ci pass

pepeiborra:
No, we should merge them both :)

jneira:
ok, will include it here

jneira:
dicho y hecho

jneira:
the ghc-lib build had `test:true` by default

pepeiborra (approved):
Thanks!

jneira:
Too much jobs, the matrix dont "overwrite" major versions with test enabled but create new ones, investigating

jneira:
After further investigation in my repo:
* include and exclude dont support lists like the matrix so you have to write all combinations one by one :disappointed: 
* we could make two separate jobs but it would duplicate allmost all steps
  * it seems the unique current way to share steps between jobs is create a dedicate github action
* we have several axis and corner cases:
  * os, ghc version, testability, using ghc-lib
    * we have three versions to test and four versions to only build 
  * windows only supports some of those ghc versions (to build and to test) and has a specific one (ghc-8.10.2.2) that can be testable

So the unique solution i can think of is create a matrix with all combinations (except ghc-lib and windows ghc-8.10.2.2) and exclude one by one the other ones. Or create a minimal matrix over test (3 versions) or only build (4 versions) without windows and include all desired cases one by one.
In both cases the list of inclusions or exclusions will be annoying. :worried: 

I am gonna start to sketch the second option (minimal matrix build over only build + inclusions)

.github/workflows/test.yml
```diff
@@ -1,27 +1,54 @@
...
+        # inclusions will modify the major ones to mark them as testables
+        os: [macOS-latest, ubuntu-latest]
+        ghc: ['8.10.2', '8.10.1', '8.8.4', '8.8.3', '8.8.2', '8.6.5', '8.6.4']
         ghc-lib: [false]
-        exclude:
-          - os: windows-latest
-            ghc: '8.10.2' # broken due to https://gitlab.haskell.org/ghc/ghc/-/issues/18550
-          - os: windows-latest
-            ghc: '8.8.4' # also fails due to segfault :(
         include:
```
jneira, line 16: this is the less worse version i got to work, feel free to suggest another better one :slightly_smiling_face: 
//cc @pepeiborra @ndmitchell

jneira:
I've opened a issue in github actions to let use lists in matrix include and exclude props: https://github.com/actions/runner/issues/857

jneira:
@pepeiborra Unfortunately we already released ghcide-0.6 in hackage (but no in github afaics, only exists the tag)
Could we release 0.6.1 with this merged, to use it in hls?

pepeiborra:
I don't have any objections against a 0.6.0.1 release in Hackage. But why is this needed though? 
- hls-plugin-api need only depend on ghcide 0.6.0
- HLS too, but it doesn't get released to Hackage anyway
- We can make a Hackage revision for ghcide 0.6.0 that lists it as incompatible with ghc 8.8.2 and 8.8.3

pepeiborra:
> 
> EDIT: I've just make the github 0.6.0 release, using the pr list from ChangeLog.md

I don't see the value in GitHub releases so I don't bother with them

jneira:
> I don't have any objections against a 0.6.0.1 release in Hackage. But why is this needed though?

If someone wants to build its own version of hls without the ghcide submodule, or a project (a standalone private plugin?) that uses ghcide or hls-plugin-api from hackage, with ghc-8.8.3 or ghc-8.8.2, it will fail.

Not sure if that case exists in the wild but it is possible and it is nicer for those users being able to use its current ghc version instead warn them in the package metadata that they can't.

But we can wait for someone asking for it, so as you consider.

> I don't see the value in GitHub releases so I don't bother with them

Well they are merely informative but we did releases for previous versions, so at least for consistency.

PR #948, opened 2020-12-10
Remove language extension completions.

gdevanla:
This PR removes Completions and CodeActions from ghcide thus de-duping the same functionality they is currently available in HLS through the Pragmas plugin module [here](https://github.com/haskell/haskell-language-server/blob/master/plugins/default/src/Ide/Plugin/Pragmas.hs)

Also addresses (https://github.com/haskell/haskell-language-server/issues/380)


pepeiborra:
Did you port the tests too?

gdevanla:
I  reviewed the existing tests in HLS. I think they  provide similar coverage consistent with the tests I removed. Here are the tests I looked at

https://github.com/haskell/haskell-language-server/blob/8f8aa7f50dac2a1952fa187e4eb9b9e29e6e76f2/test/functional/FunctionalCodeAction.hs#L368-L403

https://github.com/haskell/haskell-language-server/blob/8f8aa7f50dac2a1952fa187e4eb9b9e29e6e76f2/test/functional/Completion.hs#L96-L124

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -561,8 +560,9 @@ getCompletions ideOpts CC { allModNamesAsNS, unqualCompls, qualCompls, importabl
       result
         | "import " `T.isPrefixOf` fullLine
         = filtImportCompls
+        -- we leave this condition here and return empty list since hls implements this completion
```
pepeiborra, line 563: This looks odd. I would mention the relevant Github issues

pepeiborra (approved):
Thank you @gdevanla !

pepeiborra:
> I reviewed the existing tests in HLS. I think they provide similar coverage consistent with the tests I removed. Here are the tests I looked at
> 
> https://github.com/haskell/haskell-language-server/blob/8f8aa7f50dac2a1952fa187e4eb9b9e29e6e76f2/test/functional/FunctionalCodeAction.hs#L368-L403
> 
> https://github.com/haskell/haskell-language-server/blob/8f8aa7f50dac2a1952fa187e4eb9b9e29e6e76f2/test/functional/Completion.hs#L96-L124

I'm not convinced that the tests in HLS provide the same coverage. In particular, they do not check that completions for specific language pragmas are generated.

gdevanla:
Um. I am sorry I don't understand. I did not remove any completion test. I just added a test to make sure we don't provide any pragma related completions from ghcide.

Perhaps, you are referring to 'CodeActions'? If yes, I see the HLS tests are checking for specific pragmas, though they are not the same pragmas ghcide tests are covering. I assume, we would not need tests for all kinds of pragmas.

Please could you clarify and I can address them.

pepeiborra:
You are right, I didn't look closely enough at those tests and thought they were checking for completions, but they are not.

PR #950, opened 2020-12-13
Update URLs after move to haskell github org

jhrcek:
Although Github has redirects from the old URLs, there's no point in taking people through redirects.
Also removes a bit of potential for FUD :smile: 
I checked that all the renamed links still work


src/Development/IDE/Spans/Documentation.hs
```diff
@@ -87,7 +87,7 @@ getDocumentationsTryGhc env mod sources names = do
     unwrap _ n = mkSpanDocText n
 
     mkSpanDocText name =
-      pure (SpanDocText (getDocumentation sources name)) <*> getUris name
+      SpanDocText (getDocumentation sources name) <$> getUris name
```
jhrcek, line 90: Sorry, this is not URL fix, but couldn't help it :smile:

pepeiborra (approved):
Awesome, thanks!

PR #952, opened 2020-12-13
Deduplicate module not found diagnostics

pepeiborra:
Reworked version of #943. The root problem was in the use of stale data in rules - after adding a missing import the `GetDependencies` rule would fail, but `TypeCheck` would go ahead anyway with the last successful result of `GetDependencies` and generate a missing module diagnostic which duplicates the one already generated by `GetDependencies`.

In general it seems a very bad idea to use stale data in core rules, if only because it makes it very difficult to reason about diagnostics. I have removed all other uses I've seen. 

While I was debugging this with the command line driver, I found the default options to check all targets very noisy, and I changed the defaults (for the command line driver only). Moreover, I added tracing of rule failures.

Fixes haskell/haskell-language-server#630

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -894,7 +894,15 @@ getLinkableType f = do
 
 needsCompilationRule :: Rules ()
 needsCompilationRule = defineEarlyCutoff $ \NeedsCompilation file -> do
-  (ms,_) <- use_ GetModSummaryWithoutTimestamps file
+  -- It's important to use stale data here to avoid wasted work.
+  -- if NeedsCompilation fails for a module M its result will be  under-approximated
+  -- to False in its dependencies. However, if M actually used TH, this will
+  -- cause a re-evaluation of GetModIface for all dependencies
+  -- (since we don't need to generate object code anymore).
+  -- Once M is fixed we will discover that we actually needed all the object code
+  -- that we just threw away, and thus have to recompile all dependencies once
+  -- again, this time keeping the object code.
+  (ms,_) <- fst <$> useWithStale_ GetModSummaryWithoutTimestamps file
   -- A file needs object code if it uses TemplateHaskell or any file that depends on it uses TemplateHaskell
   res <-
     if uses_th_qq ms
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -553,16 +553,16 @@ getBindingsRule =
 getDocMapRule :: Rules ()
 getDocMapRule =
     define $ \GetDocMap file -> do
-      (tmrTypechecked -> tc,_) <- useWithStale_ TypeCheck file
-      (hscEnv -> hsc,_) <-useWithStale_ GhcSessionDeps file
-      (refMap -> rf, _) <- useWithStale_ GetHieAst file
+      (tmrTypechecked -> tc) <- use_ TypeCheck file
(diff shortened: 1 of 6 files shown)


src/Development/IDE/Core/Rules.hs
```diff
@@ -553,16 +553,16 @@ getBindingsRule =
 getDocMapRule :: Rules ()
 getDocMapRule =
     define $ \GetDocMap file -> do
-      (tmrTypechecked -> tc,_) <- useWithStale_ TypeCheck file
-      (hscEnv -> hsc,_) <-useWithStale_ GhcSessionDeps file
-      (refMap -> rf, _) <- useWithStale_ GetHieAst file
+      (tmrTypechecked -> tc) <- use_ TypeCheck file
```
wz1000, line 556: I don't think `GetDocMap` is a core rule - it is only used for showing haddocks on hover, and I think it should be as resilient as possible.

The impact of it having the incorrect information is very low.

src/Development/IDE/Core/Rules.hs
```diff
@@ -894,7 +894,7 @@ getLinkableType f = do
 
 needsCompilationRule :: Rules ()
 needsCompilationRule = defineEarlyCutoff $ \NeedsCompilation file -> do
-  ((ms,_),_) <- useWithStale_ GetModSummaryWithoutTimestamps file
+  (ms,_) <- use_ GetModSummaryWithoutTimestamps file
```
wz1000, line 897: Consider that `NeedsCompilation` is the only case where information flows from the top of the module hierarchy to the bottom i.e. to dependencies from root modules. This means that a failure in a module high up in the hierarchy can cause everything it (transitively) depends on to fail.

In this case, if `NeedsCompilation` fails for something at the top, its result will be under-approximated to be `False` in the dependencies. However, if this module actually used TH,  this will cause a re-evaluation of `GetModIface` for all dependencies (since we don't need to generate object code anymore). Once the module is fixed, we will discover that we actually needed all the object code that we just threw away, and thus have to recompile all dependencies once again, this time keeping the object code.

This is a lot of wasted work, so we need to think carefully about removing it.

src/Development/IDE/Core/Rules.hs
pepeiborra: Given that `getAtPoint` already calls `GetDocMap` with `useE`, can you produce an example where `useWithStale_` inside the `GetDocMap`rule will make it more resilient than `use_`?

wz1000: When we have a stale AST, but the user hasn't hovered, so we haven't computed a stale `DocMap` yet.

src/Development/IDE/Core/Rules.hs
pepeiborra: I agree that removing this `useWithStale_` is more tricky. What can be done:

1. Leave it as it is
2. Move the `useWithStale` to the recursive `NeedsCompilation` call
3. Remove the over approximation 

3 is not an option, it would be very counterintuitive behaviour for diagnostics to propagate from dependencies to root modules.

2 and 1 are semantically close but not equivalent. I think 2 is slightly better: it has fail-fast semantics and it's more intentional. But that said the risk of getting it wrong is high and it's not easy to test it, so I'll leave it as it is

src/Development/IDE/Core/Rules.hs
pepeiborra: Fair enough...

wz1000 (approved):

PR #954, opened 2020-12-14
Rescue stack windows build

pepeiborra:
I finally figured this puzzle out.

### DIFF
--- a/.azure/windows-stack.yml
+++ b/.azure/windows-stack.yml
@@ -1,0 +1,41 @@
+jobs:
+- job: ghcide_stack_windows
+  timeoutInMinutes: 120
+  pool:
+    vmImage: 'windows-2019'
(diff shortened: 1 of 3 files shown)


jneira (approved):
Many thanks for take care of the windows build. Moreover not being your local dev env (afaik)

pepeiborra:
At work we still use cabal v1 (sandboxes) to build Stackage and I ran into this problem today when trying to upgrade to ghcide 0.6.0. 

The problem is not specific to Windows but to any ghc built with DYNAMIC_GHC_PROGRAMS=NO (and cabal v1). 

To reproduce it locally I used a handy Nix expression: https://github.com/pepeiborra/ghcide/commit/5b90873f61d4d902c58d63bbd0c845856ddf24ae

PR #956, opened 2020-12-15
Expose Documentation module

gdevanla:
In an effort to move Completions into its own hls-plugin package we have a dependency to access the getDocumentation function exposed in this module. Therefore, can we expose this module so that we will be able to access that function.


jneira (approved):
LGTM, thanks!

PR #958, opened 2020-12-17
Prepare release 0.6.0.2

pepeiborra:
This is a point release to disable the auto extend import snippets from completions because of the issues reported in https://github.com/haskell/haskell-language-server/issues/679


jneira (approved):
lgtm

PR #959, opened 2020-12-20
Fix diagnostics update bug

### LINKED ISSUE #949: Sending LSP messages very quickly causes all diagnostics to disappear
This is an issue I encounter when using ghcide using [ALE](https://github.com/dense-analysis/ale) in neovim. (However, the issue can be reproduced without any of that; see below.)

**Symptom**
While working on code (in neovim using ALE), an edit action, as well as a `:w` save action, has a reasonable probability of removing all diagnostics from the current file. This is especially frequent on save and on pressing `ESC` to exit insert mode. It can usually be fixed by doing another edit action and undo'ing that, but it is quite annoying nevertheless.

**Reproduction**
Save [this script](https://gist.github.com/tomsmeding/7151a79106127f1b3540319bfb071cc0) to a file, say `ghcide-fast-send-bug.py`, and then run `python3 ghcide-fast-send-bug.py fast | ghcide --lsp`. Note that the last LSP message printed by ghcide is an empty diagnostic list, while there is certainly still an error in the file.
Removing the `fast` command-line argument adds a 2-second delay between the last two LSP messages printed by the script (please see the script!), which makes the bug disappear on my machine.

I believe the script should work on any machine, but perhaps things like the root path need to be changed to an existing path on your machine. I'm not familiar enough with the LSP protocol to say something sensible about that.

**System**
Arch Linux, ghcide manually built on commit `27b4250` (one above v0.6.0, and current HEAD). ghcide was built using cabal 3.2.0.0, ghc version 8.8.4.

pepeiborra:
Given a FOI F with non null typechecking diagnostics D, imagine the following scenario:

1. An edit notification for F is received, creating a new version
2. GetModTime is executed, producing 0 diagnostics.
  2.1 updateFileDiagnostics is called
  2.2 setStageDiagnostics is called
  2.3 LSP.updateDiagnostics is called with a new version, resetting all the
  diagnostics for F
  2.4 newDiags=[] in updateFileDiagnostics, which is different from D (the last
  published diagnostics), which enqueues a new publishDiagnostics [] in the
  Debouncer
3. An edit notification for F is received before typechecking has a chance to
run which undoes the previous edit
4. The debouncer publishes the empty set of diagnostics after waiting 0.1s
5. GetFileContents runs and since the contents of the file haven't changed since
the last time it ran, early cutoff skips everything donwstream

Since TypeCheck is skipped, the empty set of diagnostics stays published until
another edit comes. This is a bug.

The goal of this change is to prevent setStageDiagnostics from losing
diagnostics from other stages. To achieve this, we recover the old diagnostics
for all stages and merge them with the new stage.

Fixes #949

The test suite includes a new custom command that can be used to wait for arbitrary Shake actions, which can be used to implement the improvements outlined in #955

### DIFF
--- a/src/Development/IDE/Core/Rules.hs
+++ b/src/Development/IDE/Core/Rules.hs
@@ -734,7 +734,7 @@ getModSummaryRule = do
                 getModSummaryFromImports session fp modTime (textToStringBuffer <$> mFileContent)
         case modS of
             Right res@(ms,_) -> do
-                let fingerPrint = hash (computeFingerprint f dflags ms, hashUTC modTime)
+                let fingerPrint = hash (computeFingerprint f (fromJust $ ms_hspp_buf ms) dflags ms, hashUTC modTime)
                 return ( Just (BS.pack $ show fingerPrint) , ([], Just res))
             Left diags -> return (Nothing, (diags, Nothing))
 
@@ -742,16 +742,18 @@ getModSummaryRule = do
         ms <- use GetModSummary f
         case ms of
             Just res@(msWithTimestamps,_) -> do
-                let ms = msWithTimestamps { ms_hs_date = error "use GetModSummary instead of GetModSummaryWithoutTimestamps" }
+                let ms = msWithTimestamps {
+                    ms_hs_date = error "use GetModSummary instead of GetModSummaryWithoutTimestamps",
+                    ms_hspp_buf = error "use GetModSummary instead of GetModSummaryWithoutTimestamps"
+                    }
                 dflags <- hsc_dflags . hscEnv <$> use_ GhcSession f
-                -- include the mod time in the fingerprint
-                let fp = BS.pack $ show $ hash (computeFingerprint f dflags ms)
+                let fp = BS.pack $ show $ hash (computeFingerprint f (fromJust $ ms_hspp_buf msWithTimestamps) dflags ms)
                 return (Just fp, ([], Just res))
             Nothing -> return (Nothing, ([], Nothing))
     where
         -- Compute a fingerprint from the contents of `ModSummary`,
         -- eliding the timestamps and other non relevant fields.
-        computeFingerprint f dflags ModSummary{..} =
+        computeFingerprint f sb dflags ModSummary{..} =
             let fingerPrint =
                     ( moduleNameString (moduleName ms_mod)
                     , ms_hspp_file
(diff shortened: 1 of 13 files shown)


src/Development/IDE/Core/Rules.hs
```diff
@@ -734,24 +734,26 @@ getModSummaryRule = do
...
             Left diags -> return (Nothing, (diags, Nothing))
 
     defineEarlyCutoff $ \GetModSummaryWithoutTimestamps f -> do
         ms <- use GetModSummary f
         case ms of
             Just res@(msWithTimestamps,_) -> do
-                let ms = msWithTimestamps { ms_hs_date = error "use GetModSummary instead of GetModSummaryWithoutTimestamps" }
+                let ms = msWithTimestamps {
+                    ms_hs_date = error "use GetModSummary instead of GetModSummaryWithoutTimestamps",
+                    ms_hspp_buf = error "use GetModSummary instead of GetModSummaryWithoutTimestamps"
```
wz1000, line 747: Why this change?

pepeiborra: Because the buffer is not tracked by the fingerprint

src/Development/IDE/Core/Shake.hs
```diff
@@ -1051,77 +1040,45 @@ actionLogger = do
...
-    where
-        diagsBySource = Map.singleton (Just stage) (SL.toSortedList diags)
-        uri = filePathToUri' fp
+setStageDiagnostics uri ver stage diags ds = newDiagsStore where
+    -- When 'ver' is a new version, updateDiagnostics throws away diagnostics from all stages
+    -- This interacts bady with early cutoff, so we make sure to preserve diagnostics
+    -- from other stages when calling updateDiagnostics
+    -- But this means that updateDiagnostics cannot be called concurrently
+    -- for different stages anymore
+    updatedDiags = Map.singleton (Just stage) (SL.toSortedList diags) <> oldDiags
```
wz1000, line 1062: This is relying on the left-biasedness of `<>`, correct?

Minor nitpick, but I think `Map.insert` would be clearer.

wz1000:
Looks good, must have been a tough one to crack!

wz1000 (approved):

pepeiborra:
It was quite tricky, yeah. I have a vague memory of the diagnostics store semantics for version causing trouble in the past. @cocreature @ndmitchell do you recall this?

jneira (approved):
Thanks for improve test infra on the way

tomsmeding:
I have used this branch (well, https://github.com/haskell/ghcide/pull/959/commits/aad22e656d8422a7bb3f4e0ad2aad9d1b264afd9 to be precise) for an hour or two and it seems my problems reported in #949 are indeed fixed! Thanks a lot for the hard work!

PR #960, opened 2020-12-20
Strip prefixes from labels

gdevanla:
Addresses https://github.com/haskell/haskell-language-server/issues/682

### DIFF
--- a/src/Development/IDE/Plugin/Completions/Logic.hs
+++ b/src/Development/IDE/Plugin/Completions/Logic.hs
@@ -191,7 +191,7 @@ mkNameCompItem origName origMod thingType isInfix docs !imp = CI{..}
     insertText = case isInfix of
             Nothing -> case getArgText <$> thingType of
                             Nothing -> label
-                            Just argText -> label <> " " <> argText
+                            Just argText -> (stripPrefix label) <> " " <> argText
             Just LeftSide -> label <> "`"
 
             Just Surrounded -> label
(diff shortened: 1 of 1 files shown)


gdevanla:
For me to add a test for this, I also need to include multiple modules as part of the helper functions that test for completions. There is no ability to do that right now. Should I spend time on that or can I address that as part of the hls plugin change?

pepeiborra:
Whichever you prefer. I think that tests are best written at the same time as features or bug fixes

src/Development/IDE/Plugin/Completions/Logic.hs
```diff
@@ -191,7 +191,7 @@ mkNameCompItem origName origMod thingType isInfix docs !imp = CI{..}
     insertText = case isInfix of
             Nothing -> case getArgText <$> thingType of
                             Nothing -> label
-                            Just argText -> label <> " " <> argText
+                            Just argText -> (stripPrefix label) <> " " <> argText
```
pepeiborra, line 194: Why not change `stripAutogenerated` so that it applies `stripPrefix` to the `insertText` field too?

https://github.com/haskell/ghcide/blob/ee184ebd7d9f76711130d1f9dc5596ac0511c297/src/Development/IDE/Plugin/Completions/Logic.hs#L641-L645

pepeiborra (changes requested):
minor nitpick

src/Development/IDE/Plugin/Completions/Logic.hs
gdevanla: Yes. That was the initial proposed solution that was discussed https://github.com/haskell/haskell-language-server/issues/682.   But, the `insertText` of record snippets have `:` in them and that would break record snippet completions.  

In addition, I feel doing this in a targetting manner while constructing `label`  or `insertText` is desirable and give more control during construction.

pepeiborra: Seems reasonable

pepeiborra (approved):

PR #963, opened 2020-12-27
Invert the dependency with hls-plugin-api

pepeiborra:
This PR includes changes to implement the reorg described in https://github.com/haskell/ghcide/issues/936#issuecomment-751437853


ghcide.cabal
```diff
@@ -161,6 +162,9 @@ library
         Development.IDE.Plugin
         Development.IDE.Plugin.Completions
         Development.IDE.Plugin.CodeAction
+        Development.IDE.Plugin.Formatter
```
ndmitchell, line 165: Why do these end up in Ghcide? I guess they leave in a future diff?

ndmitchell (approved):
I agree the repo inversion is the right thing to do. The details are less important and easier to clean up after, so stamping it with approval so you can do so. (I did review the patch though, seems sensible)

ghcide.cabal
pepeiborra: `Formatter` contains machinery needed to convert an HLS `FormattingProvider` into a `PartialHandler` for ghcide

pepeiborra:
Closed in favour of https://github.com/haskell/haskell-language-server/pull/701/

PR #965, opened 2020-12-29
Prepare for archiving the Github project

pepeiborra:
Last PR before archiving

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,10 @@
+# Moved
+
+This project is now a historical artifact. It has been merged and all issues transferred into:
+
+https://github.com/haskell/haskell-language-server
+
+
 # `ghcide` - A library for building Haskell IDE tooling
 
 Our vision is that you should build an IDE by combining:


jneira (approved):

alanz:
What about the issues? They will also become read only. Should they have matching ones on HLS first?

jneira:
@alanz yeah, maybe we should link from ghcide to hls, close some of them without translating, etc

jneira:
After further debate in irc the plan would be handle all issues one by one:
* transferring the missing ones in hls
* "merge" with the correlated hls issue (linking or copying relevant comments)
* closing with a comment the no reproducible or with missing feedback from the author

Issues will be closed and the repo should be keep active meanwhile

pepeiborra:
I think we need to transfer all the closed issues too, to prevent lost knowledge. The "transfer issue" link rewrites # references in comments to point to the old repo, and installs a redirection from the old # to the new # so it should all work fine.

